PR #3, opened 2023-12-14
Permissive task capability

ivakoleva:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] WHOSUSING.md

**NOTE**: Please remember to run `./gradlew spotlessApply` to fix any format violations. (Check.)

Changes in this PR
----
**Why:**
We need idempotent forked tasks, meaning all tasks get executed, but any failures are still detected upon join. Freezed repo feature request https://github.com/Netflix/conductor/issues/3861 (Issue #)

**What:**
Introduced the concept of Permissive tasks.
A Permissive task skips the fast-fail, allowing idempotent execution of other tasks. It is similar to an Optional task, permitting the other tasks to continue. The difference is, a Permissive task failure affects a join or end of workflow accordingly.
Result is:
1. Forked Permissive tasks will let each other be evaluated, until all the forked tasks had terminated. Only then, the join task should fail. In case of Permissive optional tasks, the join will not fail.
2. Permissive sequential tasks will let subsequent tasks continue. While at the end, the workflow will fail in case a permissive task had failed. The workflow would not fail in case of Permissive optional task failure.

**Testing done:** 
TestDeciderOutcomes.testPermissive() added,
WorkflowAndTaskConfigurationSpec
"Test simple workflow which has a permissive task" and "Test simple workflow which has a permissive optional task added" that cover retry, ForkJoinSpec
"Test a simple workflow with fork join permissive failure flow" added and "Test retrying a failed permissive fork join workflow" added. 

In addition, performed e2e tests locally running a Conductor instance. Did build a docker image with the code changes made, started it locally, and started a SampleWorker to poll 3 tasks in parallel. Verified e2e scenarios of task_def_permissive, task_def_permissive_optional, task_def_simple.json, task_def_simple_optional.json, each joining on 6 forked tasks, then running simple task 7 after join.

Alternatives considered
----


v1r3n:
Hi @ivakoleva how is this different from marking task as optional.  I think the same result can be achieved by marking tasks as optional.

ivakoleva:
> Hi @ivakoleva how is this different from marking task as optional. I think the same result can be achieved by marking tasks as optional.

The capability is similar to a Simple optional in terms of not terminating the workflow in case of failure,
however differs from optional when a join/end of workflow is reached where the status still matters.

v1r3n (approved):

PR #9, opened 2023-12-17
Fixes #7, another dormant bug

### LINKED ISSUE #7: Basename feature in  #3722 doesn't work after "yarn build"
**Describe the bug**
The feature in  #3722 breaks the "NavLink" with "newTab". The package's homepage (aka basename) doesn't work after "yarn build". 

This works on v3.15.0 tag, shall we revert ?

@haricane8133

**Details**
Conductor UI on the latest main.

**To Reproduce**
1. Build and run the conductor UI
2. Goto Workflow Definitions
3. Click the "Executions" 's Query Link,
4. It doesn't work.

**Expected behavior**

**Screenshots**

**Additional context**

haricane8133:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #7 

Alternatives considered
----

### DIFF
--- a/ui/src/components/NavLink.jsx
+++ b/ui/src/components/NavLink.jsx
@@ -26,7 +26,8 @@ export default React.forwardRef((props, ref) => {
       </Link>
     );
   } else {
-    const href = absolutePath ? url.toString() : cleanDuplicateSlash(getBasename() + url.toString());
+    // Note: + '/' + is required here
+    const href = absolutePath ? url.toString() : cleanDuplicateSlash(getBasename() + '/' + url.toString());
     return (
       <Link ref={ref} target="_blank" href={href}>
         {rest.children}
--- a/ui/src/plugins/fetch.js
+++ b/ui/src/plugins/fetch.js
@@ -41,6 +41,12 @@ export function fetchWithContext(
     });
 }
 
+/**
+ * @param {string} path 
+ * @returns path with '/' not duplicated except at ://
+ * 
+ * Note: we need the second .replace() because the first one doesn't catch multiple slashes at the start of the string
+ */
 export function cleanDuplicateSlash(path) {
-  return path.replace(/([^:]\/)\/+/g, "$1");
+  return path.replace(/([^:]\/)\/+/g, "$1").replace(/^(\/)+/, '/');
 }
--- a/ui/src/plugins/fetch.js
+++ b/ui/src/plugins/fetch.js
@@ -43,10 +43,9 @@ export function fetchWithContext(
 
 /**
  * @param {string} path 
- * @returns path with '/' not duplicated except at ://
+ * @returns path with '/' not duplicated, except at ://
  * 
- * Note: we need the second .replace() because the first one doesn't catch multiple slashes at the start of the string
  */
 export function cleanDuplicateSlash(path) {
-  return path.replace(/([^:]\/)\/+/g, "$1").replace(/^(\/)+/, '/');
+  return path.replace(/(:\/\/)\/*|(\/)+/g, "$1$2");
 }


ui/src/components/NavLink.jsx
```diff
@@ -26,7 +26,8 @@ export default React.forwardRef((props, ref) => {
       </Link>
     );
   } else {
-    const href = absolutePath ? url.toString() : cleanDuplicateSlash(getBasename() + url.toString());
+    // Note: + '/' + is required here
+    const href = absolutePath ? url.toString() : cleanDuplicateSlash(getBasename() + '/' + url.toString());
```
nhandt2021, line 30: You add 1 more slash then replace it. Does it make sense?

ui/src/plugins/fetch.js
```diff
@@ -41,6 +41,12 @@ export function fetchWithContext(
...
 
+/**
+ * @param {string} path 
+ * @returns path with '/' not duplicated except at ://
+ * 
+ * Note: we need the second .replace() because the first one doesn't catch multiple slashes at the start of the string
+ */
 export function cleanDuplicateSlash(path) {
-  return path.replace(/([^:]\/)\/+/g, "$1");
+  return path.replace(/([^:]\/)\/+/g, "$1").replace(/^(\/)+/, '/');
```
nhandt2021, line 51: Why do we need to replace twice?
I think we should find another regex for this.

ui/src/components/NavLink.jsx
haricane8133: `getBaseName()` can return either 

`/abc/def`, or `/abc/def/`

We would have to cover both scenarios

haricane8133: Those two options are possible because we leave the work of filling the `homepage` field, to the user.

The user can enter the homepage url with a trailing slash, or not... It's up to them, and we should make sure things work both way

ui/src/plugins/fetch.js
haricane8133: The second replace isn't a costly operation as the associated regex has `^`.

Having two regex also ensures the code is more readable with simpler regexes...

nhandt2021: @haricane8133 Happy new year!
Sorry for late replying.
Here is the regex that I found: `/(https?:\/\/)|(\/)+/g`
Ex: `path.replace(/(https?:\/\/)|(\/)+/g, "$1$2")`

image

haricane8133: @nhandt2021, this is clever.

But it currently fails for
`'https:///some-domain.com//sub-domain/'.replace(/(https?:\/\/)|(\/)+/g, "$1$2")`

haricane8133: I would just add a `\/*` after `(https?:\/\/)` to fix it

`'https:///some-domain.com//sub-domain/'.replace(/(https?:\/\/)\/*|(\/)+/g, "$1$2")`

Do you think it would be better not to pin the protocol? The Util may be used else-where too...

How about this? We can remove `https?` from the regex
`'https:///some-domain.com//sub-domain/'.replace(/(:\/\/)\/*|(\/)+/g, "$1$2")`

haricane8133: I have pushed the code with the latest regex, and rebased commits from upstream.

nhandt2021: Actually I don't think we need to cover this case because it's completely wrong syntax.
How about if user input a URL like this: `https:#$@@#$@*&#@////////some-domain.abc`

haricane8133: We aren't solving for 100% of the possibilities...
(we have a `+ '/'` in our code base elsewhere and we cannot be sure where that extra '/' would be appended)

But we **MUST** make sure that the function, which deals only with '/', does what is expected out of it.

nhandt2021: > Do you think it would be better not to pin the protocol? The Util can be used elsewhere too... (the existing code didn't specify the protocol)
> 
> How about this? We can remove `https?` from the regex `'https:///some-domain.com//sub-domain/'.replace(/(:\/\/)\/*|(\/)+/g, "$1$2")`
> 
> Let us have the function as generic as possible

Agree with this :)

nhandt2021 (approved):
Looks good to me.

v1r3n (approved):

PR #36, opened 2024-01-08
Remove unnecessary Autowired from Constructors

tiffanyfay:
Pull Request type
----

- [ ] Bugfix
- [ ] Feature
- [ ] WHOSUSING.md

Changes in this PR
----
Removed Autowired import/bean
This doesn't change the code behavior/functionality. It's just cleanup.

Issue # N/A

Alternatives considered
N/A
----

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/EventQueues.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/EventQueues.java
@@ -18,7 +18,6 @@ import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.lang.NonNull;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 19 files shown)


c4lm (changes requested):
This is a good idea. Could you please make it so that full build completes successfully?
Not possible to just remove `@Autowired` from property classes

c4lm:
> @c4lm Thanks for reviewing and the feedback! I hope I addressed it as intended and added them back to property classes, otherwise more feedback would be appreciated!
> 
> In the property class, why is this treated differently? Thanks.
> 
> (Also, builds aren't currently working correctly in my machine right now which is why I didn't know it would fail in advance.)

For `@Component`, `@Service`, and the like, explicit `@Bean`s which don't require additional set up to show up in the context (like `@EnableConfigurationScan` etc.), as long as there's only one constructor, `@Autowired` should have no meaning and should be safe to remove.
But as for `@ConfigurationProperties` classes - they are usually supposed to be populated via field injection, not a constructor. So  I think in our case Spring simply does not pick that constructor unless we mark it as `@Autowired`.
I think that the most proper way would be to get rid of that dependency and make RedisProperties and ConductorProperties entirely separate, of course.

c4lm (approved):

tiffanyfay:
@c4lm Ahh ok. Thanks for the detailed explanation!

v1r3n (approved):

PR #66, opened 2024-02-07
Implement removeWorkflow and removeTask for PostgresIndexDAO

bjpirt:
Pull Request type
----
- [x] Feature

Changes in this PR
----
This PR implements `removeWorkflow` and `removeTask` in the `PostgresIndexDAO` class so that when you call `/api/workflow/{workflowId}/remove?archiveWorkflow=false` the workflow is removed from both the main workflow store as well as the index in postgres.

`archiveWorkflow=true` still has no effect because all this does when you're using ElasticSearch is to move the workflow data into the index, but there's no point doing this in Postgres since it's just moving from one place to another in the same database.


v1r3n (approved):

bjpirt:
@v1r3n Thanks for approving - are you OK to merge?

PR #68, opened 2024-02-09
Add a SetCorrelationId system task

bjpirt:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

The `SET_CORRELATION_ID` task allows users to set the correlation ID on a workflow from within the workflow. Normally the correlation ID would be set by the process starting the workflow, however in certain occasions, such as starting a workflow from an event subscription, this may not be possible.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskType.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskType.java
@@ -41,6 +41,7 @@ public enum TaskType {
     KAFKA_PUBLISH,
     JSON_JQ_TRANSFORM,
     SET_VARIABLE,
+    SET_CORRELATION_ID,
     NOOP;
 
     /**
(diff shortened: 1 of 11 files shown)


c4lm:
> However in certain occasions, such as starting a workflow from an event subscription, this may not be possible.

Hey, unless I misunderstand you, from what I see [here](https://github.com/conductor-oss/conductor/blob/1fb3405e6ebc56bbb5f177ea4a647ffb0d27d8f7/core/src/main/java/com/netflix/conductor/core/events/SimpleActionProcessor.java#L203), setting correlationId from event payload should work if you specify that in the event handler.
Could you describe your use-case in more detail? I think it would be better to add the capability to specify correlationId wherever you need before workflow is started (if it's currently missing in some place) instead of altering existing workflow.

bjpirt:
@c4lm The reason I implemented this is because we're using Conductor to process XML messages we receive in an SQS queue - these messages contain an external correlation ID we want to use for all of our subsequent Conductor workflows but we can't set this on the initial workflow, only the ones that it starts after it has parsed the message

PR #69, opened 2024-02-11
Fix issue: Return 404 instead of 2xx codes for methods handling non-existent tasks#38

vmsh1:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
Describe the new behavior from this PR, and why it's needed
Screenshot 2024-02-11 at 1 17 54 AM
Screenshot 2024-02-11 at 1 18 33 AM
Screenshot 2024-02-11 at 1 21 25 AM
Screenshot 2024-02-11 at 1 25 28 AM

Issue #

Alternatives considered
----

Describe alternative implementation you have considered

### DIFF
--- a/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
+++ b/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
@@ -73,7 +73,7 @@ public class TaskResource {
         // for backwards compatibility with 2.x client which expects a 204 when no Task is found
         return Optional.ofNullable(
                         taskService.batchPoll(taskType, workerId, domain, count, timeout))
-                .filter(logs -> !logs.isEmpty())
+                .filter(tasks -> !tasks.isEmpty())
                 .map(ResponseEntity::ok)
                 .orElse(ResponseEntity.notFound().build());
     }
--- a/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
+++ b/rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import com.netflix.conductor.core.exception.NotFoundException;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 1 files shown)


rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
```diff
@@ -103,8 +104,11 @@ public void log(@PathVariable("taskId") String taskId, @RequestBody String log)
 
     @GetMapping("/{taskId}/log")
     @Operation(summary = "Get Task Execution Logs")
-    public List<TaskExecLog> getTaskLogs(@PathVariable("taskId") String taskId) {
-        return taskService.getTaskLogs(taskId);
+    public ResponseEntity<List<TaskExecLog>> getTaskLogs(@PathVariable("taskId") String taskId) {
```
v1r3n, line 107: consider using https://github.com/conductor-oss/conductor/blob/386dde51069bff2280efc1b1cbb572b6da81a6da/core/src/main/java/com/netflix/conductor/core/exception/NotFoundException.java when the entity is not found.

vgopari: The code has been updated to utilize the `NotFoundException` class in all the relevant places where an entity is not found. Thank you for your input!
image
image
image
image

JCHacking:
Note that changing /poll and /poll/batch can break clients (since the expected codes were 200 with content if I get homework, and 204 without content if there was none).

If you also want those methods to give 404, this change should be labeled as BREAKING CHANGE.

vgopari:
@JCHacking, are you suggesting that we undo the code changes for `/poll` and `/poll/batch` to maintain the current behavior?

JCHacking:
> @JCHacking, are you suggesting that we undo the code changes for `/poll` and `/poll/batch` to maintain the current behavior?

No, I think it is more correct that if it fails to get a task it returns 404.

But if you make this change you may have to change the client package as well and the SDKs in other languages.

dcore94:
I don't think that for poll it is more correct to return 404. This could be correct if task definition is not known because in that case it would actually reflect a possible error condition. 204 means operation has no errors but returns no content which is much more correct and it would not break anything.

JCHacking:
> I don't think that for poll it is more correct to return 404. This could be correct if task definition is not known because in that case it would actually reflect a possible error condition. 204 means operation has no errors but returns no content which is much more correct and it would not break anything.

You right

vgopari:
I will rever the code to make the poll APIs return 204.

bradyyie (approved):

PR #78, opened 2024-02-15
Add an in-memory implementation of PollDataDAO

bjpirt:
Pull Request type
----
- [x] Feature

Changes in this PR
----

This PR adds a simple in-memory PollDataDAO implementation.

Why?
-----

When using Postgres as a backend, persisting this to the database results in hundreds of updates per second depending on the number of workers you have. Given this data is only used to provide basic info on when a queue was polled, using an in-memory implementation is much more efficient. The downside is that you only get the data stored in memory on the Conductor instance your API request routes to, however with many workers querying multiple times per second this makes very little tangible difference.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/polldata/MemoryPollDataDAO.java
+++ b/core/src/main/java/com/netflix/conductor/core/polldata/MemoryPollDataDAO.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2022 Conductor Authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/netflix/conductor/core/polldata/MemoryPollDataDAO.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.netflix.conductor.common.metadata.tasks.PollData;
+import com.netflix.conductor.dao.PollDataDAO;
+
+/** In-memory implementation of {@link PollDataDAO} which keeps the poll data locally in memory */
+public class MemoryPollDataDAO implements PollDataDAO {
+
+    private HashMap<String, HashMap<String, PollData>> pollData =
```
c4lm, line 27: HashMap is not thread-safe and these calls will come from different threads.

c4lm:
Thinking more about this - IMO it would be better to have this as a No-OP implementation instead to deter people from using it in production

PR #82, opened 2024-02-21
Implementation of Postgres PollDataDAO with write caching

bjpirt:
Pull Request type
----
- [x] Feature

Changes in this PR
----

- Split the PostgresPollDataDAO out into its own file
- Add caching with a periodic flush to the database
- Add read caching with a configurable stale period
- Add a `postgresql` document to the docs to bring all of the Postgres config together

### DIFF
--- a/docs/documentation/advanced/postgresql.md
+++ b/docs/documentation/advanced/postgresql.md
@@ -1,0 +1,50 @@
+# PostgreSQL
+
+By default conductor runs with an in-memory Redis mock. However, you
+can run Conductor against PostgreSQL which provides workflow management, queues and indexing.
+There are a number of configuration options that enable you to use more or less of PostgreSQL functionality for your needs.
+It has the benefit of requiring fewer moving parts for the infrastructure, but does not scale as well to handle high volumes of workflows.
+You should benchmark Conductor with Postgres against your specific workload to be sure.
+
+
+## Configuration
+
+To enable the basic use of PostgreSQL to manage workflow metadata, set the following property:
+
+```properties
+conductor.db.type=postgres
+spring.datasource.url=jdbc:postgresql://postgres:5432/conductor
+spring.datasource.username=conductor
+spring.datasource.password=password
+# optional
+conductor.postgres.schema=public
+```
+
+To also use PostgreSQL for queues, you can set:
+
+```properties
+conductor.queue.type=postgres
+```
+
+You can also use PostgreSQL to index workflows, configure this as follows:
+
+```properties
+conductor.indexing.enabled=true
+conductor.indexing.type=postgres
+conductor.elasticsearch.version=0
+```
+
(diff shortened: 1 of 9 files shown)


docs/documentation/advanced/postgresql.md
```diff
@@ -0,0 +1,48 @@
...
+You can also use PostgreSQL to index workflows, configure this as follows:
+
+```properties
+conductor.indexing.enabled=true
+conductor.indexing.type=postgres
+conductor.elasticsearch.version=0
+```
+
+By default, Conductor writes the latest poll for tasks to the database so that it can be used to determine which tasks and domains are active. This creates a lot of database traffic.
+To avoid some of this traffic you can configure the PollDataDAO with a write buffer so that it only flushes every x seconds:
```
VerstraeteBert, line 36: Important to note the effect here. What concession do I make when flushing periodically?

docs/documentation/advanced/postgresql.md
```diff
@@ -0,0 +1,48 @@
...
+
+```properties
+# Flush the data every 5 seconds
+conductor.postgres.pollDataFlushInterval=5
+```
+
+You can also configure a duration when the cached poll data will be considered stale. This means that the PollDataDAO will try to use the cached data, but if it is older than the configured period, it will check against the database.
```
VerstraeteBert, line 43: Same as above.

docs/documentation/advanced/postgresql.md
```diff
@@ -0,0 +1,48 @@
...
+You should benchmark Conductor with Postgres against your specific workload to be sure.
+
+
+## Configuration
+
+To enable the basic use of PostgreSQL to manage workflow metadata, set the following property:
+
+```properties
+conductor.db.type=postgres
+spring.datasource.url=jdbc:postgresql://postgres:5432/conductor
```
romankh3, line 16: Is it okay that you're not defining db schema?

postgres-persistence/src/test/java/com/netflix/conductor/postgres/dao/PostgresPollDataDAOCacheTest.java
```diff
@@ -0,0 +1,155 @@
...
+        properties = {
+            "conductor.app.asyncIndexingEnabled=false",
+            "conductor.elasticsearch.version=0",
+            "conductor.indexing.type=postgres",
+            "conductor.postgres.pollDataFlushInterval=200",
+            "conductor.postgres.pollDataCacheValidityPeriod=100",
+            "spring.flyway.clean-disabled=false"
+        })
+@SpringBootTest
+public class PostgresPollDataDAOCacheTest {
```
romankh3, line 63: thanks for adding test

docs/documentation/advanced/postgresql.md
bjpirt: Have added more detail to the docs

v1r3n (approved):

PR #87, opened 2024-02-23
Initial implementation of an experimental queue optimisation using Postgres LISTEN/NOTIFY

bjpirt:
Pull Request type
----
- [x] Feature

Changes in this PR
----
As more workers query Conductor for tasks, the load on Postgres increases. This uses LISTEN/NOTIFY to share the state of queues across all of the Conductor servers without needing to query every time. In tests it has reduced the number of queries against the `queue_message` by an order of magnitude. It also means that you can add more workers without linearly increasing the load on the database through this query.


postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+        return true;
+    }
+
+    public Integer getSize(String queueName) {
+        checkUpToDate();
+        handleNotifications();
+        if (notificationIsStale() || !connected) {
+            connect();
+            return null;
+        }
```
VerstraeteBert, line 87: Nit: prefer returning optional here, implicit nulls, while handled well now may result in bugs later.

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+    private void requestStats() {
+        try {
+            boolean previousAutoCommitMode = conn.getAutoCommit();
+            conn.setAutoCommit(true);
+            try {
+                conn.prepareStatement("SELECT queue_notify()").execute();
+                connected = true;
+            } catch (Throwable th) {
+                conn.rollback();
+                logger.info(th.getMessage());
```
VerstraeteBert, line 142: Leftover debug message?

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+            } finally {
+                conn.setAutoCommit(previousAutoCommitMode);
+            }
+        } catch (SQLException e) {
+            if (e.getSQLState() != "08003") {
```
VerstraeteBert, line 147: Assuming getSqlState returns a String, should use equals to compare values. == or != checks references.
```suggestion
            if (!e.getSQLState().equals("08003")) {
```

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+                    boolean previousAutoCommitMode = conn.getAutoCommit();
+                    conn.setAutoCommit(true);
+                    try {
+                        conn.prepareStatement("LISTEN conductor_queue_state").execute();
+                    } catch (Throwable th) {
+                        conn.rollback();
+                        logger.info(th.getMessage());
+                    } finally {
+                        conn.setAutoCommit(previousAutoCommitMode);
+                        connected = true;
```
VerstraeteBert, line 119: If the listen failed, why do we mark the connection as connected?

VerstraeteBert (commented):
This is very interesting work, if we manage to optimise the PG implementation + support for locking, there's no need for any other infra component anymore! I'll have a more in-depth look later, have a few initial questions regarding the big picture.

I think this feature is complex enough to document in writing on the modules. Is my understanding correct:
- Every Conductor instance will start a listen on 'conductor_queue_state'
- Every interaction with a queue message (i.e., pop, push, update) will result in a notify on that channel, containing the queueName and the earliest delivery time. This is cached by the listener.
- Every request for pop (e.g., during worker polls) will check the listener cache first, and only if there is at least one outstanding message on the queue, with a delivery time lower than the current time, the actual DB select is executed.

I'm mostly having a hard time reasoning about concurrent access to this, there's a few class members being accessed and mutated from multiple threads. Will try to reason about this more later.

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+
+    public boolean hasMessagesReady(String queueName) {
+        checkUpToDate();
+        handleNotifications();
+        if (notificationIsStale() || !connected) {
+            connect();
+            return true;
+        }
+
+        QueueStats queueStats = queues.get(queueName);
```
VerstraeteBert, line 69: Thinking out loud while learning :grin:. There's multiple methods where a queue.get() is done, which may be done concurrently with a full overwrite of the queues member to a new map. If we don't synchronise this, hasMessagesReady and getSize methods may read out-of-date data. Is this a problem? 

I'm thinking no, since the queue is supposed to be polled periodically by the workers anyways, the off chance of missing a last update to this is better than the overhead of synchronization.

postgres-persistence/src/main/resources/db/migration_postgres/V9__notify.sql
```diff
@@ -0,0 +1,50 @@
...
+
+CREATE TRIGGER queue_insert_delete
+  AFTER INSERT OR DELETE ON queue_message
+  FOR EACH ROW
+  EXECUTE FUNCTION queue_notify_trigger();
+
+CREATE TRIGGER queue_delete
+  AFTER DELETE ON queue_message
+  FOR EACH ROW
+  EXECUTE FUNCTION queue_notify_trigger();
```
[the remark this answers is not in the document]
bjpirt: Good spot - no it's not

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
bjpirt: No - should be logger.error instead though

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
bjpirt: Good spot - I've moved this

postgres-persistence/src/main/resources/db/migration_postgres/V9__notify.sql
```diff
@@ -0,0 +1,50 @@
+CREATE OR REPLACE FUNCTION queue_notify() RETURNS void
+LANGUAGE SQL
+AS $$
+  SELECT pg_notify('conductor_queue_state', (
+    SELECT
+      COALESCE(jsonb_object_agg(KEY, val), '{}'::jsonb) ||
+      jsonb_build_object('__now__', (extract('epoch' from CURRENT_TIMESTAMP)*1000)::bigint)
```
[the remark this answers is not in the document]
bjpirt: Yes, that's correct - good idea about the comment

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        return true;
+    }
+
+    public Integer getSize(String queueName) {
+        checkUpToDate();
+        handleNotifications();
+        if (notificationIsStale() || !connected) {
+            connect();
```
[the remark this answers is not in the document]
bjpirt: It shouldn't be. We request a new notification at 75% of the stale period, but being defensive, If it hasn't arrived then I assume there's something wrong so reconnect.

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private boolean notificationIsStale() {
+        return System.currentTimeMillis() - lastNotificationTime > this.stalePeriod;
+    }
+
+    private void connect() {
+        if (!connecting) {
+            connected = false;
+            connecting = true;
```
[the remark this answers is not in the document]
bjpirt: hmmm good point - not sure what the best approach to this is

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
bjpirt: Yeah, my thinking was that the data will move so quickly that this shouldn't be an issue. In a way, client 1 will get the latest data at the time they call the function and if client 2 changes the data while they are accessing the old data then that's fine.

bjpirt:
@VerstraeteBert 

> Is my overall understanding correct?

Yes - that's a good description - In one of the other PRs I've added a doc for postgres-specific configuration so once that's merged I'll add a description in there and consolidate all of the config. I'm also going to add some comments in the code too to make it clearer

I'm going to look in more detail at the concurrent access issues you've pointed out

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
VerstraeteBert: Entirely reasonable, I don't think the synchronization overhead is worth it.

postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresQueueListener.java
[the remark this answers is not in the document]
VerstraeteBert: To expand on the problem a little more. We can go a two ways, either we always re-connect if an issue occurs (like is done now), or first ensure that the connection is actually closed.

I'm leaning towards the latter, but then we need to assume that the LISTEN implementation is correct on the PG library we're using (which I think is a fair assumption).  With some help from openAI, I got a rather elegant solution that considers a few more edge cases, namely:
- A lock is used to ensure that no two threads attempt to connect at the same time
- The existing connection is checked for validity with [conn.isValid](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#isValid-int-)
- The old connection is closed if it is not valid

```java
import java.sql.Connection;
import javax.sql.DataSource;
import java.util.concurrent.locks.ReentrantLock;
import java.util.concurrent.locks.Lock;

public class YourClass {
    private volatile Connection conn;
    private final DataSource dataSource;
    private final Lock lock = new ReentrantLock();

    public YourClass(DataSource dataSource) {
        this.dataSource = dataSource;
    }

    public void connect() {
        // Attempt to acquire the lock without waiting.
        if (!lock.tryLock()) {
            // If the lock is not available, return early.
            return;
        }

        try {
            // Check if the connection is null or not valid.
            if (conn == null || !conn.isValid(1)) {
                // Close the old connection if it exists and is not valid.
                if (conn != null) {
                    try {
                        conn.close();
                    } catch (Exception e) {
                        // Log or handle exception from closing the old connection.
                    }
                }

               // Could nullify the connection member here first, but I'm scared of nullPointer exceptions
                
                // Establish a new connection.
                try {
                    conn = dataSource.getConnection();
                    // Perform any required initialization for the new connection here.
                } catch (Exception e) {
                    // Handle exceptions from obtaining a new connection.
                }
            }
        } catch (Exception e) {
            // Log or handle exception from checking connection validity.
        } finally {
            // Ensure the lock is always released.
            lock.unlock();
        }
    }

    // Other methods that use conn.
}
```

v1r3n (approved):

PR #92, opened 2024-02-26
Fix an issue with the postgres indices as they weren't being used properly

bjpirt:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----
There were some issues with the existing indexing of the Postgres index tables:
 - The two indexes on the task_index table were incorrectly created on the workflow_index table
 - The query we were making for fulltext search wasn't hitting the index properly
 - The query for JSON searching wasn't actually working as expected

This PR:
 - Removes the unnecessary indexes
 - Recreates the fulltext index on the task_index table
 - Creates new `gin` indexes on both tables to allow json queries to work properly
 - Updates the Java query generator to generate a query that will use the full text index properly

All creation and removal of the indices uses `CONCURRENTLY` to avoid any database locks. All statements also use `IF / IF NOT EXISTS` so that you can run the statements yourself manually and the migration will still work (see warning)

Note: This sets the flyway property `flyway.postgresql.transactional.lock=false` ([docs](https://documentation.red-gate.com/fd/postgresql-transactional-lock-224919738.html)) as there is an incompatibility with using `CONCURRENTLY` when creating indices (which should _always_ be used unless you like downtime during migrations) - see thread [here](https://github.com/flyway/flyway/issues/3491)

**Warning** If you already have lots of data in Postgres, these indexes can take some time to generate - you should run the statements manually at a time that is suitable for you to avoid them running whenever you deploy as Conductor may appear to hang while they are running.

### DIFF
--- a/postgres-persistence/src/main/java/com/netflix/conductor/postgres/config/PostgresConfiguration.java
+++ b/postgres-persistence/src/main/java/com/netflix/conductor/postgres/config/PostgresConfiguration.java
@@ -13,6 +13,7 @@
 package com.netflix.conductor.postgres.config;
 
 import java.sql.SQLException;
+import java.util.Map;
 import java.util.Optional;
 
 import javax.sql.DataSource;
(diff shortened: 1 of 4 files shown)


bjpirt:
I've had mixed success with this migration - whilst setting the transactional lock to false works fine for the tests it looks like it might be causing issues when running against a real server - still investigating

v1r3n (approved):

romankh3:
@bjpirt, I've faced problem with running from the scratch app with conductor, which is throwing exception:

`Caused by: org.flywaydb.core.api.FlywayException: Unknown configuration property: flyway.postgresql.transactional.lock`

could you please help?

PR #109, opened 2024-03-22
Fix of exception handler trying to write response to a disconnected c…

ivakoleva:
…lient

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
On Conductor server side, an exception gets raised `Failure in @ExceptionHandler com.netflix.conductor.rest.controllers.ApplicationExceptionMapper#handleAll(HttpServletRequest, Throwable)`. Cause is - attempting to write while the client aborted the connection `org.apache.catalina.connector.ClientAbortException: java.io.IOException: Broken pipe`

Added a ClientAbortException handler.

Testing done: Local run. Not trivial to emulate the issue in a test. Such a test may rely on timing out the client before server responded.

Alternatives considered
----

### DIFF
--- a/rest/src/main/java/com/netflix/conductor/rest/controllers/ApplicationExceptionMapper.java
+++ b/rest/src/main/java/com/netflix/conductor/rest/controllers/ApplicationExceptionMapper.java
@@ -15,6 +15,7 @@ package com.netflix.conductor.rest.controllers;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.catalina.connector.ClientAbortException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.Order;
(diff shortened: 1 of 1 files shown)


rest/src/main/java/com/netflix/conductor/rest/controllers/ApplicationExceptionMapper.java
```diff
@@ -53,7 +53,16 @@ public class ApplicationExceptionMapper {
         EXCEPTION_STATUS_MAP.put(NoResourceFoundException.class, HttpStatus.NOT_FOUND);
     }
 
+    @ExceptionHandler(ClientAbortException.class)
```
saksham2105, line 56: Please check the build failure, Probably a testcase is failing

manan164:
Hi @ivakoleva , Do we need this change? Since the exception is clear on the client side.

PR #117, opened 2024-04-04
Join task fixes

VerstraeteBert:
A Join task will wrongfully be marked as COMPLETED_WITH_ERRORS if any task that is optional fails, regardless of the presence of a non-optional failure. 

This PR fixes that, adds a more comprehensive test suite for the join task, and refactors the join routine, additionally adds more null checks in case of unscheduled tasks.

Pull Request type
----
- [x] Bugfix

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
@@ -37,7 +37,6 @@ public class Join extends WorkflowSystemTask {
     @SuppressWarnings("unchecked")
     public boolean execute(
             WorkflowModel workflow, TaskModel task, WorkflowExecutor workflowExecutor) {
-
         StringBuilder failureReason = new StringBuilder();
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -47,41 +45,47 @@ public boolean execute(
...
                     joinOn.stream()
                             .map(name -> TaskUtils.appendIteration(name, task.getIteration()))
-                            .collect(Collectors.toList());
+                            .toList();
         }
+
+        boolean allTasksTerminal =
+                joinOn.stream()
+                        .map(workflow::getTaskByRefName)
+                        .allMatch(t -> t != null && t.getStatus().isTerminal());
```
mr-celo, line 54: Do we need to loop over twice? Can't we just piggy-back on the loop below?

v1r3n (approved):

PR #121, opened 2024-04-09
Fix Issue: If failed acquire lock, system task pushed queue

ghkdwlgns612:
Pull Request type
----
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

| Case1 | Case2 |
|-------|-------|
|image|image|

Lock plays an important role in multi-thread environments (such as sweep-thread, system-task-worker-thread, etc.). However, because of Lock, SIMPLE tasks often did not become SCHEDULED after JOIN tasks. How about putting the decide() method back in the queue until it runs normally?

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -194,8 +194,27 @@ public class AsyncSystemTaskExecutor {
             }
             // if the current task execution has completed, then the workflow needs to be evaluated
             if (hasTaskExecutionCompleted) {
-                workflowExecutor.decide(workflowId);
+                forScheduleTaskEvaluate(task, queueName, workflowId);
             }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -194,11 +194,30 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
             }
             // if the current task execution has completed, then the workflow needs to be evaluated
             if (hasTaskExecutionCompleted) {
-                workflowExecutor.decide(workflowId);
+                forScheduleTaskEvaluate(task, queueName, workflowId);
```
manan164, line 197: Hi @ghkdwlgns612 , When multiple updates are coming at the same time for different tasks. The lock acquisition will fail. However, when next time the sweeper runs it must solve the issue and schedule the simple task. If that is not happening that actual issue is that.

ghkdwlgns612: Thank you for your response, let me check more:)

PR #145, opened 2024-05-07
Fix: Wrap PGExecutionDAO createTasks in a single transaction

VerstraeteBert:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----
We’ve observed during load tests that forked tasks are sometimes partially not scheduled during high load peaks.

After debugging, we’ve identified that this may occur due to the createTasks routine in the PostgresExecutionDAO has a separate transaction per task that is getting inserted. If an error / crash occurs in the middle, all the tasks inserted before the crash will be picked up by the WorkflowReconciler and pushed into the queue, while the ones not pushed will never be scheduled. This potentially leads to a workflow hanging forever.

Wrapping the createTasks routine in a single transaction should resolve this issue.

### DIFF
--- a/postgres-persistence/src/main/java/com/netflix/conductor/postgres/dao/PostgresExecutionDAO.java
+++ b/postgres-persistence/src/main/java/com/netflix/conductor/postgres/dao/PostgresExecutionDAO.java
@@ -136,9 +136,10 @@ public class PostgresExecutionDAO extends PostgresBaseDAO
     public List<TaskModel> createTasks(List<TaskModel> tasks) {
         List<TaskModel> created = Lists.newArrayListWithCapacity(tasks.size());
 
-        for (TaskModel task : tasks) {
-            withTransaction(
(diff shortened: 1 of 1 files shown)


postgres-persistence/src/main/java/com/netflix/conductor/postgres/dao/PostgresExecutionDAO.java
```diff
@@ -155,7 +156,7 @@ public List<TaskModel> createTasks(List<TaskModel> tasks) {
                                             + task.getReferenceTaskName()
                                             + ", key="
                                             + taskKey);
-                            return;
+                            continue;
```
VerstraeteBert, line 159: Should not return early when a task is already scheduled, one of the next ones could still not be.

bjpirt (approved):

v1r3n (approved):

PR #204, opened 2024-07-11
Python as Scripting language for inline task

saksham2105:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR

Added Support for Python as scripting language for inline task using Jython library. Along with this there is a restriction for import keywords and various inbuilt functions for safety constraints. Added support for variable replacements for Simple ,Nested JSON types usin `$.variable.variable2....` etc. Along with the Script expression in `inputs` map. We also have to add one more field in map named `outputIdentifier: something`

Use of `outputIdentifier`

Sample code :-
```
def add(a,b):
 return a + b
sum = add(1,2)
sum
```

In this case we will keep `outputIdentifier: "sum"` because this is the way we extract result of our Python code from Java using below snippet.

```
        String outputIdentifier = inputs.get("outputIdentifier").toString();
        PythonInterpreter interpreter = new PythonInterpreter();
        interpreter.exec(script);
        PyObject result = interpreter.get(outputIdentifier);

```

#### Roadmap : https://github.com/conductor-oss/conductor/blob/main/ROADMAP.md

----

Issue # `New feature in OSS roadmap`

Alternatives considered
We can also use `GraalPy` but it is super complex to setup
----

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -42,6 +42,8 @@ dependencies {
     implementation "com.github.ben-manes.caffeine:caffeine"
 
     implementation "org.openjdk.nashorn:nashorn-core:15.4"
+    implementation "org.python:jython-slim:${revJython}"
+
 
     // JAXB is not bundled with Java 11, dependencies added explicitly
     // These are needed by Apache BVAL
--- a/core/src/main/java/com/netflix/conductor/core/events/ScriptEvaluator.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/ScriptEvaluator.java
@@ -75,6 +75,9 @@ public class ScriptEvaluator {
      */
     private static Object evalPython(String script, Object input) throws ScriptException {
         Map<String, Object> inputs = (Map<String, Object>) input;
+        if (!inputs.containsKey("outputIdentifier")) {
+            throw new ScriptException("outputIdentifier is missing from task input");
+        }
         String outputIdentifier = inputs.get("outputIdentifier").toString();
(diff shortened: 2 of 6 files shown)


izrail-p:
Hi @saksham2105 
I’m having trouble returning a dictionary or list from my inline Python code. It works fine when I return a string, though. Any ideas on what might be going wrong?

sauronmesh:
Hey @izrail-p are you using jython? i think jython has issues

izrail-p:
@sauronmesh  yes. I am using  "evaluatorType": "python".

saksham2105:
Hey @izrail-p no my ques is are you using jython or graal?

izrail-p:
@saksham2105 I am using Jython.

izrail-p:
`        if (input instanceof Map<?, ?>) {
            Map<String, Object> inputs = (Map<String, Object>) input;
            if (inputs.containsKey("evaluatorType")
                    && inputs.get("evaluatorType").toString().equals("python")) {
                return evalPython(script, input);`
@saksham2105 python is added here as evaluatorType right. I tried returning object as well.

saksham2105:
I am not 100% About how to cast list and other data structure to jython PyObject

aliakbar-nathani:
Are you looking into this to support of returning dictionary from the inline python task ? Or is there any roadblock here?,  since it is quite common usecase to return a dictionary from the inline script.

Issue #213, opened 2024-07-18
Bug: Race Condition in WorkflowSweeper Leading to Inconsistent Workflow States


rq-dbrady:
**Describe the bug**
We have identified a race condition in the WorkflowSweeper class, which causes workflows to be in inconsistent states across different threads. This issue is critical as it affects the reliability and correctness of workflow execution and completion checks.

**Details**
Conductor version: 3.17
Persistence implementation: Postgres,Opensearch
Queue implementation: RedisCluster
Lock: Redis

**Steps to Reproduce:**
Deploy the application with at least 30 replicas in a Kubernetes environment.
Use a high sweeper rate of about 25ms and a high thread count.
Use a Redis cluster with Redis lock for workflow execution.
Execute workflows at a rate of approximately 75-90 workflows per second.
Monitor the state of workflows and observe for inconsistencies.

**Observed Behavior**
Workflows are fetched from executionDaoFacade before acquiring a lock.
The verifyAndRepair method mutates the workflow state without proper synchronization.
The workflow lock is released before the workflow is removed from the queue.
These conditions create a time window of roughly 50µ to 100µ seconds where a workflow can be in two states concurrently on different threads.
Workflow listeners or completion checks may fail as a result, with workflows erroneously marked as "Running" even after triggering the finish.

**Expected Behavior**
Workflows should maintain consistent states across all threads.
Proper locking should be enforced to prevent state mutations without synchronization.
Workflow locks should only be released after the workflow is securely removed from the queue.

**Screenshots**
<img width="2429" alt="Screenshot 2024-07-18 at 10 58 28" src="https://github.com/user-attachments/assets/b8472809-6818-41ed-8dab-8708fdb66c20">

v1r3n:
Hi @rq-dbrady we are investigating.

bradyyie:
Closing my own issue with fix.

PR #214, opened 2024-07-18
Fix Issue: Ensure proper locking in WorkflowSweeper to prevent race conditions

rq-dbrady:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
[Issue:#213](https://github.com/conductor-oss/conductor/issues/213)
Correct locking mechanism in WorkflowSweeper to prevent race conditions

Previously, the WorkflowSweeper.class had a potential race condition due to the order of operations in the sweep method. The workflow was being fetched from the executionDaoFacade before acquiring the lock, followed by a verifyAndRepair operation that could mutate the state. This sequence allowed for a small window (~50µ to 100µ seconds) where a workflow could be in two different states across different threads, causing inconsistencies and failures in workflow listeners or completion checks.

Changes made:
- Removed the `decideWithLock` method.
- Moved the locking logic directly into the `sweep` method to ensure the workflow is locked before any operations are performed on it.
- Ensured the workflow lock is only released after it is removed from the queue to prevent race conditions.

Observed issues:
- Race conditions were observed at large scale (30 replicas in Kubernetes, Redis cluster with Redis lock, ~75-90 workflows/sec).
- Workflows could be in a "Running" state even after triggering the finish, leading to listener and completion check failures.

New implementation in `sweep` method:
- Acquire lock before fetching the workflow and performing any operations.
- Handle verify and repair within the lock.
- Ensure the workflow is locked throughout the decision process.
- Release the lock only after removing the workflow from the queue.

This fix ensures atomicity in operations on workflows, preventing the race conditions previously observed.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
+++ b/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
@@ -17,6 +17,7 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 
+import org.apache.commons.lang3.time.StopWatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.annotation.Async;
@@ -35,6 +36,7 @@ import com.netflix.conductor.metrics.Monitors;
 import com.netflix.conductor.model.TaskModel;
 import com.netflix.conductor.model.TaskModel.Status;
 import com.netflix.conductor.model.WorkflowModel;
+import com.netflix.conductor.service.ExecutionLockService;
 
 import static com.netflix.conductor.core.config.SchedulerConfiguration.SWEEPER_EXECUTOR_NAME;
 import static com.netflix.conductor.core.utils.Utils.DECIDER_QUEUE;
@@ -49,6 +51,7 @@ public class WorkflowSweeper {
     private final WorkflowRepairService workflowRepairService;
     private final QueueDAO queueDAO;
     private final ExecutionDAOFacade executionDAOFacade;
+    private final ExecutionLockService executionLockService;
 
     private static final String CLASS_NAME = WorkflowSweeper.class.getSimpleName();
 
@@ -57,12 +60,14 @@ public class WorkflowSweeper {
             Optional<WorkflowRepairService> workflowRepairService,
             ConductorProperties properties,
             QueueDAO queueDAO,
-            ExecutionDAOFacade executionDAOFacade) {
+            ExecutionDAOFacade executionDAOFacade,
+            ExecutionLockService executionLockService) {
         this.properties = properties;
         this.queueDAO = queueDAO;
         this.workflowExecutor = workflowExecutor;
         this.executionDAOFacade = executionDAOFacade;
         this.workflowRepairService = workflowRepairService.orElse(null);
+        this.executionLockService = executionLockService;
         LOGGER.info("WorkflowSweeper initialized.");
     }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -78,20 +83,27 @@ public void sweep(String workflowId) {
             WorkflowContext workflowContext = new WorkflowContext(properties.getAppId());
             WorkflowContext.set(workflowContext);
             LOGGER.debug("Running sweeper for workflow {}", workflowId);
-
+            if (!executionLockService.acquireLock(workflow.getWorkflowId())) {
```
marcellusm2, line 86: here it will always be npe, should be just workflowId

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -78,20 +83,27 @@ public void sweep(String workflowId) {
...
+            StopWatch watch = new StopWatch();
+            watch.start();
+            try {
+                workflow = workflowExecutor.decide(workflow);
+                if (workflow != null && workflow.getStatus().isTerminal()) {
+                    queueDAO.remove(DECIDER_QUEUE, workflowId);
+                    return;
+                }
+            } finally {
+                executionLockService.releaseLock(workflow.getWorkflowId());
```
marcellusm2, line 103: release may never happen

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -78,20 +83,27 @@ public void sweep(String workflowId) {
             WorkflowContext workflowContext = new WorkflowContext(properties.getAppId());
             WorkflowContext.set(workflowContext);
             LOGGER.debug("Running sweeper for workflow {}", workflowId);
-
+            if (!executionLockService.acquireLock(workflow.getWorkflowId())) {
+                return;
```
marcellusm2, line 87: redundant

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
rq-dbrady: Resolved thanks for the catch !

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
rq-dbrady: Moved to the parent try catch finally.

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
rq-dbrady: Mirroring the existing functionality , if we cant acquire the lock we want to end the sweep flow for this workflow id , as there is the potential that another thread has it at this moment in time.

marcellusm2: sorry, I wanted to refer to the `return` of the release block, as there was the `finally`, but it has already been moved

core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowSweeper.java
```diff
@@ -74,24 +79,25 @@ public CompletableFuture<Void> sweepAsync(String workflowId) {
 
     public void sweep(String workflowId) {
         WorkflowModel workflow = null;
+        StopWatch watch = new StopWatch();
```
manan164, line 82: Hi @rq-dbrady , Instead of using stopwatch, Can we try to acquire a lock here and if successful, then we do repair or come out of sweep logic?

rq-dbrady: So currently it follows the logic:

1. Create stopwatch.
2. Set the workflow context.
3. AcquireLock (if fail return from sweep)
4. Gets workflow from store.
5. Run repair
6. Start Stop Watch.
7. Decide
8. Remove From DeciderQueue
9. Release lock and Stop StopWatch.

Are you proposing we Acquire the lock at the method start and then start the stop watch ? so something like

```
WorkflowModel workflow = null;
StopWatch watch = new StopWatch();
watch.start();
  //Do repair logic 
// Decide etc ...
}
```

manan164: Hi @rq-dbrady , Yes, the first thing we should try is to acquire a lock. then start a stop watch, run decide, stop stopwatch. release a lock. Optionally we dont need stop watch to measure execution time. We can use `System.currentTimeMillis()`

rq-dbrady: @manan164  , Applied changes in the logic order suggested  ! 

1. Acquire lock 
2. Run repair
3. Start timer
4. Decide
5. End timer
6. Release lock

v1r3n (approved):

PR #227, opened 2024-08-01
Updated workflow sweeper to use workflowOffsetTimeout for in progress tasks

danmiller192:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

**Preface:** Currently when the sweeper processes a workflow and the task is in progress the sweeper picks up the task timeout and queues a message on the decider queue with this time to next process the workflow. This can have the effect of leaving workflows in limbo for long periods with their tasks not being scheduled until that timeout is reached.

This fix pr addresses this by changing unack method to just use the workflowOffsetTimeout. This means that by default the workflow will be reprocessed in 30 seconds rather than the default 60 mins.

I have looked at ways to push the workflow to the decider but due to the asynchronous nature this can on occasion fall out of step.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
@@ -42,6 +42,10 @@ public class ConductorProperties {
     @DurationUnit(ChronoUnit.SECONDS)
     private Duration workflowOffsetTimeout = Duration.ofSeconds(30);
 
+    /** The maximum timeout duration to set when a workflow is pushed to the decider queue. */
+    @DurationUnit(ChronoUnit.SECONDS)
+    private Duration maxPostponeDurationSeconds = Duration.ofSeconds(3600);
+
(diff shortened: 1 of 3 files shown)


v1r3n:
@danmiller192 the current approach ensures that the workflows are not sweeped unless they need to be.  Changing this will have implications in the systems with large no. of running workflows.

danmiller192:
@v1r3n totally agree, I think it would be better to avoid the extra load. I did test with some large workflows running about 600 workflows per min and it didn't seem to cause any performance issues. When running that volume without the fix I did see significant volume of tasks that would get stuck waiting. I think I was intending this to be a workaround rather than a fix. I will try and see if I can find another way to address the issue.

v1r3n:
@danmiller192 what would be the advantage of sweeping a workflow before the task timeout?

danmiller192:
I have updated the proposal to use a config value MaxPostponeDurationSeconds to allow for tasks that may have a long and varied response time. Creating a new PR.

PR #248, opened 2024-08-26
Refactor to allow customization of conductor client library

maggopensrc:
The HTTP facing element of conductor client is refactored to allow effective customization. Also, supplying of http headers (as simple as an auth token when required) is taken care with providing defaults and hence with no changes in behavior.

Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [✓] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
The HTTP facing element of conductor client is refactored to allow effective customization. Also, supplying of http headers (as simple as an auth token when required) is taken care with providing defaults and hence with no changes in behavior.

### DIFF
--- a/client/src/main/java/com/netflix/conductor/client/http/ClientBase.java
+++ b/client/src/main/java/com/netflix/conductor/client/http/ClientBase.java
@@ -81,25 +81,48 @@ public abstract class ClientBase {
     }
 
     protected void delete(String url, Object... uriVariables) {
-        deleteWithUriVariables(null, url, uriVariables);
+        deleteWithUriVariables(null, null, url, uriVariables);
+    }
+
+    protected void delete(Map<String, Object> headers, String url, Object... uriVariables) {
+        deleteWithUriVariables(null, headers, url, uriVariables);
     }
 
     protected void deleteWithUriVariables(
             Object[] queryParams, String url, Object... uriVariables) {
-        delete(queryParams, url, uriVariables, null);
+        delete(queryParams, Map.of(), url, uriVariables, null);
+    }
+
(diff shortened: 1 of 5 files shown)


jmigueprieto:
Hi @maggopensrc - thank you for submitting this PR. 

We cannot merge it due to conflicts and the fact that tests [are failing](https://github.com/conductor-oss/conductor/actions/runs/10561912827/job/29583848537?pr=248). 

Also, I would like to better understand how you intend to use these changes. I have 2 concerns/question.

(1)

This PR is in fact changing the API of different clients. Not the public API but it's exposing a `protected` methods.

In `MetadataClient` I see something like:

```java
protected void registerWorkflowDef(WorkflowDef workflowDef, Map<String, Object> headers) {
   Validate.notNull(workflowDef, "Workflow definition cannot be null");
   postForEntityWithRequestOnly("metadata/workflow", headers, workflowDef);
}
```

I understand this is an invitation to use inheritance to use these methods. I'm not very fond of that idea and mixing headers with with a domain object (`WorkflowDef` in this case) can be seen as a violation of separation of concerns.

Is this the intended usage?

(2)

If you want to supply an auth token or other headers you have the option of filters e.g.: https://github.com/orkes-io/orkes-conductor-client/blob/0.2.0/src/main/java/io/orkes/conductor/client/http/auth/AuthorizationClientFilter.java

Maybe there is a valid use case for this other than auth headers, could you please share it to consider if we should move forward with this PR.

Last but not least, a new version of the client is going to be published soon which makes it even easier to add headers to requests.

PR #275, opened 2024-10-03
Feature/python inline task

inv-ajin:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
This PR implements support for Python as a scripting language for inline tasks in Netflix Conductor, based on the roadmap outlined [here](https://github.com/conductor-oss/conductor/blob/main/ROADMAP.md).
The feature extends Conductor's usability by enabling lightweight Python code execution as inline tasks.
The Python environment used for executing scripts is Jython.
Key changes include:

-     Added a new inline task type that supports executing Python scripts (previously only JavaScript was supported).
-     Introduced a new dependency for Jython in the build.gradle file inside conductor/core.
-     Implemented a new Evaluator to execute Python scripts located in conductor/core/java/core/executions/evaluators.
-     Added a new task type Python in the Java SDK.

Issue #

Alternatives considered
----
One alternative approach is to use GraalVM's Python engine instead of Jython, which could offer improved performance and compatibility.
----
Screenshot from 2024-10-03 15-04-33

Screenshot from 2024-10-03 15-02-08
Screenshot from 2024-10-03 15-01-14
Screenshot from 2024-10-03 15-00-14

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -43,6 +43,9 @@ dependencies {
 
     implementation "org.openjdk.nashorn:nashorn-core:15.4"
 
+    //jython dependency
+    implementation "org.python:jython-standalone:2.7.4"
+
     // JAXB is not bundled with Java 11, dependencies added explicitly
     // These are needed by Apache BVAL
     implementation "jakarta.xml.bind:jakarta.xml.bind-api:${revJAXB}"
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -43,8 +43,10 @@ dependencies {
 
     implementation "org.openjdk.nashorn:nashorn-core:15.4"
 
-    //jython dependency
-    implementation "org.python:jython-standalone:2.7.4"
+    //gralVm dependencies for executing python
+    implementation("org.graalvm.polyglot:polyglot:24.1.0")
+    implementation("org.graalvm.polyglot:python:24.1.0")
+    implementation "org.graalvm.sdk:graal-sdk:24.1.0"
 
     // JAXB is not bundled with Java 11, dependencies added explicitly
     // These are needed by Apache BVAL
--- a/core/src/main/java/com/netflix/conductor/core/execution/evaluators/PythonEvaluator.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/evaluators/PythonEvaluator.java
@@ -12,44 +12,35 @@
  */
 package com.netflix.conductor.core.execution.evaluators;
 
-import java.io.ByteArrayOutputStream;
-import java.io.PrintStream;
 import java.util.Map;
-import java.util.Properties;
 
-import org.python.core.PyObject;
-import org.python.util.PythonInterpreter;
+import org.graalvm.polyglot.Context;
+import org.graalvm.polyglot.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+
+import com.netflix.conductor.core.exception.TerminateWorkflowException;
 
 @Component(PythonEvaluator.NAME)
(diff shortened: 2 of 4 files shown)


core/build.gradle
```diff
@@ -43,6 +43,9 @@ dependencies {
 
     implementation "org.openjdk.nashorn:nashorn-core:15.4"
 
+    //jython dependency
+    implementation "org.python:jython-standalone:2.7.4"
```
v1r3n, line 47: How about using something like https://www.graalvm.org/python/

v1r3n (changes requested):
Thanks for the PR.  I would suggest take a look at the GraalVM implementation that is supposed to offer better performance overall.

https://www.graalvm.org/python/

It will also be useful to add support have pre-defined set of libraries added to the engine, which will allow working with 3P libraries that are statically defined and can be added at runtime.

inv-ajin:
Hi,
@v1r3n  I've made the requested changes and implemented using GralVm. Could you please review it again?

dcore94:
Sorry for jumping in but I'd like to ask that one think that I'm missing is the possibility that expression may be an array of strings where each array element represents a line of the expression. This would make writing the expression in JSON documents much much easier and more maintainable than using \n. Applies in general to INLINE Tasks...

v1r3n (approved):

v1r3n:
I am going to merge this for now @dcore94 I think the suggestion to support multiple lines is a good one.  I also wonder if there is a way to also reference external files etc.

@inv-ajin I suggest if you can followup with a PR to add additional modules (could be part of the build process or env variables etc at the startup) that will really improve the usability.
e.g. I would personally like to import langchain (https://github.com/langchain-ai/langchain) libraries so I can write LLM apps with python system task worker.

core/src/main/java/com/netflix/conductor/core/execution/evaluators/PythonEvaluator.java
```diff
@@ -0,0 +1,78 @@
...
+                // Execute the wrapped expression
+                context.eval("python", wrappedExpression.toString());
+
+                // Get the result
+                Value result = context.getBindings("python").getMember("result");
+
+                // Convert the result to a Java object and return it
+                return result.as(Object.class);
+            } else {
+                return null;
```
RamOrkes, line 71: Is there a usecase where an expression is useful for the workflow, and can be formed without inputs from the  InlineTask?

korolkevich7: Yes, uuid generation for example

core/src/main/java/com/netflix/conductor/core/execution/evaluators/PythonEvaluator.java
```diff
@@ -0,0 +1,78 @@
...
+import com.netflix.conductor.core.exception.TerminateWorkflowException;
+
+@Component(PythonEvaluator.NAME)
+public class PythonEvaluator implements Evaluator {
+    public static final String NAME = "python";
+    private static final Logger LOGGER = LoggerFactory.getLogger(PythonEvaluator.class);
+
+    @Override
+    public Object evaluate(String expression, Object input) {
+        try (Context context = Context.newBuilder("python").allowAllAccess(true).build()) {
```
tykarol, line 32: This is a security vulnerability; I can access os variables and the config_file (read it, but maybe even modify filesystem)

tykarol: @v1r3n ⬆️

Issue #237, opened 2024-08-15
workflow_index.update_time NOT NULL contraint breaking upgrade


lbestatlas:
**Describe the bug**
This bug has several aspects:

1. When running the postgres v13 flyway upgrade we have found some workflows without update_time in the json_data column, which causes update_time to be null when the data is backfilled in the v13 script. The subsequent application of the NOT NULL constraint fails
2. As we are running old and new versions of conductor concurrently as we rollout, workflows executed in the old version have failures indexing workflows as workflow_index.update_time is NULL
3. The new code does not set the update_time in the case of an update only an insert, so is unlikely to fix the issue it was intended to fix.
4. The backfill script on the workflow_index can lock up the table for minutes in environments where there is a large number of workflows.

We detected these issues when upgrade our service running v3.20.0 to v3.21.3. We run conductor with the following relevant properties as per the defaults
ConductorProperties.asyncIndexingEnabled = false
PostgresProperties.onlyIndexOnStatusChange = false

**Details**
Conductor version: 3.21.3
Persistence implementation: Postgres
Queue implementation: dynomite
Lock: redis
Workflow definition: n/a
Task definition: n/a
Event handler definition: n/a

**To Reproduce**
Run the flyway upgrade with running workflows

**Expected behavior**
The upgrade succeeds and no errors occur during the rollout where new and old versions are executing workflows

**Screenshots**

**Additional context**
This is an examples of the error from the flyway upgrade
```
SQL State  : 23502
Error Code : 0
Message    : ERROR: column "update_time" of relation "workflow_index" contains null values
Location   : db/migration_postgres/V13__workflow_index_columns.sql (/opt/service/nested:/opt/service/service.jar/!BOOT-INF/lib/conductor-postgres-persistence-3.21.3.jar!/db/migration_postgres/V13__workflow_index_columns.sql)
Line       : 7
Statement  : Run Flyway with -X option to see the actual statement causing the problem

```

lbestatlas:
Our proposed fix is to:
1. Remove the v13 migration
2. Add a v14 migration which either adds update_time with a default of 'epoch' if the column doesn't exist, and sets the default of 'epoch' if the column does exist
3. Fix the PostgresIndexDao.indexWorkflow to also set the update_time during an UPDATE
4. Either remove the backfill entirely as this data does not really require to be backfilled for the purpose of this change

dingliu1:
[...]

PR #288, opened 2024-10-15
workflow_index.update_time bugfix

lbestatlas:
Pull Request type
----
- [ x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Issue # https://github.com/conductor-oss/conductor/issues/237 

1. When running the postgres v13 flyway upgrade we have found some workflows without update_time in the json_data column, which causes update_time to be null when the data is back filled in the v13 script. The subsequent application of the NOT NULL constraint fails.
2. When running old and new versions of conductor concurrently during rollout, workflows executed in the old version have failures indexing workflows as workflow_index.update_time is NULL
3. The new code does not set the update_time in the case of an update only an insert, so this seems incorrect if the intention was to prevent out of sequence updates to the workflow_index.
4. The back-fill script on the workflow_index can potentially lock up the table for minutes in environments where there is a large number of workflows indexed.
5. Also, the back-fill script time format incorrectly removes precision from the milliseconds component.

To fix these issues this PR make the following changes:

1. Replaces the v13 flyway migration script with V13.1 that adds update_time with a default of 'epoch' if the column doesn't exist, and sets the default of 'epoch' if the column does exist. Adding a default in postgres is inexpensive, as it does not update existing rows. The default means the constraint can be added without running the back-fill and stops the issues when have old and new versions running at the same time.

2. Fix the PostgresIndexDao.indexWorkflow to also set the update_time during an UPDATE. This will ensure updateTime accurately reflects the most recent write.

3. Address the issues with the back-fill script logic, to ensure it doesn't try to insert NULL values and correctly formats the timestamp.

4. Make the back-fill script for update_time optional, so that expensive data migrations do not need to be run at start up.

Alternatives considered
----

I considered removing the not null constraint. However, this meant more complications dealing with null values in the consuming code. So given a default is inexpensive to add this seemed preferable.

I also considered dropping the back-fill script entirely, as it doesn't appear to impact any logic. However, I decided to add the ability to disable data migrations, for expensive scripts. Flyway community edition does not support https://flywaydb.org/blog/skipExecutingMigrations so it is impossible to defer scripts from being run during the migration if using that edition.


v1r3n (approved):

lbestatlas:
@v1r3n I have made 2 updates to the PR

1. to fix the build as I had not applied spotless
2. a change to the backfill script, as I identified further issues with this when I re-tested the script

postgres-persistence/src/main/resources/db/migration_postgres_data/V13.2__workflow_index_backfill_update_time.sql
```diff
@@ -0,0 +1,4 @@
+-- Optional back-fill script to populate updateTime historically.
+UPDATE workflow_index
+SET update_time = to_timestamp(json_data->>'updateTime', 'YYYY-MM-DD"T"HH24:MI:SS.MS')::timestamp AT TIME ZONE '00:00'
```
lbestatlas, line 3: Further testing found that the previous scripts formatting was not interpreting hours and time zones correctly.
Postgres does not interpret the timezone Z from the json->>'updateTime', it has to be set explicitly. Also, the T has to be explicitly skipped by enclosing it in double quotes.

PR #309, opened 2024-11-14
Feature/939208 fix persistence

lironleizer:
Pull Request type: Bugfix

**Changes in this PR:** Improve Workflow Reliability during Gradual Restarts with Optional Redis Idempotence

**Description:** This PR addresses an issue encountered when Conductor is deployed across multiple machines, particularly during a gradual restart sequence. Previously, if a workflow event was consumed from RabbitMQ and saved to Redis but had not yet been processed when a machine shut down, the event would reappear in the RabbitMQ queue. At this point, another machine could re-consume the event, but since the original entry in Redis had not yet expired, Conductor would incorrectly assume that this event was already processed by another machine, resulting in an early acknowledgment to RabbitMQ. This led to the loss of events and untriggered workflows, impacting reliability and predictability.

**The Fix:** This update introduces a new property, eventExecutionPersistenceEnabled, which offers greater flexibility and control over event handling. When set to false, this property disables the Redis persistence mechanism for idempotence, delegating responsibility for deduplication and idempotence to business services. By default, eventExecutionPersistenceEnabled is set to true, preserving the current behavior for backward compatibility.

**Key Benefits:**

**Enhanced Reliability:** Prevents workflow event loss during machine restarts by allowing business services to handle idempotence directly.
**Backward Compatibility:** Maintains the existing Redis persistence behavior by default, ensuring seamless integration for existing deployments.
**Greater Control:** Empowers operators to manage idempotence in ways that best suit their distributed environments.
This update brings added robustness to distributed setups and ensures workflows are processed reliably, even during server transitions.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
+++ b/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
@@ -78,8 +78,7 @@ public class ExecutionService {
         this.queueTaskMessagePostponeSecs =
                 properties.getTaskExecutionPostponeDuration().getSeconds();
 
-        this.isEventExecutionPersistenceEnabled =
-                properties.isEventExecutionPersistenceEnabled();
+        this.isEventExecutionPersistenceEnabled = properties.isEventExecutionPersistenceEnabled();
 
         this.systemTaskRegistry = systemTaskRegistry;
         this.taskStatusListener = taskStatusListener;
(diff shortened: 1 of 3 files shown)


lironleizer:
@dilip-lukose  @v1r3n 
please advise.

server/src/main/resources/application.properties
```diff
@@ -73,6 +73,9 @@ conductor.default-event-queue.type=sqs
 conductor.app.workflow-execution-lock-enabled=false
 conductor.workflow-execution-lock.type=noop_lock
 
+#Used to enable/disable the duplicate message delivery persistence
+conductor.app.eventExecutionPersistenceEnabled=true
```
manan164, line 77: Should we keep default value to false?

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -576,15 +580,22 @@ public List<Task> getPendingTasksForTaskType(String taskType) {
...
     public void removeEventExecution(EventExecution eventExecution) {
-        executionDAOFacade.removeEventExecution(eventExecution);
+        if (isEventExecutionPersistenceEnabled) {
+            executionDAOFacade.removeEventExecution(eventExecution);
+        }
     }
 
     public void updateEventExecution(EventExecution eventExecution) {
-        executionDAOFacade.updateEventExecution(eventExecution);
+        if (isEventExecutionPersistenceEnabled) {
```
manan164, line 596: Why are we only updating event execution when the eventExecutionPersistenceEnabled is set to true. We should update in any case. Please help me understand.

server/src/main/resources/application.properties
v1r3n: +1 to preserve backwards compatibility this should be false.

bradyyie:
Closing Due to Stale MR.

PR #313, opened 2024-11-19
Adding total timeout to task models

shaileshpadave:
Pull Request type
----
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
Added new parameter total task timeout to taskmodel and task def. This will be useful to timeout task if it exceeds total timeout value (including timeout in retry as well), this PR is only for model changes. Core service changes will be in another PR.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
@@ -97,27 +97,19 @@ public class Task {
     @ProtoField(id = 9)
     private String taskDefName;
 
-    /**
-     * Time when the task was scheduled
-     */
+    /** Time when the task was scheduled */
     @ProtoField(id = 10)
     private long scheduledTime;
 
-    /**
-     * Time when the task was first polled
-     */
+    /** Time when the task was first polled */
     @ProtoField(id = 11)
     private long startTime;
 
-    /**
-     * Time when the task completed executing
-     */
+    /** Time when the task completed executing */
     @ProtoField(id = 12)
     private long endTime;
 
-    /**
-     * Time when the task was last updated
-     */
+    /** Time when the task was last updated */
     @ProtoField(id = 13)
     private long updateTime;
 
@@ -216,8 +208,7 @@ public class Task {
     // If the task is an event associated with a parent task, the id of the parent task
     private String parentTaskId;
 
-    public Task() {
-    }
+    public Task() {}
 
     /**
      * @return Type of the task
@@ -429,7 +420,7 @@ public class Task {
 
     /**
      * @return True if the task has completed its lifecycle within conductor (from start to
-     * completion to being updated in the datastore)
+     *     completion to being updated in the datastore)
      */
     public boolean isExecuted() {
         return executed;
@@ -491,7 +482,7 @@ public class Task {
 
     /**
      * @param responseTimeoutSeconds - timeout for task to send response. After this timeout, the
-     *                               task will be re-queued
+     *     task will be re-queued
      */
     public void setResponseTimeoutSeconds(long responseTimeoutSeconds) {
         this.responseTimeoutSeconds = responseTimeoutSeconds;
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
```diff
@@ -202,6 +202,9 @@ public boolean isRetriable() {
     @ProtoField(id = 42)
     private boolean subworkflowChanged;
 
+    @ProtoField(id = 43)
+    private long firstStartTime;
```
jmigueprieto, line 206: should be added to `equals`, `hashcode` and `toString`?

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -220,6 +232,20 @@ public void setTimeoutSeconds(long timeoutSeconds) {
         this.timeoutSeconds = timeoutSeconds;
     }
 
+    /**
+     * @return the total timeout in seconds
+     */
+    public long getTotalTimeoutSeconds() {
```
jmigueprieto, line 238: `equals` & `hashCode`?

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -142,11 +146,17 @@ public TaskDef(String name, String description) {
         this.description = description;
     }
 
-    public TaskDef(String name, String description, int retryCount, long timeoutSeconds) {
+    public TaskDef(
```
jmigueprieto, line 149: Can a new constructor be added instead of changing the signature of this one?

jmigueprieto (changes requested):
- There are changes to backend and client v4 - split this PR.
- The client changes are breaking changes. Do NOT modify the signature of existing constructors. They might be in use by someone and adding a property should NOT break existing code.

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
v1r3n: Actually, I think we should change the hashCode to be aware of the task's PK -- which is taskName. Thoughts @jmigueprieto

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
v1r3n: +1

common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -464,6 +468,14 @@ public void setEnforceSchema(boolean enforceSchema) {
         this.enforceSchema = enforceSchema;
     }
 
+    public long getTotalTimeoutSeconds() {
+        return totalTimeoutSeconds;
+    }
+
+    public void setTotalTimeoutSeconds(@NotNull long totalTimeoutSeconds) {
```
jmigueprieto, line 475: NIT: a `long` (primitive type) cannot be null. The `@NotNull` is redundant.

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
jmigueprieto: @v1r3n  `hashCode` is considering `taskDefName`

```
  @Override
    public int hashCode() {
        return Objects.hash(
                getTaskType(),
                getStatus(),
                getInputData(),
                getReferenceTaskName(),
                getWorkflowPriority(),
                getRetryCount(),
                getSeq(),
                getCorrelationId(),
                getPollCount(),
                getTaskDefName(),
                ...
}
```

Or do you want to use just "task name"? 
**PROS**
- It will be simpler 
- (small) Performance gain

**CONS**
- We'll need to modify `equals`:  if only id is used in hashCode but equals considers additional fields, the contract will be broken. 
- We'll have an incomplete equality: this might break existing code e.g.: tests that assert `assertNotEquals(task0, task1)`. If both task have the same name but different fields, the assertion is `true` now but will be `false` is we just consider task name.

jmigueprieto: I think we can analyze this further in another PR. But, for the moment I believe that we should stick to considering the added fields to `equals` and `hashCode`

jmigueprieto (approved):
LGTM after NIT changes

PR #334, opened 2024-12-12
fix: latter task cannot be executed because taskResult.getLogs() can return null and decide(workflowId) will not be invoked

ColinZou:
…return null and decide(workflowId) will not be invoked

Pull Request type
----
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
Added null check for return value taskResult.getLogs before saving task execution log when update task.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
@@ -869,10 +869,13 @@ public class WorkflowExecutorOps implements WorkflowExecutor {
                             task.getTaskId(), workflowId);
             LOGGER.error(errorMsg, e);
         }
-
-        taskResult.getLogs().forEach(taskExecLog -> taskExecLog.setTaskId(task.getTaskId()));
-        executionDAOFacade.addTaskExecLog(taskResult.getLogs());
-
+        List<TaskExecLog> taskLogs = taskResult.getLogs();
+        Optional.ofNullable(taskLogs)
+          .ifPresent(
+                  logs -> {
+                      logs.forEach(taskExecLog -> taskExecLog.setTaskId(task.getTaskId()));
+                      executionDAOFacade.addTaskExecLog(taskLogs);
+                  });
         if (task.getStatus().isTerminal()) {
             long duration = getTaskDuration(0, task);
             long lastDuration = task.getEndTime() - task.getStartTime();


ColinZou:
@jmigueprieto could you please check this PR. The input logs of TaskResult can be null somehow and nullpointer exception throws.

jmigueprieto:
> @jmigueprieto could you please check this PR. The input logs of TaskResult can be null somehow and nullpointer exception throws.

Hey @ColinZou - Thanks for bringing this up. I'll take a look.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
```diff
@@ -869,10 +869,13 @@ public TaskModel updateTask(TaskResult taskResult) {
                             task.getTaskId(), workflowId);
             LOGGER.error(errorMsg, e);
         }
-
-        taskResult.getLogs().forEach(taskExecLog -> taskExecLog.setTaskId(task.getTaskId()));
-        executionDAOFacade.addTaskExecLog(taskResult.getLogs());
-
+        List<TaskExecLog> taskLogs = taskResult.getLogs();
+        Optional.ofNullable(taskLogs)
```
jmigueprieto, line 873: @ColinZou - A `TaskResult` instance is initialized with `logs` as an empty List. A NPE could only happen if `taskResult.setLogs(null)` is called before `updateTask(taskResult)`.

Could you please add a test to reproduce the issue or explain the execution path? Thanks!

Also, I would much rather do a if null check instead of `Optional` because:

1. Overhead: very minimal but wrapping taskLogs in an `Optional` introduces some.
2. Not the intended use case: `Optional` is primarily meant for return types rather than as a replacement for null checks in existing code. (Check the [javadoc])(https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Optional.html)

jmigueprieto (changes requested):
- Add a test or explain the execution path that leads to a NPE
- Change Optional usage.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
ColinZou: Sure, I found this NPE when invoking update task from rest API. The update task API accepts logs=null. I will also change the code soon.

ColinZou:
I found this NPE when invoking update task from rest API. The update task API accepts logs=null and the NPE get triggered in the place I fixed.

v1r3n (approved):

v1r3n:
> * Add a test or explain the execution path that leads to a NPE
> * Change Optional usage.

Hi @ColinZou can you take care of the comments and we can merge.

ColinZou:
> > * Add a test or explain the execution path that leads to a NPE
I've reply the message.  this one:
> I found this NPE when invoking update task from rest API. The update task API accepts logs=null and the NPE get triggered in the place I fixed.

> > * Change Optional usage.
It was DONE at  Feb 5 2025.

> 
> Hi @ColinZou can you take care of the comments and we can merge.

@v1r3n I am really not sure which comment left to be resolved.

nthmost-orkes:
Closing in favor of tracking issue #791. Thank you @ColinZou for identifying this null safety issue!

PR #352, opened 2025-01-04
Reduce FORK/JOIN latency

jmigueprieto:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Currently, a JOIN execution is delayed exponentially up to a maximum of `30s` in the default configuration. This may cause a JOIN to complete with a significant delay compared to the actual completion of all forked tasks.

These changes introduce a configurable postpone calculation (with a less aggressive increase) and a configurable Queue Pop Timeout. By tweaking those values we can reduce the JOIN latency at the start and the end.

SEE also: https://github.com/orkes-io/orkes-queues/pull/19

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
@@ -241,11 +241,11 @@ public class ConductorProperties {
             200; // 10 seconds based on default systemTaskWorkerPollInterval of 50ms
 
     /**
-     * Timeout in milliseconds used by {@link
-     * com.netflix.conductor.core.execution.tasks.SystemTaskWorker} when polling, i.e.: call to
-     * {@link com.netflix.conductor.dao.QueueDAO#pop(String, int, int)}.
+     * Timeout used by {@link com.netflix.conductor.core.execution.tasks.SystemTaskWorker} when
+     * polling, i.e.: call to {@link com.netflix.conductor.dao.QueueDAO#pop(String, int, int)}.
      */
-    private int systemTaskQueuePopTimeout = 200;
+    @DurationUnit(ChronoUnit.MILLIS)
+    private Duration systemTaskQueuePopTimeout = Duration.ofMillis(100);
 
     public String getStack() {
         return stack;
@@ -597,11 +597,11 @@ public class ConductorProperties {
         return systemTaskPostponeThreshold;
     }
 
-    public int getSystemTaskQueuePopTimeout() {
+    public Duration getSystemTaskQueuePopTimeout() {
         return systemTaskQueuePopTimeout;
     }
 
-    public void setSystemTaskQueuePopTimeout(int systemTaskQueuePopTimeout) {
+    public void setSystemTaskQueuePopTimeout(Duration systemTaskQueuePopTimeout) {
         this.systemTaskQueuePopTimeout = systemTaskQueuePopTimeout;
     }
 }
--- a/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
@@ -229,16 +229,15 @@ public class ConductorProperties {
     private int taskExecLogSizeLimit = 10;
 
     /**
-     * This threshold defines the default number of executions after which SystemTasks implementing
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -119,10 +119,13 @@ public boolean execute(
     @Override
     public Optional<Long> getEvaluationOffset(TaskModel taskModel, long defaultOffset) {
         int index = taskModel.getPollCount() > 0 ? taskModel.getPollCount() - 1 : 0;
-        if (index == 0) {
+        // Assuming SystemTaskWorker 50ms pollInterval this will cause the join to be evaluated
+        // continuously during the first 5 seconds and the FORK/JOIN will end without any delay.
+        if (index <= 100) {
```
jmigueprieto, line 124: To reduce the delay of the JOIN task when all forked tasks complete, poll continuously for reasonable amount of time (useful for fork/joins that are meant to finish in less than N seconds).

Then gradually increase the offset which introduces the delay in the JOIN task evaluation - (might end up changing the exponential backoff and will make the N configurable)

rest/src/main/java/com/netflix/conductor/rest/controllers/TaskResource.java
```diff
@@ -77,10 +77,10 @@ public ResponseEntity<List<Task>> batchPoll(
             @RequestParam(value = "workerid", required = false) String workerId,
             @RequestParam(value = "domain", required = false) String domain,
             @RequestParam(value = "count", defaultValue = "1") int count,
-            @RequestParam(value = "timeout", defaultValue = "100") int timeout) {
+            @RequestParam(value = "timeout", defaultValue = "0") int timeout) {
         // for backwards compatibility with 2.x client which expects a 204 when no Task is found
-        return Optional.ofNullable(
-                        taskService.batchPoll(taskType, workerId, domain, count, timeout))
+        // "long polling" is not working as expected
+        return Optional.ofNullable(taskService.batchPoll(taskType, workerId, domain, count, 0))
```
jmigueprieto, line 83: `orkes-queues` requires changes and we need to fix this "long-polling" behaviour.

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
jmigueprieto: Or we make JOIN sync again (?)

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -117,12 +124,17 @@ public boolean execute(
     }
 
     @Override
-    public Optional<Long> getEvaluationOffset(TaskModel taskModel, long defaultOffset) {
-        int index = taskModel.getPollCount() > 0 ? taskModel.getPollCount() - 1 : 0;
-        if (index == 0) {
+    public Optional<Long> getEvaluationOffset(TaskModel taskModel, long maxOffset) {
+        int pollCount = taskModel.getPollCount();
```
jmigueprieto, line 128: To reduce the delay of the JOIN task when all forked tasks complete it will not be postponed until `pollCount` is above a given configurable threshold. 

Once `pollCount` is greater than that threshold, the delay will increase until reaching `maxOffset`.

core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
```diff
@@ -226,6 +228,25 @@ public class ConductorProperties {
...
+     */
+    private int systemTaskPostponeThreshold =
+            200; // 10 seconds based on default systemTaskWorkerPollInterval of 50ms
+
+    /**
+     * Timeout in milliseconds used by {@link
+     * com.netflix.conductor.core.execution.tasks.SystemTaskWorker} when polling, i.e.: call to
+     * {@link com.netflix.conductor.dao.QueueDAO#pop(String, int, int)}.
+     */
+    private int systemTaskQueuePopTimeout = 200;
```
jmigueprieto, line 248: Will change this type to `Duration`

v1r3n (approved):

PR #360, opened 2025-01-09
Fix task def name npe

shaileshpadave:
Pull Request type
----
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
About the taskDef npe exception for http tasks!
When task definition is not provided and we are trying to use inputSchema in task, then during deserialisation, name will be set to null.  We never checked for taskDef name before so never faced issue.

Fix:
----
Check if taskdef.getName() is null then set taskName.

Issue #

Alternatives considered
----

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
@@ -530,10 +530,10 @@ public class WorkflowTask {
     }
 
     /**
-     * Sets the TaskDef for this instance. If the passed TaskDef does not have a name,
-     * this method will set it to the name of the current instance.
+     * Sets the TaskDef for this instance. If the passed TaskDef does not have a name, this method
+     * will set it to the name of the current instance.
      *
-     * NOTE: This method mutates the passed TaskDef object.
+     * <p>NOTE: This method mutates the passed TaskDef object.
      *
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
```diff
@@ -533,6 +533,9 @@ public TaskDef getTaskDefinition() {
      * @param taskDefinition Task definition
      */
     public void setTaskDefinition(TaskDef taskDefinition) {
+        if (taskDefinition != null && taskDefinition.getName() == null) {
+            taskDefinition.setName(this.name);
```
jmigueprieto, line 537: I would not expect a `setX(X arg)` to mutate the passed parameter. It’s not ideal to introduce side effects to a passed parameter. 

Either avoid the side effect, document it clearly or create a copy.

jmigueprieto (changes requested):
- Please fix the description format. As it is, it's difficult to read. Remove the placeholder text.

Screenshot 2025-01-09 at 11 18 14

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java
v1r3n: @shaileshpadave can you check where is this being called from and the caller should fix this.

PR #381, opened 2025-02-09
Added support to bind workflow instance id to parameter for worker task in SDK

l0lcat:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Added annotation WorkflowInstanceIdInputParam to the SDK. This can be used on a String parameter of a WorkerTask method. The parameter will be assigned the value of the workflowInstanceId if it is annotated with the WorkflowInstanceIdInputParam. 

The workflowInstanceId is useful for persisting with entities in a data store and for logging purposes. The alternative is to have a Task type parameter and get it from there, but it cannot be combined with the InputParam annotation. It's also more elegant to use an annotation.

Alternatives considered
----


l0lcat:
In the presented implementation the workflow instance id is bound to a String parameter. Perhaps it's better and nicer to bind it to a UUID parameter. 

Even both could be supported: if the parameter is type String, just bind it. If the parameter type is UUID, parse it and bind it.

Let me know what you think!

l0lcat:
Removed the support of UUID parameter for assigning the workflowId based on feedback (we can't be sure it wil always be a UUID).

v1r3n (approved):

Issue #369, opened 2025-01-23
batchPoll with count=1  in TaskServiceImpel return many tasks


shlomibd:
**Describe the bug**
I am using client 4.0.4 with TaskRunner - it is requesting 1 task but from time to time it gets more than 1 task and it crash the client.

**Details**
Conductor version: 3.21.11
Persistence implementation: Postgres
Queue implementation: Postgres
Lock: ?
Workflow definition:
Task definition:
Event handler definition:

**To Reproduce**
just using client with 1 task count
the workflow has 1000 dynamic forked subworkflows with the task being polled and another task

![Image](https://github.com/user-attachments/assets/796b43df-b0d6-4690-869d-f28ce91734a0)

subworkflow:

![Image](https://github.com/user-attachments/assets/36b57b38-0d39-4260-a837-ee8d3e0896f3)

**Expected behavior**

**Screenshots**

**Additional context**
logs from the client:
2025-01-23 08:16:14.604 TRACE [TaskRunner 1] [] c.n.c.c.a.TaskRunner : Polling task of type: task.aluma.dferip.v2 in domain: 'null' with size 1
2025-01-23 08:16:14.604 TRACE [TaskRunner 1] [] c.n.c.c.a.TaskRunner : in memory queue size for tasks: 0  
2025-01-23 08:16:14.604 DEBUG [TaskRunner 1] [] c.n.c.c.w.Worker : Setting worker id to worker-rip-deployment-996bbc489-fhlrz  
2025-01-23 08:16:14.604 DEBUG [TaskRunner 1] [] c.n.c.c.a.TaskRunner : poll task.aluma.dferip.v2 in the domain null with batch size 1    
2025-01-23 08:16:23.864 ERROR [TaskRunner 1] [] c.n.c.c.a.TaskRunner : null                                                                                                                                                       java.lang.IllegalArgumentException: null                                                                                                                                                                                          
[...]

PR #391, opened 2025-02-12
Issue 369 - postgres query exceeding the limit

shlomibd:
Pull Request type
----
- [X ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
change the query to always get the limit, and throw exception if return messages is more than the requested count

Issue #369 
new query is honoring the limit in the query, it is also limiting the popped messages twice so even if the select part is returning more than the requested count, the query will update only the count out of it.
it is also throwing if in any reason it will return more than the requested count - so that the clients won't fail


v1r3n:
@shlomibd can you update your branch and re-submit the PR? It seems that the github workflow actions are deprecated and  failing the builds.

Robban1980:
I see there is an issue here that can be reproduced with the queue_publish Workflow Event Listener when no event handlers are configured to process the events.

However, this PR does not fully address the described issue. The current query still returns too many results if there are multiple entries with the same message_id, which is valid across multiple queues.

image

The main issue lies in the UPDATE statement—it can update the wrong or too many entries because its WHERE clause does not align with the conditions in the subquery, the current implementation, or the CTE used in this PR.

To ensure only a single correct entry is updated, here is an example query that achieves that:

image

The SQL for the above query:
```sql
BEGIN TRANSACTION;
WITH to_be_processed AS (
    SELECT message_id
    FROM queue_message
    WHERE queue_name = '_callbackFailureQueue'
      AND popped = false
      AND deliver_on <= (current_timestamp + (1000 || ' microseconds')::interval)
    ORDER BY priority DESC, deliver_on, created_on
    LIMIT 1
    FOR UPDATE SKIP LOCKED
)
UPDATE queue_message qm
SET popped = true
FROM to_be_processed
WHERE qm.message_id = to_be_processed.message_id
  AND qm.queue_name = '_callbackFailureQueue'
  AND qm.popped = false
  AND qm.deliver_on <= (current_timestamp + (1000 || ' microseconds')::interval)
RETURNING qm.message_id, qm.priority, qm.payload;
```

shlomibd:
Resubmitted the PR - https://github.com/conductor-oss/conductor/pull/421

PR #402, opened 2025-02-20
Fix sweeper locking: expose both a decide with lock and without lock

VerstraeteBert:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----
HELP NEEDED: where should I best add a test for this fix?

Currently the `WorkflowSweeper` is a no-op, This issues causes workflows not to advance in case of an incomplete `updateTask` or when using subworkflows.

The current `sweeper` implementation takes a lock on the workflow level before calling `decide`. Within the decide method, another lock is attempted on the same key, which is then already taken, and the method returns without doing anything. 

In this PR, another method  `decideWithoutLock` is added to the `WorkflowExecutor`. The sweeper uses this new method, all other old callers use the decide with explicit locking.


VerstraeteBert:
We've come up with a better solution now that fixes locking in more places. We'll make a new PR that includes this new pattern.

PR #417, opened 2025-03-13
added example for python as inline script

pyarun:
Pull Request type
----
- [x] Other (please describe):
Update documentation for inline task to add examples for python scripting.

Changes in this PR
----
No code changes. only documentation updated.
Document was missing any reference about support for python as scripting for inline tasks.

Issue #
Documentation do not state that python is available as scripting language for inline tasks.

Alternatives considered
----
NA


v1r3n:
@pyarun can you resolve conflicts?

PR #420, opened 2025-03-17
[Feature] Adding support for Kafka in the Workflow Event Listener

Robban1980:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
This PR adds the new Workflow Event Publisher `kafka`. It gives support to publish workflow events to Kafka. In addition to the existing events, Completed, Terminated and Finalized the following new events have been added:
- STARTED
- RERAN
- RETRIED
- PAUSED
- RESUMED
- RESTARTED
- COMPLETED (supported by queue_publisher)
- TERMINATED (supported by queue_publisher)
- FINALIZED (supported by queue_publisher)

That addition of these new events does not change the behavior of the existing publishers, they are unchanged.

A new type `WorkflowSummaryExtended` was added that extends the WorkflowSummary, it maintains input and output properties as objects when publishing objects to Kafka.

<strike>Moved rising of Workflow `FINALIZED` event from `cancelNonTerminalTasks` to `endExecution` in the `WorkflowExecutionOps.java`  to make sure it is only raised once. Currently when using a Terminate Task in a workflow it is razed twice.</strike>

To ensure that `FINALIZED` is raised to correct number of times, an overloaded `cancelNonTerminalTasks` was introduced to control if this event should be raised or not, by default it is raised. This is because the `cancelNonTerminalTasks` is called twice when `endExecution` is part of the flow. When `cancelNonTerminalTasks` is called from `endExecution` it should not raise the event again.

Issue #

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/run/WorkflowSummaryExtended.java
+++ b/common/src/main/java/com/netflix/conductor/common/run/WorkflowSummaryExtended.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2025 Conductor Authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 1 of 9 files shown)


Robban1980:
Sent back to draft as i found that the does not work for the case where the workflow is terminated through the API/UI does not raise the Finalize event.

Robban1980:
finalized is now raised the correct amount of times.

bradyyie (approved):
Looks good - code is clean and allows the statuses to be pushed back to kafka as stated

v1r3n (approved):

PR #463, opened 2025-04-08
Update SDK POJOs to have Lombok annotations

orkes-harshil:
Description : 
This PR moves all  java sdk POJOs to lombok annotations. 
We will have
@Data
@Builder
@NoArgsConstructor
@AllArgsConstructor

Our goal is to not break current backward compatibility. We have added unit tests in a previous [PR](https://github.com/conductor-oss/conductor/pull/479/files) for all the accessors and constructors which are deleted as part of Lombokisation.

This will be a base on which we will sync sdk POJOs with server POJOs.

### DIFF
--- a/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/TagObject.java
+++ b/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/TagObject.java
@@ -16,6 +16,7 @@ import lombok.*;
 
 @EqualsAndHashCode
 @ToString
+@Data
 @Builder
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 @NoArgsConstructor
--- a/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/TagString.java
+++ b/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/TagString.java
@@ -16,6 +16,7 @@ import lombok.*;
 
 @EqualsAndHashCode
 @ToString
+@Data
 @Builder
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 @NoArgsConstructor
--- a/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/integration/IntegrationApiUpdate.java
+++ b/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/integration/IntegrationApiUpdate.java
@@ -14,8 +14,6 @@ package io.orkes.conductor.client.model.integration;
 
 import java.util.Map;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.*;
 
 @Data
@@ -27,29 +25,5 @@ public class IntegrationApiUpdate {
     private Map<String, Object> configuration;
     private String description;
     private Boolean enabled;
-    private Long maxTokens;
-    private Frequency frequency;
-
-    @Getter
-    public enum Frequency {
-        DAILY("daily"), WEEKLY("weekly"), MONTHLY("monthly");
-
-        @JsonProperty
-        private final String value;
-
-        Frequency(String value) {
-            this.value = value;
-        }
-
-        @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
-        public static Frequency fromValue(String value) {
-            for (Frequency frequency : values()) {
-                if (frequency.getValue().equalsIgnoreCase(value)) {
-                    return frequency;
-                }
-            }
-            throw new IllegalArgumentException("Unknown frequency: " + value);
-        }
-    }
 
 }
--- a/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/integration/ai/LLMResponse.java
+++ b/conductor-clients/java/conductor-java-sdk/orkes-client/src/main/java/io/orkes/conductor/client/model/integration/ai/LLMResponse.java
@@ -22,46 +22,4 @@ public class LLMResponse {
     private Object result;
     private String finishReason;
     private int tokenUsed;
-
-    public static LLMResponseBuilder builder() {
-        return new LLMResponseBuilder();
-    }
-
-    public static class LLMResponseBuilder {
-        private Object result;
-        private String finishReason;
-        private int tokenUsed;
-
-        private LLMResponseBuilder() {
-        }
-
-        public LLMResponseBuilder result(Object result) {
-            this.result = result;
-            return this;
-        }
-
-        public LLMResponseBuilder finishReason(String finishReason) {
-            this.finishReason = finishReason;
-            return this;
-        }
-
-        public LLMResponseBuilder tokenUsed(int tokenUsed) {
-            this.tokenUsed = tokenUsed;
-            return this;
-        }
-
-        public LLMResponse build() {
-            LLMResponse llmResponse = new LLMResponse();
-            llmResponse.result = this.result;
-            llmResponse.finishReason = this.finishReason;
-            llmResponse.tokenUsed = this.tokenUsed;
-            return llmResponse;
-        }
-
-        public String toString() {
-            return "LLMResponse.LLMResponseBuilder(result=" + this.result +
-                    ", finishReason=" + this.finishReason +
-                    ", tokenUsed=" + this.tokenUsed + ")";
-        }
-    }
 }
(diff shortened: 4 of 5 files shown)


conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/events/EventExecution.java
```diff
@@ -41,126 +47,6 @@ public enum Status {
 
     private Map<String, Object> output = new HashMap<>();
 
-    public EventExecution() {
-    }
-
-    public EventExecution(String id, String messageId) {
```
jmigueprieto, line 47: Getting rid of a public constructor is a breaking change. If anyone is using it, it's going to break.

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
```diff
@@ -158,547 +161,6 @@ public boolean isRetriable() {
...
-     * @param updateTime the updateTime to set
-     */
-    public void setUpdateTime(long updateTime) {
-        this.updateTime = updateTime;
-    }
-
-    /**
-     * @return the queueWaitTime
-     */
-    public long getQueueWaitTime() {
```
jmigueprieto, line 344: This is not a simple getter, we are losing this logic.

conductor-clients/java/conductor-java-sdk/tests/src/test/java/io/orkes/conductor/client/http/SchedulerClientTests.java
```diff
@@ -78,21 +78,23 @@ void testDebugMethods() {
     @Test
     @DisplayName("It should set the timezone to Europe/Madrid")
     void testTimeZoneId() {
-        var schedule = new SaveScheduleRequest()
+        var schedule = SaveScheduleRequest.builder()
```
jmigueprieto, line 81: These are breaking changes.

jmigueprieto:
This PR contains too many breaking and potentially breaking changes. Shouldn't be merged.

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/events/EventExecution.java
```diff
@@ -41,126 +47,6 @@ public enum Status {
...
-    }
-
-    /**
-     * @param output the output to set
-     */
-    public void setOutput(Map<String, Object> output) {
-        this.output = output;
-    }
-
     public boolean equals(Object o) {
```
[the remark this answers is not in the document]
RamOrkes: it seems Lombok won't generate those if the class already has them.. but .. 

⚠️ Gotcha to watch for:
If your hashCode() includes fields that were forcibly initialized to default values (null, 0, etc.), then you might get unexpected or inconsistent hash codes until those fields are explicitly set.

So while hashCode() still works, the object's initial hash may be based on incomplete/default data.

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -288,134 +135,12 @@ public Integer getRateLimitFrequencyInSeconds() {
         return rateLimitFrequencyInSeconds == null ? 1 : rateLimitFrequencyInSeconds;
     }
 
-    /**
```
jmigueprieto, line 291: documentation is lost

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/events/EventExecution.java
orkes-harshil: @NoArgsConstructor will generate a constructor with no parameters. If this is not possible (because of final fields), a compiler error will result instead, unless @NoArgsConstructor(force = true) is used, then all final fields are initialized with 0 / false / null.
from the official documentation of Lombok.

I am writing tests for all the removed methods. it should prove that there are no breaking changes.
here's the plan : https://orkesinternal.slack.com/archives/C05DZTWV6LA/p1744266011764189?thread_ts=1744221117.794119&cid=C05DZTWV6LA

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/model/BulkResponse.java
```diff
@@ -12,42 +12,29 @@
...
      * Key - entityId Value - error message processing this entity
      */
-    private final Map<String, String> bulkErrorResults;
+    private Map<String, String> bulkErrorResults;
 
-    private final List<T> bulkSuccessfulResults;
+    private List<T> bulkSuccessfulResults;
 
-    private final String message = "Bulk Request has been processed.";
+    private String message = "Bulk Request has been processed.";
```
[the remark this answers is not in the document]
orkes-harshil: Why do we want this field to be immutable ?
Why do we want to have special treatment ?

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/Auditable.java
```diff
@@ -12,6 +12,13 @@
  */
 package com.netflix.conductor.common.metadata;
 
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
```
jmigueprieto, line 21: `@AllArgsConstructor` is just a bad idea. Add a field and you've broken backwards compatibility

orkes-harshil: Makes sense. We should remove @AllArgsConstructor, and let users use empty constructor + setters or builder. This approach will not have future backward compatibility issues.
Thank you!

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/events/EventExecution.java
orkes-harshil: Here's the test : https://github.com/conductor-oss/conductor/blob/ca49d63a13184b87f80313e489b1fb256f4a6798/conductor-clients/java/conductor-java-sdk/conductor-client/src/test/java/com/netflix/conductor/common/metadata/events/TestEventExecutionPojoMethods.java#L27

This test should resolve this conversation

conductor-clients/java/conductor-java-sdk/tests/src/test/java/io/orkes/conductor/client/http/SchedulerClientTests.java
orkes-harshil: Reverted

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/Auditable.java
orkes-harshil: I have removed all AllArgsConstructors from the code. 
Wherever  they were already there I have replaced them with actual constructor. This will ensure we do not break backward compatibility while making sure we do not introduce future backward compatibility problem.

As a side effect - I've removed Builder annotations from the code and replace with actual builder code. as @Builder does not work without @AllArgsConstructor

orkes-harshil: ref : https://orkesinternal.slack.com/archives/C05DZTWV6LA/p1744782380194339?thread_ts=1744628405.412479&cid=C05DZTWV6LA

to conclude the discussion on @AllArgs and @Builder annotations:
We will have
@Data
@Builder
@NoArgsConstructor
@AllArgsConstructor
Our goal is to not break current backward compatibility. @AllArgs will break backward compatibility in future but we are okay with that. Users should move to @Builder pattern to resolve future issues.

I'm resolving this conversation

conductor-clients/java/conductor-java-sdk/conductor-client/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
orkes-harshil: added at field level.

v1r3n (approved):

Issue #482, opened 2025-04-16
[FEATURE]: Abstraction of time so that time can be manipulated in tests


mk-verma:
## Describe the Feature Request
We have a use case where we want to manipulate time during testing. Currently, conductor uses native methods (`System.currentTimeMillis`) to get the time at different places. This makes testing hard incase you want to manipulate time.
We want an ability in conductor to have a abstracted service that manages the time, so that we can use that service to modify the time as per the need during testing. 

## Describe Preferred Solution
A service having static methods to return current time.

PR #483, opened 2025-04-16
Added a class to abstract time to a centralised place to improve testing.

mk-verma:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md
- [x] Other (please describe): Improve testing.

Changes in this PR
----
We have a use case where we need to change time during execution of integration tests. This is an attempt to abstract out the time in conductor to a single place, so that during testing we can manipulate the time as per the need.

Issue #482
This PR adds a service to fetch current time in conductor via static methods. These methods can be overriden in tests to manipulate time. 

Alternatives considered
I also thought of using Bean (via @Bean config) to return clock instance or having TimeService as a `Component` but that would then require injecting the clock or TimeService into different classes which fetch current time. This breaks the backward compatibility and hence rejected those options.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/TimeService.java
+++ b/core/src/main/java/com/netflix/conductor/service/TimeService.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2025 Conductor Authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.service;
+
+import java.time.Clock;
+
+/**
+ * Centralised time service to get the current time. This is aimed to abstract the time source and
+ * allow for interacting with time during testing.
+ */
+public class TimeService {
+    public static boolean useCustomClockForTests = false;
+    public static Clock customClockForTests = Clock.systemDefaultZone();
+
+    public static long currentTimeMillis() {
+        if (!useCustomClockForTests) {
+            return TimeService.currentTimeMillis();
+        } else {
+            return customClockForTests.millis();
+        }
+    }
+}
--- a/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
@@ -47,6 +47,7 @@ import com.netflix.conductor.dao.*;
 import com.netflix.conductor.metrics.Monitors;
 import com.netflix.conductor.model.TaskModel;
 import com.netflix.conductor.model.WorkflowModel;
+import com.netflix.conductor.service.TimeService;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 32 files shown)


AWOrpington (changes requested):
You're on the right track with removing System.currentTimeMillis, but it should be replaced directly with Java.time.Clock which then can be manipulated directly in tests.

core/src/main/java/com/netflix/conductor/service/TimeService.java
```diff
@@ -0,0 +1,32 @@
...
+package com.netflix.conductor.service;
+
+import java.time.Clock;
+
+/**
+ * Centralised time service to get the current time. This is aimed to abstract the time source and
+ * allow for interacting with time during testing.
+ */
+public class TimeService {
+    public static boolean useCustomClockForTests = false;
```
AWOrpington, line 22: Generally classes shouldn't know about _how_ they're used, especially with relation to testing.

c4lm:
There are also places in Conductor which invoke Instant.now(), LocalDate.now() etc.
Could you please describe in more detail what kind of tests you are trying to run?

mk-verma:
> There are also places in Conductor which invoke Instant.now(), LocalDate.now() etc.

Yeah, I missed it actually 😓 . I will cover them as well

> Could you please describe in more detail what kind of tests you are trying to run?
Actually, we use conductor-oss as self-hosted orchestration engine for our service. We want to write test where we can change time. For example: We have a use case where we want to wait for outside event to arrive, now this event can happen let's say 1 day later or 10 days later. We can't write a test which waits for that long. Also, we do not want to work under the assumption that our system work correctly if it works correctly for small time. We do not know how the system behaves if lets say a workflow is in progress and waiting for an event to come at let's say 10 day later or event lets say a month later. Will the underlying system that is conductor or queues which we use or some other dependency, has some mechanism which discards that workflow or task if in running state for too long.
We want to write tests to verify this assumption. Hence, we need a mechanism to may be change or increase the time tests so that we can simulate long time in tests.
I hope this makes sense.

mk-verma:
> You're on the right track with removing System.currentTimeMillis, but it should be replaced directly with Java.time.Clock which then can be manipulated directly in tests.

Yes, but there are some problem to that.
The first approach that I thought was using @Bean which returns the Clock. In production, this returns normal clock. In tests ,it can return mocked clock. But we need to be injected in constructor of each class where ever we need time. This poses backward compatibility issues as some of them (classes) are extendible by the services that use conductor.
For example `WorkflowExecutorOps` uses time and we need to modify the current constructor of add that clock bean. If we do so then the services which have their own  `WorkflowExecutorOps` will break on version bump. If we create another constructor with clock, we need to somehow manage the Injection.
I am not sure about @Autowired on field/setter though. Let me know your thoughts.

v1r3n:
I think overall this is a good idea.  The approach I think that makes sense here is as the following:
1. Have a @Bean that provides Clock with system default
2. Tests can override this clock provider to inject custom clock, update the ticks etc
3. Replace usage of system.currentTimeInMillis, Instant.now() etc with the clock.

core/src/main/java/com/netflix/conductor/service/TimeService.java
```diff
@@ -0,0 +1,32 @@
...
+    public static Clock customClockForTests = Clock.systemDefaultZone();
+
+    public static long currentTimeMillis() {
+        if (!useCustomClockForTests) {
+            return TimeService.currentTimeMillis();
```
kirked, line 27: Looks like an infinite loop, or one waiting to happen (calling this same method).

core/src/main/java/com/netflix/conductor/service/TimeService.java
```diff
@@ -0,0 +1,32 @@
...
+        } else {
+            return customClockForTests.millis();
+        }
+    }
+}
```
kirked, line 32: better might be:
```java

import java.time.Clock;
import java.util.concurrent.atomic.AtomicReference;

public class TimeService {
    public static AtomicReference<Clock> clock = new AtomicReference<>(Clock.systemDefaultZone());

    public static long currentTimeMillis() {
        return clock.get().millis();
    }
}
```
tests can then do this in setups:
```java
TimeService.clock.set(customTestingClock);
```

PR #506, opened 2025-05-13
Fix dependency issue

Sudakatux:
Pull Request type
----
- [X ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] WHOSUSING.md
- [ ] Other (please describe):

**NOTE**: Please remember to run `./gradlew spotlessApply` to fix any format violations.

Changes in this PR
Fixes a dependency issue which caused "Could not initialize class org.opensearch.Version"
----

_Describe the new behavior from this PR, and why it's needed_
Issue #505

Alternatives considered
----

_Describe alternative implementation you have considered_

Issue #505, opened 2025-05-12
Could not initialize class org.opensearch.Version


vuon9:
**Describe the bug**

I tried to see how the Redis/Open Search could work but got some 500 errors from search API.
But I dont see the same errors if I use `docker/docker-compose.yaml`

**Details**
Conductor version: latest (main)
Persistence implementation: Redis Standalone
Queue implementation: Redis Standalone
Lock: Redis
Workflow definition: default
Task definition: default
Event handler definition: default
Docker app: OrbStack

**To Reproduce**
Use main branch, and this command:
```bash
docker compose -f docker/docker-compose-redis-os.yaml up -d

 ✔ conductor-server                         Built                                                                                                                                                       0.0s 
 ✔ Network docker_internal                  Created                                                                                                                                                     0.0s 
 ✔ Volume "docker_osdata-conductor"         Created                                                                                                                                                     0.0s 
 ✔ Container docker-conductor-opensearch-1  Healthy                                                                                                                                                    10.8s 
 ✔ Container docker-conductor-redis-1       Healthy                                                                                                                                                    30.8s 
 ✔ Container conductor-server               Started     
```

And I got some errors like this below

![Image](https://github.com/user-attachments/assets/dae50c14-845f-4aa0-af20-00fed666823b)

Here is an cURL
```bash
curl 'http://conductor-server.docker.orb.local:5000/api/workflow/search?start=0&size=15&sort=startTime%3ADESC&freeText=&query=startTime%3E1746971197369%20AND%20status%20IN%20%28FAILED%2CTERMINATED%2CTIMED_OUT%2CPAUSED%29' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US,en;q=0.9,vi;q=0.8' \
  -H 'Connection: keep-alive' \
  -H 'DNT: 1' \
  -H 'Referer: http://conductor-server.docker.orb.local:5000/' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36' \
  --insecure
```

Its response:
```json
{
    "status": 500,
    "message": "Handler dispatch failed: java.lang.NoClassDefFoundError: Could not initialize class org.opensearch.Version",
    "instance": "ee873df7d589",
    "retryable": false
}
```

server container logs:
```
[...]

PR #527, opened 2025-06-14
CCOR-12529 Create dynamic lucene dependency selection based on index engine

kgoeltner:
Calling `api/.../search` in the conductor-oss ui searching for workflows or task executions fails. The conductor-server throw a `NoSuchFieldError: LUCENE_7_0_0` due to incompatible lucene versions on the classpath.

**Server logs:**
Screenshot 2025-06-13 at 8 10 06 PM

**UI logs:**
Screenshot 2025-06-13 at 7 35 15 PM

Pull Request type
----
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

**History:**
- [ISSUE-505](https://github.com/conductor-oss/conductor/issues/505): original incident issue from OSS user
- [PR-506](https://github.com/conductor-oss/conductor/pull/506/files): removed the lucene dependency entirely since OS provides its own. This fixed the api/../search issue failing for executions on docker-containers using OS. Issue was still present for ES setups.
- [PR-524](https://github.com/conductor-oss/conductor/pull/524/files): updated the lucene dependency from 8.11.1 to 10.2.1, but this is incompatible with ES 7.17.11

**Cause:**
- ElasticSearch (ES 7.17.11) and OpenSearch (OS 2.18.0) which are used for indexing the db both internally depend on lucene
- ES 7.17.11 uses lucene 8.11.1, while OS 2.18.0 uses lucene 9.12.0. Using one lucene version over the other causes runtime conflicts in the build
- The lucene dependencies previously taken out from es-7-persistence and server are needed for build

**Fix:**
- Primarily support ES with lucene 8.11.1, with instructions written to support OS with lucene 9.12.0

Alternatives considered
----
- Upgrading ES to use the same lucene v9.12.0 as OS: would require going past ES v7 to v8 which would introduce breaking changes and deprecation i.e. `elasticsearch-rest-high-level-client`
- Introducing a `SEARCH_ENGINE` build arg to conditionally resolve the correct lucene version based on the target indexing search engine would support both ES & OS but will produce a compile dependency leading to downstream issues

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -68,6 +68,5 @@ ext {
     revNats = '2.16.14'
     revStan = '2.2.3'
     revFlyway = '10.15.2'
-    revApacheLucene = '8.11.1'  // Change this to 9.12.0 for Opensearch support
     revConductorClient = '4.0.10'
 }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -68,6 +68,6 @@ ext {
     revNats = '2.16.14'
     revStan = '2.2.3'
     revFlyway = '10.15.2'
-    revApacheLucene = (System.getenv('SEARCH_ENGINE') == 'OS') ? '9.12.0' : '8.11.1'
+    revApacheLucene = '8.11.1'  // Change this to 9.12.0 for Opensearch support
     revConductorClient = '4.0.10'
 }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -68,6 +68,6 @@ ext {
     revNats = '2.16.14'
     revStan = '2.2.3'
     revFlyway = '10.15.2'
-    revApacheLucene = '10.2.1'
+    revApacheLucene = (System.getenv('SEARCH_ENGINE') == 'OS') ? '9.12.0' : '8.11.1'
     revConductorClient = '4.0.10'
 }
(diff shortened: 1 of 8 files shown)


dependencies.gradle
```diff
@@ -68,6 +68,6 @@ ext {
     revNats = '2.16.14'
     revStan = '2.2.3'
     revFlyway = '10.15.2'
-    revApacheLucene = '10.2.1'
+    revApacheLucene = (System.getenv('SEARCH_ENGINE') == 'OS') ? '9.12.0' : '8.11.1'
```
v1r3n: we should add separate dependency and module ofr open search instead of this.

kgoeltner: Elasticsearch (ES) and OpenSearch (OS) use different versions of lucene — **8.11.1** and **9.12.0** respectively. Currently, when we build the project, both ES and OS are included, and each transitively pulls in its own lucene dependency. This leads to a version conflict, as lucene is not designed to support multiple versions on the classpath. As a result, the higher version (**9.12.0** from OS) takes precedence by gradle, which causes compatibility issues with ES.

I attempted to exclude the explicit lucene import entirely, but the conflict persists since both ES and OS transitively import lucene through their respective dependencies.

Should I try and isolate ES & OS and not have them [both be imported in the server module](https://github.com/conductor-oss/conductor/blob/main/server/build.gradle#L44-L45) simultaneously? The goal is to avoid pulling in conflicting lucene versions during server build.

kgoeltner: Offline convo: provide support for ES or OS not both due to conflicting lucene versions.

v1r3n (approved):

v1r3n (approved):

PR #535, opened 2025-06-24
Added cache config for workflow

vishesh-orkes:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
Added cache config that would be used to cache the workflow

This is needed to cache the workflow based on the cache config set.

Alternatives considered
----

### DIFF
--- a/grpc/src/main/proto/model/cacheconfig.proto
+++ b/grpc/src/main/proto/model/cacheconfig.proto
@@ -1,0 +1,12 @@
+syntax = "proto3";
+package conductor.proto;
+
+
+option java_package = "com.netflix.conductor.proto";
+option java_outer_classname = "CacheConfigPb";
+option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/model";
+
+message CacheConfig {
+    string key = 1;
+    int32 ttl_in_second = 2;
+}
--- a/common/src/main/java/com/netflix/conductor/common/metadata/workflow/CacheConfig.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/workflow/CacheConfig.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 Conductor Authors.
+ * <p>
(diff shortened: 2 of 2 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
```diff
@@ -37,6 +37,32 @@ public enum TimeoutPolicy {
         ALERT_ONLY
     }
 
+    @ProtoMessage
+    public static class CacheConfig {
```
v1r3n: can you consolidate with 
https://github.com/conductor-oss/conductor/blob/acd6120b505bebf8536505cd97192b4eaadc9667/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java#L42?

grpc/src/main/proto/model/cacheconfig.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package conductor.proto;
+
+
+option java_package = "com.netflix.conductor.proto";
+option java_outer_classname = "CacheConfigPb";
+option go_package = "github.com/netflix/conductor/client/gogrpc/conductor/model";
+
+message CacheConfig {
+    string key = 1;
```
manan164, line 10: Do we need value here?

manan164 (approved):

grpc/src/main/proto/model/cacheconfig.proto
vishesh-orkes: no

common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowDef.java
vishesh-orkes: Done. Created a new CacheConfig and used it in both places

PR #569, opened 2025-08-08
Added support for the dynamic rate limiting using StartWorkflowRequest

hchepey-clari:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Implemented a runtime-configurable rate limiting feature that allows per-workflow-instance rate limit customization while maintaining full backward compatibility.

Issue #

We have a requirement where we want to rate limit the tasks as per the different client rate limit applicable. That is dynamic based on the Workflow Inputs supplied to the Workflow.

When starting a workflow, users can provide optional taskRateLimitOverrides in the request
These overrides are keyed by reference task name (most specific) or task definition name
Task mappers check for overrides before falling back to static TaskDef values
The existing Redis rate limiting infrastructure works unchanged with the dynamic values

Alternatives considered
----


v1r3n:
Have you considered adding TaskDef with rate limits in the workflow definition when starting a workflow?

hchepey-clari:
> Have you considered adding TaskDef with rate limits in the workflow definition when starting a workflow?

I just checked. Thanks for pointing that out. Let me think about it.

One thing for sure, that would mean I have to pass the entire workflowdef json in every StartWorkflow Request. And we have to operate on a scale where we have to invoke thousands of workflows every hour at a regular cadence.

hchepey-clari:
> > Have you considered adding TaskDef with rate limits in the workflow definition when starting a workflow?
> 
> I just checked. Thanks for pointing that out. Let me think about it.
> 
> One thing for sure, that would mean I have to pass the entire workflowdef json in every StartWorkflow Request. And we have to operate on a scale where we have to invoke thousands of workflows every hour at a regular cadence.

This change is backward compatible and would not need to pass the WorkflowDef everytime in the StartWorkflowRequest payload. We will be triggering millions of workflows everyday. Do you foresee any other concern with this change, @v1r3n ?

hchepey-clari:
@v1r3n @manan164 - Bumping up in case if this slipped off your radar.

v1r3n:
Passing WorkflowDef is going to be OK -- I do not see any performance issues with that, actually it might overall be a little quicker because it completely avoids DB read to get the workflow metadata.

The purpose makes sense, but I think the change itself is a bit too complex.  We are internally discussing to see if there is a simpler way.  The easiest thing to do here might to be be able to set the parameterized values for task definition.

hchepey-clari:
Any update on this @v1r3n ?

hchepey-clari:
Any update on this @v1r3n ?

hchepey-clari:
@v1r3n @manan164 - Any update on this?

v1r3n:
Hi @hchepey-clari the best way to handle this is via dynamic workflows, which conductor already supports.

hchepey-clari:
Ok. Can you help me understand the issues with the changes in these PR that I might have missed out? May be I am missing the bigger picture or something, I am not sure. Can you help me understand and bridge that gap, please?

v1r3n:
Hi @hchepey-clari do you want to find some time and discuss this over the call?

hchepey-clari:
Yes...

v1r3n:
Lets connect over the slack and we can find time.

nthmost-orkes:
Closing in favor of tracking issue #790. Thank you @hchepey-clari for the dynamic rate limiting implementation!

nthmost-orkes:
@hchepey-clari sorry i missed your reply on the new tracking issue - i've reopened this PR if you want to fix this up.

I closed this one because it had been left for over a year.  changes might be needed beyond the conflicts above so just be prepared for that.

hchepey-clari:
@nthmost-orkes - Conflicts are resolved and PR is ready too.

nthmost-orkes (commented):
Hey @hchepey-clari, thanks for sticking with this one — I can see it's been a long road getting eyes on this PR. The feature makes sense and the `TaskMapperUtils` consolidation is a genuinely clean move. I have a few things that need fixing before this can go in though.

**The one that will silently eat your data:** `WorkflowModel.taskRateLimitOverrides` has no counterpart in the `Workflow` DTO in `common/`. That means if a workflow suspends (WAIT task, human task, anything that causes a round-trip through the DB) the overrides are gone when it resumes. You'll need to add the field to `Workflow` and make sure the DAO persists it.

**Two gRPC bugs:** In `AbstractProtoMapper.toProto()`, `getRateLimitPerFrequency()` returns `Integer` (nullable) but you're passing it straight to a proto setter that takes `int` — NPE if it's null. On the flip side, `fromProto()` will read proto's default `0` as an explicit override value, so a round-trip through gRPC silently converts "not set" into "overridden to 0". Both sides need a null/default check.

**Tiny accident:** The `@Override` annotation got accidentally deleted from `WorkflowModel.toString()` — looks like a casualty of the surrounding edit.

**Namespace:** `TaskMapperUtils` is a brand new class and needs to land in `org.conductoross.*`, not `com.netflix.conductor.*`. That's the project convention for all new code.

A couple of smaller things: the dual-sentinel pattern in `TaskRateLimitOverride` (`NOT_SET = -1` int field *and* a separate boolean flag) is doing the same job twice — you could just use a nullable `Integer` field and drop both sentinels. And worth double-checking that `workflowTask.getName()` is actually the task definition name for system task types — if it's the task type name instead, the by-def-name lookup will silently miss.

Tests look solid for the happy path. Once the persistence bug is fixed, a test for the suspend/resume case would be good to add.

hchepey-clari:
@nthmost-orkes - Addressed the PR comments.

hchepey-clari:
@nthmost-orkes - Is there anything I can help with so that we can get this PR merged ASAP? This has been on hold for too long and I would appreciate if we could get this going across the finish line.

PR #571, opened 2025-08-12
fix(elasticsearch-dao): wait for index refresh and guarantee changes are visible for search

ClemDoum:
…lowID

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
## Bug description

It's hard to reproduce the bug in a consistent manner but `ElasticSearchDAOV6` and `ElasticSearchDAOV7` do not use the `wait_for` refresh policy when performing index/update/delete/bulk requests.

As [described here](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter) this can result in the action not being immediately visible to the user depending on the ES refresh interval.

In my particular case it is very problematic as I start some workflows then poll them until they are complete. Sometimes when polling for the first time the workflow indexing is not visible yet to search hence my function returns because there are not running workflow.

This PR update the `ElasticSearchDAOV7` to use the `wait_for` / `WAIT_UNTIL` policy to that we wait until the indexed data is visible to search before returning, returning that indexed task and workflow will be visible for search.

**Important notes**:
- I made the change only for `indexTask`/`indexWorkflow`,  while I kept the `updateTask`/`updateWorkflow` and `removeTask`/`removeWorkflow` as is. This is a debatable choice some issues could also arise from the deletion/update not being reflected in the search. I my opinion the most problematic use case is returning an ID of an object which not yet exists and yet only fixed the index logic as using the `wait_for` refresh policy add some latency. Please let me know if you think I should also update the delete/update logic
- the change is not easy to test so I haven't implemented any, a valid test could be to intercept the queries made to ES and verify that they are done using the `wait_for` query parameter, let me know if you think that make sense.

### DIFF
--- a/es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchDAOV6.java
+++ b/es6-persistence/src/main/java/com/netflix/conductor/es6/dao/index/ElasticSearchDAOV6.java
@@ -22,6 +22,8 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.elasticsearch.ResourceAlreadyExistsException;
 import org.elasticsearch.action.DocWriteResponse;
 import org.elasticsearch.action.admin.indices.create.CreateIndexRequest;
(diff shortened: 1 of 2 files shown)


v1r3n:
@ClemDoum can you fix the build issues?

ClemDoum:
@v1r3n I think the failure is due to an unrelated flaky test, other builds for other PRs fail with the same error: https://github.com/conductor-oss/conductor/actions/runs/17066138206/job/48690211796

PR #600, opened 2025-10-08
fix(os-persistence): Resolve race condition in OpenSearch bulk indexing

### LINKED ISSUE #592: Bug: Silent Task Indexing Failures with Async Indexing Enabled
**Describe the bug**
When `asyncIndexingEnabled=true`, tasks from multi-task workflows randomly fail to index to the Search DB (here OpenSearch) due to unhandled `CompletableFuture` exceptions, causing data inconsistency between the primary database and search index.

**Details**
Conductor version: Latest (main branch)
Queue implementation: Redis standalone
Lock: Redis
Workflow definition: Any workflow with 2+ tasks
Task definition: Any SIMPLE tasks
Event handler definition: N/A

**To Reproduce**
1. Set `conductor.app.asyncIndexingEnabled=true` in configuration
2. Create a workflow with 2 or more tasks
3. Start and complete the workflow (all tasks complete successfully)
4. Check the search index (OpenSearch/Elasticsearch) - only some tasks are indexed
5. Check PostgreSQL - all tasks exist correctly
6. Repeat with different workflows - random tasks are missing from search index

**Expected behavior**
All completed tasks should be indexed to the Search DB and be searchable via the API and UI.

**Screenshots**

Image
Image

**Additional context**

- Root cause: In `ExecutionDAOFacade.java`, `asyncIndexTask()` CompletableFuture returns are ignored, causing silent failures
- Affects all deployments with async indexing enabled
- Workaround: Set `asyncIndexingEnabled=false` (impacts performance)

**Proposed Solution**
 Replace the current individual async indexing approach with bulk task indexing, following the same pattern already used for task execution logs. This approach would:

   1. Add bulk indexing methods to the `IndexDAO` interface similar to the existing `addTaskExecutionLogs` methods
  2. Implement these methods in all `IndexDAO` implementations using their existing bulk request infrastructure
  3. Update `ExecutionDAOFacade` to collect all tasks and submit them as a single bulk operation instead of individual async calls
  4. Leverage existing error handling, retry logic, and monitoring that's already proven to work with bulk operations

Akhil-Pathivada:
## Problem
Issue #592 reports random task indexing failures where tasks are saved in the DB but fail to index in OpenSearch, causing inconsistency between the Primary DB and Search index.

## Root Cause
A race condition exists in the `OpenSearchRestDAO.indexObject()` method. The critical section (add to buffer → check size → flush) is not **atomic**, allowing concurrent threads to interfere during the flush operation.

**The issue occurs when:**
1. Thread A adds an item to the buffer (size = 1)
2. Thread A checks size (1 >= threshold), begins flushing to OpenSearch
3. **Thread B adds an item to the buffer** (size = 2) **while Thread A is sending data to OpenSearch**
4. Thread A receives success from OpenSearch and **replaces the entire buffer** with a new empty one: `bulkRequests.put(docType, new BulkRequests(...))`
5. **Thread B's item is discarded** when the buffer is replaced
6. Thread B checks size on the **new empty buffer** (size = 0), doesn't flush
7. **Result**: Thread B's item is lost

This is a **Time-Of-Check-Time-Of-Use (TOCTOU)** race condition where the buffer state changes between when Thread B adds its item and when it performs the size check.

**Why indexBatchSize=1 exacerbates this:**  
With the default `indexBatchSize=1`, every item triggers an immediate flush, maximizing the window for concurrent modifications during the OpenSearch network call.

## Solution
Add a `synchronized(this)` block around the critical section in `indexObject()`:

This ensures:
- Only one thread can add to the buffer, check size, and trigger flush at a time
- No thread can add items while another thread is flushing
- Buffer replacement happens atomically without losing items from concurrent threads

## Testing
- Tested with concurrent workflow executions and verified 100% indexing success rate

## Changes
- Modified `OpenSearchRestDAO.indexObject()` method with synchronized block
- No API changes or breaking changes

Fixes #592

### DIFF
--- a/os-persistence/src/main/java/com/netflix/conductor/os/dao/index/OpenSearchRestDAO.java
+++ b/os-persistence/src/main/java/com/netflix/conductor/os/dao/index/OpenSearchRestDAO.java
@@ -1247,14 +1247,17 @@ public class OpenSearchRestDAO extends OpenSearchBaseDAO implements IndexDAO {
         IndexRequest request = new IndexRequest(index);
         request.id(docId).source(docBytes, XContentType.JSON);
 
-        if (bulkRequests.get(docType) == null) {
-            bulkRequests.put(
-                    docType, new BulkRequests(System.currentTimeMillis(), new BulkRequest()));
-        }
+        synchronized (this) {
+            if (bulkRequests.get(docType) == null) {
+                bulkRequests.put(
+                        docType, new BulkRequests(System.currentTimeMillis(), new BulkRequest()));
+            }
 
-        bulkRequests.get(docType).getBulkRequest().add(request);
-        if (bulkRequests.get(docType).getBulkRequest().numberOfActions() >= this.indexBatchSize) {
-            indexBulkRequest(docType);
+            bulkRequests.get(docType).getBulkRequest().add(request);
+            if (bulkRequests.get(docType).getBulkRequest().numberOfActions()
+                    >= this.indexBatchSize) {
+                indexBulkRequest(docType);
+            }
         }
     }
 


Akhil-Pathivada:
@v1r3n @manan164 could you please help reviewing this?

v1r3n (approved):

PR #599, opened 2025-10-08
  Fix FORK_JOIN_DYNAMIC input wrapping when fed directly by JSON_JQ_TRANSFORM (#575)

sc-a-zhukov:
Fix FORK_JOIN_DYNAMIC input wrapping when fed directly by JSON_JQ_TRANSFORM (#575)

  ———

  Pull Request type

  - [x] Bugfix
  - [ ] Feature
  - [ ] Refactoring (no functional changes, no api changes)
  - [ ] Build related changes
  - [ ] WHOSUSING.md
  - [ ] Other (please describe):

  Changes in this PR

- Normalized JSON_JQ_TRANSFORM outputs so both result and resultList are plain Java maps/lists. In reproduction workflow fork_join_dynamic now receives inputs like
```
{
	"commonProperty": "common value",
	"property": "value1",
	"__index": 0
}
```
image

instead of
```
{
	"input": {
		"commonProperty": "common value",
		"property": "value1"
	},
	"__index": 0
}
```
in the version before changes
image

Also this behaviour matching with Orkes Cloud behaviour.

- tests:
      - JsonJqTransformTest checks resultList equals a list of standard maps.
      - ForkJoinDynamicTaskMapperTest asserts dynamic fork inputs remain unwrapped and keep their parameters.
      - Verified manually on the workflow from issue #575. Before fixing the bug, it was reproduced only when JSON_JQ_TRANSFORM went immediately before FORK_JOIN_DYNAMIC; inserting an asynchronous task (WAIT/SIMPLE/...) between them serialized the data and hid the error. 

the screenshot from the version before fix with WAIT task
image

The fix removes this dependency and the need for such workarounds

PR #627, opened 2025-10-22
FIX: correct FORK_JOIN_DYNAMIC return of INLINE task (pop out of "input" wrapper)

### LINKED ISSUE #616: FORK_JOIN_DYNAMIC wraps INLINE task output in extra input layer
# FORK_JOIN_DYNAMIC wraps INLINE task output in extra input layer

## Description
`FORK_JOIN_DYNAMIC` wraps task inputs in an unexpected nested structure when receiving output directly from `INLINE` tasks.

**Expected behavior**: Each forked sub-workflow receives:
```json
{"param1": "value", "param2": "value"}
```

**Actual behavior**: Each forked sub-workflow receives:
```json
{"input": {"param1": "value", "param2": "value"}}
```

## Context
- Original issue #575 reported this behavior for both `JSON_JQ_TRANSFORM` and `INLINE` tasks
- PR #599 fixed the `JSON_JQ_TRANSFORM` case
- The `INLINE` case remains unresolved, as noted by @sc-a-zhukov in #575

**Root cause**: `FORK_JOIN_DYNAMIC` wraps any non-Map fork input. `INLINE.result` is non-Map until serialization/deserialization converts it.

### Behavior Inconsistency
After PR #599, there is now inconsistent behavior between task types:
- `JSON_JQ_TRANSFORM` → `FORK_JOIN_DYNAMIC` works correctly (unwrapped)
- `INLINE` → `FORK_JOIN_DYNAMIC` wraps output under `input` key

This inconsistency makes it difficult for developers to predict behavior and creates confusion about whether this is intended functionality.

### Runtime Errors
Forked sub-workflows fail when expected input parameters are not found:
```
Error: Required input parameter 'param1' not found
Actual input: {"input": {"param1": "value"}}
```

Sub-workflows must either:
- Handle missing parameters (receiving null values)
- Reference parameters as `${workflow.input.input.param1}` (double-nesting)

### Workaround Required
Developers must insert an asynchronous task between `INLINE` and `FORK_JOIN_DYNAMIC` to trigger serialization that unwraps the structure. This adds unnecessary complexity to workflow definitions.

## Steps to Reproduce
1. Create an `INLINE` task that outputs a map structure
2. Configure `FORK_JOIN_DYNAMIC` to use the `INLINE` task output as `forkTaskInputs`
3. Observe that forked tasks receive `{"input": {<actual data>}}` instead of `{<actual data>}`

## Proposed Solution
Apply the same fix from PR #599 to handle `INLINE` task outputs in `ForkJoinDynamicTaskMapper.java`.

## Related
- Original issue: #575
- Fix for JSON_JQ_TRANSFORM: #599

nthmost-orkes:
Pull Request type
----
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

WIP (keep moving, working on this live)

Fixes #616  (fully described in that issue) and picks up where this PR left off:  https://github.com/conductor-oss/conductor/pull/599

Approach here is to normalize in Inline.java to match output of other returns.  This normalizes for all INLINE tasks (only) and matches what we do in other places.

Alternatives considered
----

Yeah, we could instead change ScriptEvaluator.eval() -- 

```
  // ScriptEvaluator.java
  public static Object eval(String script, Object input) throws ScriptException {
      initEngine(false);
      Bindings bindings = engine.createBindings();
      bindings.put("$", input);
      Object result = engine.eval(script, bindings);

      // Normalize JS objects to plain Java (engine-agnostic)
      return normalizeResult(result);
  }

  private static Object normalizeResult(Object result) {
      // Same reflection-based logic as in PR
 }
```

Both approaches fix the issue for: 

- INLINE tasks
- Event handlers using JS
- Any future code using ScriptEvaluator

---

Actually, maybe the issue should be fixed at ForkJoinTaskMapper -- see comment below


nthmost-orkes:
I'm about to revert this and go a different way. 

### The Real Question: Who Should Own Normalization?

  Current approach (following PR #599):
  - Each task (JSON_JQ_TRANSFORM, INLINE) normalizes its own output
  - Pro: Tasks own their output format
  - Con: Duplicated logic, every JS-using task needs the same fix

  Possible Alternative -- fix in ForkJoinDynamicTaskMapper?
  - FORK_JOIN_DYNAMIC normalizes inputs it receives
  - Pro: Single fix point, works for all current and future tasks
  - Con: ForkJoinDynamicTaskMapper becomes responsible for type conversion

### PR #599 Might be wrong 

Looking back, PR #599 might have fixed it in the wrong place. They normalized in JSON_JQ_TRANSFORM because that's where they saw the problem, but the root cause is in ForkJoinDynamicTaskMapper's type checking.

Should we refactor to fix it in ForkJoinDynamicTaskMapper instead?

Arguments for:

1. DRY Principle: Why should INLINE, JSON_JQ_TRANSFORM, and every future JavaScript-using task each implement normalization?
2. Semantic Correctness: ForkJoinDynamicTaskMapper's whole purpose is to handle dynamic inputs flexibly. It already does type detection (instanceof Map). Handling JavaScript objects is
  part of that responsibility.
3. The Bug Is in the Type Check: The real bug is that forkTaskInput instanceof Map returns false for ScriptObjectMirror even though it's map-like. That's FORK_JOIN_DYNAMIC's problem, not INLINE's.
5. Future-Proof: When someone adds a new task that uses JavaScript, it'll just work. No need to remember to normalize.
6. Simpler for GraalJS Migration: One place to update when switching engines.

Arguments against:

1. Haven't looked at all the pieces yet

PR #633, opened 2025-11-05
Fix duplicate HTTP task execution via WorkflowRepairService race condition

### LINKED ISSUE #630: [Bug] Http Task executed twice after COMPLETED status
### Describe the bug

**Describe the bug**
When an HTTP task completes successfully and triggers a downstream HTTP task, but the upstream task simultaneously schedules another execution.
From below screenshot, I executed a time-consuming http task, when the first task has successfully executed, and downstream task also has completed. But the first task triggered another scheduling. Please help to explain the reason why will trigger this scenario. Thanks!
Workflow definition and screenshot attached as below.

**Details**
Conductor version: 3.21.19
Queue implementation: Postgres
Lock: Redis

**Expected behavior**
Workflow task should execute in sequence once the upstream task has completed.

**Workflow definition**
```json
"tasks": [
    {
      "name": "cmp_iac_validate",
      "taskReferenceName": "sleep0",
      "inputParameters": {
        "http_request": {
          "headers": {},
          "method": "POST",
          "readTimeOut": 300000,
          "body": {
            "retryNum": "0",
            "test": "false",
            "param": "test",
            "orderRequestTaskId": "78591038-453b-42f2-953a-0c4aa962dc0d",
            "platform": "SELF",
            "retrySleepTime": "299000"
          },
          "uri": "http://xxx.com/iac/v1/execute/xxx",
          "connectionTimeOut": 3600000
        },
        "_executor_": "${workflow.input._executor_}"
      },
      "type": "HTTP",
      "decisionCases": {},
      "defaultCase": [],
      "forkTasks": [],
      "startDelay": 0,
      "joinOn": [],
      "optional": false,
      "asyncComplete": false,
      "loopOver": [],
      "onStateChange": {},
      "permissive": false
    },
    {
      "name": "cmp_iac_validate",
      "taskReferenceName": "normal",
      "inputParameters": {
        "http_request": {
          "headers": {},
          "method": "POST",
          "readTimeOut": 300000,
          "body": {
            "retryNum": "0",
            "test": "false",
            "param": "SELF",
            "orderRequestTaskId": "c7958c72-30cf-416b-9938-bf305f93d961",
            "platform": "SELF",
            "retrySleepTime": "0"
          },
          "uri": "http://xxx.com/iac/v1/execute/xxx",
          "connectionTimeOut": 3600000
        },
        "_executor_": "${workflow.input._executor_}"
      },
      "type": "HTTP",
      "decisionCases": {},
      "defaultCase": [],
      "forkTasks": [],
      "startDelay": 0,
      "joinOn": [],
      "optional": false,
      "asyncComplete": false,
      "loopOver": [],
      "onStateChange": {},
      "permissive": false
    },
    ...
]
```
**Configuration file**
```
# See README in the docker for configuration guide

# db.type determines the type of database used
# See various configurations below for the values
#conductor.db.type=SET_THIS

# =====================================================#
#              Redis Configuration Properties
# =====================================================#

# The last part MUST be us-east-1c, it is not used and is kept for backwards compatibility
 conductor.redis.hosts=xxxx.com:6379:us-east-1c:xxx
 conductor.redis.workflowNamespacePrefix=conductor
 conductor.redis.queueNamespacePrefix=conductor_queues
 conductor.redis.taskDefCacheRefreshInterval=60
 conductor.redis.availabilityZone=us-east-1c
 conductor.redis.queuesNonQuorumPort=22122
 conductor.redis.ssl=false

 conductor.queue.type=postgres
 conductor.external-payload-storage.type=postgres

# Redis cluster settings for locking module
 conductor.redis-lock.server-address=redis://:xxxx@xxxx.com:6379
 #conductor.app.workflow-execution-lock-enabled=true
 conductor.workflow-execution-lock.type=redis
 conductor.redis-lock.server-type=SINGLE
 conductor.redis-lock.namespace=conductorlock
 conductor.app.workflowExecutionLockEnabled=true
 conductor.app.lockLeaseTime=60000ms
 conductor.app.lockTimeToTry=500ms
 # The interval after which a system task will be checked by the system task worker for completion. Example: 30s
 conductor.app.systemTaskWorkerCallbackDuration=30s
 conductor.app.systemTaskWorkerThreadCount=20
 conductor.app.systemTaskMaxPollCount=20

# The interval at which system task queues will be polled by system task workers. Example: 50ms
 conductor.app.systemTaskWorkerPollInterval=50ms

# =====================================================#
#              Postgres Configuration Properties
# =====================================================#

 conductor.db.type=postgres
 spring.datasource.url=jdbc:postgresql://xxxx.com:3433/conductor_00
 spring.datasource.username=xxxx
 spring.datasource.password=xxxx

# When using Elasticsearch 7 for indexing, set the following
 conductor.indexing.enabled=true
 conductor.elasticsearch.url=xxxx:9800
 conductor.elasticsearch.version=7
 conductor.elasticsearch.indexName=conductor
 conductor.elasticsearch.clusterHealthColor=yellow
 conductor.metrics-prometheus.enabled=false
 management.metrics.export.prometheus.enabled=false

 loadSample=false
 
```
**Screenshots**
Image

Image

nthmost-orkes:
## Summary

Fixes #630 

Long-running HTTP tasks (and other async system tasks) were being executed multiple times (typically 4x) due to a race condition in WorkflowRepairService.

## Problem

1. SystemTaskWorker polls task → removes from queue via `ackTaskReceived()`
2. AsyncSystemTaskExecutor increments `pollCount` **in-memory only**
3. HTTP call starts and blocks for minutes
4. WorkflowRepairService reads task from database → sees `pollCount = 0` (stale value)
5. Task not found in queue + `pollCount = 0` → WorkflowRepairService re-queues it
6. Another worker picks it up → duplicate execution
7. Repeats 3-4 times during long-running operations

## Solution

Two-part fix:

**1. WorkflowRepairService (lines 146-155)**
Check `pollCount > 0` before re-queueing. If task has been polled, a worker is actively processing it, so skip re-queueing.

**2. AsyncSystemTaskExecutor (lines 152-154)**
Persist task to database **before** calling `systemTask.start()`. This ensures `pollCount` is written to DB before any blocking operations, preventing WorkflowRepairService from reading stale data.

### Current Behavior (Before Fix)
- ❌ Healthy workers executing long tasks → task re-queued immediately → **duplicate executions**
- ✅ Crashed workers → task re-queued immediately → fast recovery

### Proposed Behavior (After Fix)
- ✅ Healthy workers executing long tasks → task NOT re-queued → **no duplicates**
- ⏱️ Crashed workers → task NOT re-queued → recovery via timeout mechanisms

### Recovery for Crashed Workers

If a worker crashes with this change, recovery depends on existing timeout mechanisms:

1. **Queue unack timeout**: Message automatically returns to queue after timeout expires
2. **Response timeout** (`responseTimeoutSeconds`): DeciderService marks task as TIMED_OUT if no updates received
3. **Execution timeout** (`timeoutSeconds`): Overall time limit from first poll

This means recovery would be **delayed** by the responseTimeout duration (typically 30-60s) rather than immediate.

### Questions for Discussion

- What is the relative frequency of long-running tasks vs worker crashes in production?
- Is delayed recovery for crashed workers acceptable to prevent duplicate executions?
- Are there scenarios where aggressive re-queueing is beneficial that we haven't considered?
- Should we add additional logic (e.g., time-based thresholds) to handle both cases?

## Changes

**Modified:**
- `core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowRepairService.java` - Added pollCount check (lines 146-155)
- `core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java` - Persist task before blocking start() call (lines 152-154)
- `core/src/test/java/com/netflix/conductor/core/reconciliation/TestWorkflowRepairService.java` - Added 2 new tests:
  - `verifyScheduledTaskWithPollCountIsNotRequeued()`
  - `verifySystemTaskWithPollCountIsNotRequeued()`

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -149,6 +149,9 @@ public class AsyncSystemTaskExecutor {
             if (task.getStatus() == TaskModel.Status.SCHEDULED) {
                 task.setStartTime(System.currentTimeMillis());
                 Monitors.recordQueueWaitTime(task.getTaskType(), task.getQueueWaitTime());
+                // Persist pollCount to DB before blocking operation
+                // This prevents WorkflowRepairService from re-queueing during long-running tasks
+                executionDAOFacade.updateTask(task);
                 systemTask.start(workflow, task, workflowExecutor);
             } else if (task.getStatus() == TaskModel.Status.IN_PROGRESS) {
                 systemTask.execute(workflow, task, workflowExecutor);
--- a/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowRepairService.java
+++ b/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowRepairService.java
@@ -143,6 +143,16 @@ public class WorkflowRepairService {
             // Ensure QueueDAO contains this taskId
             String taskQueueName = QueueUtils.getQueueName(task);
             if (!queueDAO.containsMessage(taskQueueName, task.getTaskId())) {
+                // Check if task has been polled (pollCount > 0) which indicates a worker is
+                // actively processing it. Don't re-queue to avoid duplicate executions.
+                if (task.getPollCount() > 0) {
+                    LOGGER.debug(
+                            "Task {} in workflow {} has been polled (pollCount: {}), skipping re-queue to avoid duplicate execution",
+                            task.getTaskId(),
+                            task.getWorkflowInstanceId(),
+                            task.getPollCount());
+                    return false;
+                }
                 queueDAO.push(taskQueueName, task.getTaskId(), task.getCallbackAfterSeconds());
                 LOGGER.info(
                         "Task {} in workflow {} re-queued for repairs",
--- a/core/src/test/java/com/netflix/conductor/core/reconciliation/TestWorkflowRepairService.java
+++ b/core/src/test/java/com/netflix/conductor/core/reconciliation/TestWorkflowRepairService.java
@@ -95,6 +95,63 @@ public class TestWorkflowRepairService {
     }
 
     @Test
+    public void verifyScheduledTaskWithPollCountIsNotRequeued() {
+        // This test verifies the fix for duplicate task execution issue.
+        // When a task has been polled (pollCount > 0), it means a worker is actively
+        // processing it, so we should NOT re-queue it even if it's not in the queue.
+        TaskModel task = new TaskModel();
+        task.setTaskType("SIMPLE");
+        task.setStatus(TaskModel.Status.SCHEDULED);
+        task.setTaskId("abcd");
+        task.setCallbackAfterSeconds(60);
+        task.setPollCount(1); // Task has been polled by a worker
+
+        when(queueDAO.containsMessage(anyString(), anyString())).thenReturn(false);
+
+        assertFalse(workflowRepairService.verifyAndRepairTask(task));
+        // Verify that queue message is checked
+        verify(queueDAO, times(1)).containsMessage(anyString(), anyString());
+        // Verify that queue message is NOT pushed because pollCount > 0
+        verify(queueDAO, never()).push(anyString(), anyString(), anyLong());
+    }
+
+    @Test
+    public void verifySystemTaskWithPollCountIsNotRequeued() {
+        // This test verifies the fix for duplicate HTTP task execution issue.
+        // Async system tasks (like HTTP) that have been polled should not be re-queued.
+        String taskType = "HTTP";
+        TaskModel task = new TaskModel();
+        task.setTaskType(taskType);
+        task.setStatus(TaskModel.Status.SCHEDULED);
+        task.setTaskId("abcd");
+        task.setCallbackAfterSeconds(60);
+        task.setPollCount(2); // Task has been polled multiple times
+
+        when(systemTaskRegistry.isSystemTask(taskType)).thenReturn(true);
+        when(systemTaskRegistry.get(taskType))
+                .thenReturn(
+                        new WorkflowSystemTask(taskType) {
+                            @Override
+                            public boolean isAsync() {
+                                return true;
+                            }
+
+                            @Override
+                            public boolean isAsyncComplete(TaskModel task) {
+                                return false;
+                            }
+                        });
+
+        when(queueDAO.containsMessage(anyString(), anyString())).thenReturn(false);
+
+        assertFalse(workflowRepairService.verifyAndRepairTask(task));
+        // Verify that queue message is checked
+        verify(queueDAO, times(1)).containsMessage(anyString(), anyString());
+        // Verify that queue message is NOT pushed because pollCount > 0
+        verify(queueDAO, never()).push(anyString(), anyString(), anyLong());
+    }
+
+    @Test
     public void verifyAndRepairSystemTask() {
         String taskType = "TEST_SYS_TASK";
         TaskModel task = new TaskModel();


nthmost-orkes:
## 1. Unack Timeout - Why It Doesn't Prevent Duplicates

The unack timeout should work, but there's an issue with the sequence:

**All SQL-based queue implementations (Postgres, MySQL, SQLite):**
```java
public boolean ack(String queueName, String messageId) {
    return removeMessage(tx, queueName, messageId);  // DELETES the message
}
```

**The sequence that causes duplicates:**
1. Worker polls task → message marked `popped = true`  
2. SystemTaskWorker calls `ackTaskReceived()` (line 140) → message **DELETED from queue**
3. Worker starts HTTP call (takes minutes)
4. `processAllUnacks()` runs but message is already gone
5. WorkflowRepairService sees task not in queue → re-queues it

So the unack timeout can't work because `ack()` deletes the message before task completion.

**Note:** `setUnackTimeout()` is only used by WorkflowSweeper for DECIDER_QUEUE (workflow messages), not for task queue messages.

## 2. Race Condition with pollCount > 1

The fix handles this - the check is `if (task.getPollCount() > 0)`, so it catches pollCount = 1, 2, 3, etc.

(Is there another race condition we're concerned about? Want to make sure we're not missing an edge case.)

## 3. Is This Postgres-Specific?

No, this affects all SQL-based implementations:
- Postgres: `ack()` → `removeMessage()`
- MySQL: `ack()` → `removeMessage()`
- SQLite: `ack()` → `removeMessage()`

Redis/DynoQueue might behave differently.

## Potential Alternative Fixes

**Option A: WorkflowRepairService check (current PR)**
- Works for all queue implementations
- Simple: if polled, don't re-queue
- Trade-off: Delayed recovery for crashed workers (via timeout mechanisms)

**Option B: Don't call `ackTaskReceived()` in SystemTaskWorker**
- Rely on unack timeout instead
- Issue: Default 60s timeout → still duplicates for 5-min tasks
- Would need dynamic unack timeouts based on task duration

**Option C: Change SQL queue implementations**
- Make `ack()` not delete immediately, keep with `acked=true` flag
- Large architectural change
- Changes queue semantics

Option A seems most practical but open to other approaches if there's a better way.

nthmost-orkes:
## Update Based on User Feedback

Thanks to @ColinShih for testing the original PR and finding it didn't solve the problem. 

**The Problem with the Original Fix:**

The original fix only checked `pollCount > 0` in WorkflowRepairService, but `pollCount` was being incremented in-memory and only persisted to the database in the `finally` block **after** the HTTP call completed. This meant:

- Memory: `pollCount = 1` ✅
- Database: `pollCount = 0` ❌ (stale during execution)
- WorkflowRepairService reads from database → sees `pollCount = 0` → still re-queues

**The Fix:**

Added a second part to persist the task to the database immediately after incrementing `pollCount` and before calling the blocking `systemTask.start()`. This ensures the database has the correct `pollCount` value before any long-running operations begin.

**Changes:**
- `AsyncSystemTaskExecutor.java` (lines 152-154): Added `executionDAOFacade.updateTask(task)` before `systemTask.start()`

This is a more general solution than the HTTP-specific fix proposed, as it handles all async system tasks (HTTP, Lambda, etc.).

nthmost-orkes:
## Staff Feedback: Lock Lease Extension Approach

Received feedback that the current fix has a performance issue - adding a DB write before every `start()` call is expensive.

### Root Cause (Corrected Understanding)

The real issue is **workflow lock lease timeout**, not just DB state:

1. WorkflowSweeper/Decider acquires lock on workflow with default 60s lease
2. System task (HTTP) executes and blocks for 5+ minutes
3. After 60s, workflow lock expires
4. Another process acquires lock on same workflow
5. Sees task still in SCHEDULED → executes it again → duplicate

### Proposed Better Solution: Lock Lease Extension

AsyncSystemTaskExecutor should periodically extend the workflow lock lease during long-running task execution.

### Implementation: Using ScheduledExecutorService

Conductor already uses `ScheduledExecutorService` for various periodic tasks. We can use the same pattern for lease extension:

**Note**: `LocalOnlyLock` uses `ScheduledExecutorService` but for **lock expiration** (one-time `schedule()`), not extension. We need **periodic extension** using `scheduleAtFixedRate()`:

```java
// In AsyncSystemTaskExecutor or new LockLeaseExtensionService
private static final ScheduledExecutorService LEASE_EXTENDER =
    Executors.newScheduledThreadPool(10, 
        ExecutorsUtil.newNamedThreadFactory("lock-lease-extender-"));

private ScheduledFuture<?> startLeaseExtension(String workflowId, long leaseTimeMs) {
    long extensionInterval = leaseTimeMs / 2; // Extend every 30s for 60s lease
    return LEASE_EXTENDER.scheduleAtFixedRate(
        () -> executionLockService.acquireLock(workflowId, 0, leaseTimeMs),
        extensionInterval, extensionInterval, TimeUnit.MILLISECONDS
    );
}

// Usage in AsyncSystemTaskExecutor
if (task.getStatus() == TaskModel.Status.SCHEDULED) {
    ScheduledFuture<?> leaseExtender = startLeaseExtension(workflowId, lockLeaseTime);
    try {
        systemTask.start(workflow, task, workflowExecutor);
    } finally {
        leaseExtender.cancel(false);
    }
}
```

**Difference from LocalOnlyLock:**
- `LocalOnlyLock`: `schedule()` - runs ONCE to delete lock after expiry
- `Lease extension`: `scheduleAtFixedRate()` - runs PERIODICALLY to renew before expiry

### Challenges

1. **Lock interface needs extension**: Add `extendLease()` method to `Lock` interface (or just re-acquire)
2. **Redisson support**: RLock has watchdog auto-renewal but needs investigation
3. **Postgres support**: pg_advisory_lock doesn't have native lease extension
4. **All Lock implementations** need updating: RedisLock, PostgresLockDAO, LocalOnlyLock, NoopLock

### Workaround for Now

Users can increase lock lease time to accommodate long-running tasks:
```properties
conductor.lockLeaseTime=360000  # 6 minutes for 5-min HTTP tasks
```

### Questions

1. Should we keep the current PR as-is (works but has perf cost) or revert the AsyncSystemTaskExecutor change?
2. What's the priority for implementing lock lease extension properly?
3. Should we start with Redis-only implementation or all Lock types?
4. Is Redisson's auto-renewal (watchdog) acceptable for RedisLock?

Created analysis document in `PLANNING/lock-lease-extension-analysis.md` with more details.

nthmost-orkes:
> Longer term - we should extend the lease while working on the long running system tasks.

Sounds like this is outside of current scope of work, but just for future reference, what do you see as the most likely path for implementing this?

Issue #637, opened 2025-11-06
[Bug] Base image openjdk not longer available, docker build failure


albert-cg:
**Describe the bug**
The current `Dockerfile` for `conductor-server` still references the deprecated OpenJDK base image (`openjdk:17-bullseye`), which has been **intentionally removed** from Docker Hub as of late 2024.  
Because of this, any build using the official Dockerfile now fails with:

**Details**
Conductor version: 3.21.14 (and all current Docker builds)  
Persistence implementation: Postgres
Queue implementation: Postgres
Lock: N/A

**To Reproduce**
1. Clone the Conductor OSS repository  
2. Run:
   ```bash
   docker-compose -f docker/docker-compose-local.yaml build conductor-server
   ```
3. Observe the build failure when resolving the base image.

**Expected behavior**
Docker build successuflly

**Screenshots**
```bash
#2 [internal] load metadata for docker.io/library/openjdk:17-bullseye
#2 ERROR: docker.io/library/openjdk:17-bullseye: not found
ERROR: failed to solve: openjdk:17-bullseye: failed to resolve source metadata for docker.io/library/openjdk:17-bullseye
```

**Additional context**
According to commented on https://github.com/docker-library/openjdk/issues/551

> This issue is caused by the intentional removal of unmaintained OpenJDK images from Docker Hub.
> 
> Reference discussion:
> https://github.com/docker-library/openjdk/pull/550
> 
> It was 100% intentional — none of the images referenced here should be used (they’ve all been deprecated and unmaintained for 3+ years).
> 
> My personal suggestions for users affected by this removal, in decreasing order of suggested precedence:
> * Switch to hub.docker.com/_/eclipse-temurin (actively maintained)
> * Use a more specific known-outdated tag (e.g. openjdk:8u342-jre)
> * Build your own image"

albert-cg:
According to https://github.com/docker-library/openjdk/pull/550#issuecomment-3488288561

> > switch to https://hub.docker.com/_/eclipse-temurin (it's literally plain compiled builds of OpenJDK for this exact intent, created and maintained by some of the very folks who maintain OpenJDK, and is actively supported/maintained, unlike almost all tags of the openjdk image)

> openjdk:8: eclipse-temurin:8
> openjdk:11: eclipse-temurin:11
> openjdk:17: eclipse-temurin:17
> openjdk:21: eclipse-temurin:21
> openjdk:25: eclipse-temurin:25

[...]

PR #655, opened 2025-11-10
637: Updated server Dockerfile from openjdk:17-bullseye to azul/zulu-openjdk-debian:17

ghost:
Pull Request type
----
- [ x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
OpenJDK [does not provide docker images anymore](https://github.com/docker-library/openjdk/issues/505), so we have to switch to another provider. The old OpenJDK images were based on Debian, so we would prefer to stay on Debian. Azul provides Debian based images. This one uses Bookworm (oldstable) , the old image was based on Bullseye (oldoldstable, EOL support). See https://www.debian.org/releases/

Issue: https://github.com/conductor-oss/conductor/issues/637
Alternatives considered
----

We could have gone with openjdk:17-slim but that one will probably not receive updates anymore as noted in their communcation. Eclipse Temurin is another well supported image but it uses Ubuntu


nthmost-orkes (approved):
thanks, i'll run docker build checks locally to verify.

Issue #619, opened 2025-10-21
FIX: Add synchronous join support to FORK_JOIN (Q4 Roadmap 1.3)


nthmost-orkes:
# Core Problem
FORK_JOIN currently operates only asynchronously, providing no option to wait for all forked tasks to complete before proceeding. This creates workflow limitations and requires developers to implement polling workarounds for synchronous fork patterns. Adding property-based control over join behavior addresses community needs without requiring the larger parallel scheduling overhaul.

# Key Requirements

**What Needs Implementation:**
- Add configuration property to control join behavior (synchronous vs. asynchronous)
- Implement synchronous join logic that waits for all forked tasks to complete
- Maintain backward compatibility with default asynchronous behavior
- Update task definition schema to include join mode property
- Add validation for join mode property values

**Affected Files:**
- `Join.java` or FORK_JOIN handler (add sync/async logic ~30 lines)
- `ForkJoinTaskMapper.java` (property handling)
- Task definition schema
- Workflow validation logic

# Success Criteria
The implementation must provide explicit control over fork/join semantics through a property-based configuration. Synchronous mode must wait for all forked tasks before proceeding, while asynchronous mode maintains current behavior. Default behavior must remain unchanged for backward compatibility. Both modes must work correctly with all supported queue backends.

# Testing & Documentation Scope
Requirements include unit tests for both synchronous and asynchronous join logic, integration tests with complex fork structures, backward compatibility verification, performance regression testing for async mode, queue backend compatibility tests, and documentation for the join mode property with examples and best practices.

# Related Issues
- Related to #310 - FORK_JOIN performance issues
- Related to #408 - FORK_JOIN sequential scheduling
- Note: Does NOT address the parallel scheduling performance issue (deferred to future work)

PR #680, opened 2025-12-05
Add synchronous join support to FORK_JOIN (#619)

nthmost-orkes:
Implements #619 - adds property-based control for synchronous vs asynchronous join behavior in FORK_JOIN tasks.

### Property-Based Control
- Added `joinMode` property to `WorkflowTask` (values: `"SYNC"` or `"ASYNC"`)
- Property is optional; defaults to `"ASYNC"` for backward compatibility
- Passed through `JoinTaskMapper` to join task input data

### Synchronous Behavior via Evaluation Offset
Rather than overriding `isAsync()` (which doesn't support per-task control), synchronous behavior is achieved through the evaluation offset mechanism:

- **SYNC mode**: Returns 0 evaluation offset → join is evaluated every workflow decision cycle
- **ASYNC mode** (default): Uses exponential backoff → join is polled with increasing delays

This approach works within the existing framework constraints while providing responsive synchronous-like behavior.

### Usage Example
```json
{
  "name": "join_task",
  "taskReferenceName": "join_ref",
  "type": "JOIN",
  "joinOn": ["task1", "task2"],
  "joinMode": "SYNC"
}
```

### Backward Compatibility
- Existing workflows without `joinMode` continue using async behavior
- No breaking changes to APIs or existing functionality

### Tests
- Unit tests for both sync and async modes
- JoinTaskMapper tests verify property passing
- All existing tests continue to pass

### Current Orkes Implementation (OrkesJoin)
Orkes Conductor replaces the OSS `Join` component entirely with `OrkesJoin`:
- **Always synchronous**: `isAsync() = false` - no property-based control
- **Wraps OSS Join**: Internally delegates to OSS Join for basic functionality
- **Additional features**:
  - Script-based join conditions via `expression` and `evaluatorType` (GraalJS)
  - Large fork optimization (conditional output capture for 500+ tasks)
  - Enhanced execution logic with custom evaluators

### Convergence Path Options

**Option 1: Orkes adopts OSS property-based approach**
- Replace `OrkesJoin` with enhanced OSS `Join` + property control
- Set `joinMode: "SYNC"` by default in Orkes for backward compatibility
- Keep Orkes-specific features (expressions, large fork optimization) as separate enhancements
- **Benefit**: Shared core, easier maintenance

**Option 2: Keep separate implementations**
- OSS: Property-based sync/async control
- Orkes: Always-synchronous with advanced features
- Both implementations continue to coexist
- **Benefit**: No migration needed, each optimized for use case

### Implementation Compatibility
This PR's approach is convergence-friendly:
- ✅ Additive (new optional property)
- ✅ Non-breaking (defaults to async)
- ✅ Framework-aware (uses evaluation offset, not isAsync override)
- ✅ Orkes can layer expression support on top of property-based control

The property-based approach could serve as foundation for future convergence if Orkes wants to support both sync and async modes with user control.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -55,7 +55,7 @@ jobs:
           java-version: '21'
 
       - name: Login to Docker Hub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/publish_build.yaml
+++ b/.github/workflows/publish_build.yaml
@@ -93,7 +93,7 @@ jobs:
           cp server/build/libs/*boot*.jar docker/server/libs/conductor-server.jar
 
       - name: Login to Docker Hub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}


core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -125,6 +125,14 @@ public boolean execute(
 
     @Override
     public Optional<Long> getEvaluationOffset(TaskModel taskModel, long maxOffset) {
+        // Check if joinMode is set to SYNC
+        String joinMode = (String) taskModel.getInputData().get("joinMode");
+        if ("SYNC".equalsIgnoreCase(joinMode)) {
```
v1r3n: can this be an enum?

core/src/main/java/com/netflix/conductor/core/execution/mapper/JoinTaskMapper.java
```diff
@@ -59,6 +59,9 @@ public List<TaskModel> getMappedTasks(TaskMapperContext taskMapperContext) {
 
         Map<String, Object> joinInput = new HashMap<>();
         joinInput.put("joinOn", workflowTask.getJoinOn());
+        if (workflowTask.getJoinMode() != null) {
+            joinInput.put("joinMode", workflowTask.getJoinMode());
```
v1r3n: I think we can read workflow task inside the task, so we can avoid putting this into the input.  need to verify though but I think it should be possible.

nthmost-orkes:
**Update on convergence:** Replaced the `offset=0` approach with `isAsync(TaskModel task)` on `WorkflowSystemTask` — this brings OSS `Join` much closer to `OrkesJoin` than we had before when we were avoiding changing the `WorkflowSystemTask` API

| | OSS `Join` (now) | `OrkesJoin` |
|---|---|---|
| Sync execution | `isAsync(task) = false` when `joinMode=SYNC` — runs in decide thread | `isAsync() = false` always |
| Per-task control | ✅ user-controlled via `joinMode` | ❌ always sync |
| Script conditions | ❌ | ✅ (`expression` + GraalJS) |
| Large fork optimization | ❌ | ✅ (500+ tasks) |

**Convergence path for Orkes is now simple:** subclass OSS `Join`, override `isAsync(TaskModel task)` to return `false` unconditionally — no workflow definition changes or migrations needed. Orkes-specific features (GraalJS, large fork optimization) layer on top as before. The `joinMode` enum also gives Orkes a path to expose user-controlled sync/async in the future.

PR #691, opened 2025-12-13
Add support for dead letter exchange and queue arguments in AMQP

dewidyabagus:
Pull Request type
----
- [x] Feature

Changes in this PR
----
Enhanced AMQPSettings to support additional RabbitMQ queue arguments through URI parameters.

New URI Parameters:
  - deadLetterExchange: Configure x-dead-letter-exchange
  - deadLetterRoutingKey: Configure x-dead-letter-routing-key
  - messageTtl: Configure x-message-ttl (milliseconds)
  - maxLength: Configure x-max-length

Example usage:
```text
amqp_queue:myQueue?deadLetterExchange=myDLX&deadLetterRoutingKey=failed&messageTtl=60000
```

Problem
----
When workflows are triggered via Event Handlers sourced from RabbitMQ, and the workflow initiation fails due to backend database issues, the incoming message is nacked using `basic.nack` with `requeue=false`. As a result, the message is not requeued to the original queue and is expected to be routed to a Dead Letter Exchange (DLX). However, since the DLX configuration is currently not in place, the message is ultimately discarded.

Related issues: https://github.com/conductor-oss/conductor/issues/787

Solution
----
By introducing new parameters for Dead Letter Exchange (DLX) configuration that enable queues to use DLX settings, along with additional arguments for message expiration (TTL) and maximum queue length, all of which are optional.

### DIFF
--- a/amqp/src/main/java/com/netflix/conductor/contribs/queue/amqp/util/AMQPConfigurations.java
+++ b/amqp/src/main/java/com/netflix/conductor/contribs/queue/amqp/util/AMQPConfigurations.java
@@ -25,7 +25,11 @@ public enum AMQPConfigurations {
     PARAM_DURABLE("durable"),
     PARAM_EXCLUSIVE("exclusive"),
     PARAM_AUTO_DELETE("autoDelete"),
-    PARAM_MAX_PRIORITY("maxPriority");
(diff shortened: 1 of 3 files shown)


nthmost-orkes:
Linking to tracking issue #787. Thank you @dewidyabagus

PR #696, opened 2025-12-17
Terminate workflow with custom failure workflow version

phgbecker:
Pull Request type
----
- [X] Feature

🔨 Changes in this PR
----
- Adds definition, method, and documentation to allow terminating a workflow with a custom failure definition version. 

📸 Screenshots
----
- Workflow with failure definition, without version (as is):
Workflow with failure definition, without version (as is)

- Workflow with failure definition, with custom version:
Workflow with failure definition, with custom version

- Workflow search results:
Workflow search results


nthmost-orkes:
Closing in favor of tracking issue #792. Thank you @phgbecker for the failure workflow version support!

nthmost-orkes:
Sorry for the close — too aggressive on the stale PR sweep. This feature is solid and we'd like to get it merged.

One fix is needed before it can land: the proto field renumbering is a breaking wire-format change. Inserting \`failure_workflow_version\` at field 8 shifts all subsequent fields (schema_version, restartable, etc.) up by one, which breaks deserialization for any existing gRPC clients.

**Fix**: instead of inserting at id=8, add \`failureWorkflowVersion\` at a new unused field id. Field 16 is currently free in both \`WorkflowDef.java\` (\`@ProtoField\`) and \`workflowdef.proto\`. So:

In `WorkflowDef.java` — keep all existing `@ProtoField` ids as-is and add:
```java
@ProtoField(id = 16)
private Integer failureWorkflowVersion;
```

In `workflowdef.proto` — add without renumbering anything else:
```protobuf
int32 failure_workflow_version = 16;
```

Everything else in the PR — the logic in `WorkflowExecutorOps`, the interface change, tests, and docs — looks great. Just that one proto numbering fix and this should be ready to merge.

phgbecker:
> Sorry for the close — too aggressive on the stale PR sweep. This feature is solid and we'd like to get it merged.
> 
> One fix is needed before it can land: the proto field renumbering is a breaking wire-format change. Inserting `failure_workflow_version` at field 8 shifts all subsequent fields (schema_version, restartable, etc.) up by one, which breaks deserialization for any existing gRPC clients.
> 
> **Fix**: instead of inserting at id=8, add `failureWorkflowVersion` at a new unused field id. Field 16 is currently free in both `WorkflowDef.java` (`@ProtoField`) and `workflowdef.proto`. So:
> 
> In `WorkflowDef.java` — keep all existing `@ProtoField` ids as-is and add:
> 
> ```java
> @ProtoField(id = 16)
> private Integer failureWorkflowVersion;
> ```
> 
> In `workflowdef.proto` — add without renumbering anything else:
> 
> ```proto
> int32 failure_workflow_version = 16;
> ```
> 
> Everything else in the PR — the logic in `WorkflowExecutorOps`, the interface change, tests, and docs — looks great. Just that one proto numbering fix and this should be ready to merge.

Hi @nthmost-orkes, just did what you requested. Thanks for the review!

nthmost-orkes (dismissed):
Nice work overall — clean feature, solid tests, and good to see docs and proto both updated in the same PR. A couple of things need fixing before this can merge.

_**(Sorry i didn't catch this on the first review -- a lot is moving forward in parallel for Conductor OSS right now!)**_

**Proto `fromProto` will silently break existing failure workflows**

In `AbstractProtoMapper.java`, `fromProto` unconditionally calls:

```java
to.setFailureWorkflowVersion( from.getFailureWorkflowVersion() );
```

Proto3 `int32` fields default to `0` when not set, so any `WorkflowDef` that doesn't have `failureWorkflowVersion` configured will come out of gRPC deserialization with `Integer(0)` instead of `null`. From there, `lookupForWorkflowDefinition(name, 0)` fires and throws `NotFoundException` because version 0 almost certainly doesn't exist in the DB — silently breaking failure workflows for every gRPC user who hasn't set this field.

Compare with `toProto`, which correctly null-checks before setting. Fix `fromProto` to match:

```java
if (from.getFailureWorkflowVersion() != 0) {
    to.setFailureWorkflowVersion(from.getFailureWorkflowVersion());
}
```

Or declare the proto field as `optional int32 failure_workflow_version = 16;` and gate on `from.hasFailureWorkflowVersion()` — that's the cleaner proto3 approach.

**Breaking interface change**

`terminateWorkflow(WorkflowModel, String, String)` was removed from the `WorkflowExecutor` interface with no backward-compatible shim. Any external implementation of that interface won't compile. A default method covers it:

```java
default WorkflowModel terminateWorkflow(WorkflowModel workflow, String reason, String failureWorkflow) {
    return terminateWorkflow(workflow, reason, failureWorkflow, null);
}
```

**Doc type mismatch**

The JSON snippet in `handling-errors.md` shows `failureWorkflowVersion` in quotes, making it look like a string:

```json
"failureWorkflowVersion": "<Version of your compensation flow...>",
```

Should use an integer example: `"failureWorkflowVersion": 2`

**Nit**

Test typo in `WorkflowDefValidatorTest.java`: `"owner@orkers.io"` → `"owner@orkes.io"`

---

The core logic — reading `failureWorkflowVersion` from `WorkflowDef`, threading it through `terminateWorkflow` → `StartWorkflowInput.setVersion()` → `lookupForWorkflowDefinition` — is correct, and the version-specific vs. latest-version test coverage is exactly right. The proto `fromProto` bug is the one that must be addressed before merge.

phgbecker:
> Nice work overall — clean feature, solid tests, and good to see docs and proto both updated in the same PR. A couple of things need fixing before this can merge.
> 
> _**(Sorry i didn't catch this on the first review -- a lot is moving forward in parallel for Conductor OSS right now!)**_
> 
> **Proto `fromProto` will silently break existing failure workflows**
> 
> In `AbstractProtoMapper.java`, `fromProto` unconditionally calls:
> 
> ```java
> to.setFailureWorkflowVersion( from.getFailureWorkflowVersion() );
> ```
> 
> Proto3 `int32` fields default to `0` when not set, so any `WorkflowDef` that doesn't have `failureWorkflowVersion` configured will come out of gRPC deserialization with `Integer(0)` instead of `null`. From there, `lookupForWorkflowDefinition(name, 0)` fires and throws `NotFoundException` because version 0 almost certainly doesn't exist in the DB — silently breaking failure workflows for every gRPC user who hasn't set this field.
> 
> Compare with `toProto`, which correctly null-checks before setting. Fix `fromProto` to match:
> 
> ```java
> if (from.getFailureWorkflowVersion() != 0) {
>     to.setFailureWorkflowVersion(from.getFailureWorkflowVersion());
> }
> ```
> 
> Or declare the proto field as `optional int32 failure_workflow_version = 16;` and gate on `from.hasFailureWorkflowVersion()` — that's the cleaner proto3 approach.
> 
> **Breaking interface change**
> 
> `terminateWorkflow(WorkflowModel, String, String)` was removed from the `WorkflowExecutor` interface with no backward-compatible shim. Any external implementation of that interface won't compile. A default method covers it:
> 
> ```java
> default WorkflowModel terminateWorkflow(WorkflowModel workflow, String reason, String failureWorkflow) {
>     return terminateWorkflow(workflow, reason, failureWorkflow, null);
> }
> ```
> 
> **Doc type mismatch**
> 
> The JSON snippet in `handling-errors.md` shows `failureWorkflowVersion` in quotes, making it look like a string:
> 
> ```json
> "failureWorkflowVersion": "<Version of your compensation flow...>",
> ```
> 
> Should use an integer example: `"failureWorkflowVersion": 2`
> 
> **Nit**
> 
> Test typo in `WorkflowDefValidatorTest.java`: `"owner@orkers.io"` → `"owner@orkes.io"`
> 
> The core logic — reading `failureWorkflowVersion` from `WorkflowDef`, threading it through `terminateWorkflow` → `StartWorkflowInput.setVersion()` → `lookupForWorkflowDefinition` — is correct, and the version-specific vs. latest-version test coverage is exactly right. The proto `fromProto` bug is the one that must be addressed before merge.

Once again, thanks for the thoughtful code review @nthmost-orkes!

NicholasDCole (approved):

PR #709, opened 2026-01-08
Ensure index operations succeed before removing workflows from ExecutionDAO

rajeshwar-nu:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

## Motivation

Workflow removal currently risks data loss when index operations fail. We want consistent behavior across archive and non‑archive removals so the source of truth (ExecutionDAO) is only deleted after index changes succeed. Also, index deletions should be idempotent: a missing index record should not block DAO
  removal because it likely means a prior step already removed it.

  ## Changes

  - Reordered removeWorkflow to always perform workflow/task index operations before deleting from ExecutionDAO.
  - Treated index “not found” as a non‑blocking condition for non‑archive removals (workflow + task index).
  - Applied the same idempotent handling to removeWorkflowWithExpiry.
  - Added comments explaining the ordering and idempotency logic.
  - Added tests covering non‑archive index “not found” cases and archive failure cases.


rajeshwar-nu:
@v1r3n can I please get a review on this

rajeshwar-nu:
Hi @nthmost-orkes , can this be reviewed and merged please

nthmost-orkes:
sorry for delay, we had to refactor tests to get builds to pass.

PR #730, opened 2026-01-23
Port improved decider postpone logic to new sweeper

rajeshwar-nu:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Improve workflow postpone scheduling in the new sweeper path:
- Ported the enhanced unack/postpone logic into `ExecutorUtils.computePostpone` to scan all eligible tasks, cap by maxPostpone, and handle WAIT/HUMAN/response timeouts.
- Wired the new computePostpone signature into `WorkflowExecutorOps`.
- Tweaked new sweeper behavior: no nested parallelism, idle backoff, capped lock backoff, and guard missing subworkflow.

Alternatives considered
----

Kept the old reconciliation sweeper logic enabled, but rejected because the new sweeper is the default and we want the improved postpone logic there instead.

### DIFF
--- a/nats/src/main/java/com/netflix/conductor/contribs/queue/nats/JetStreamObservableQueue.java
+++ b/nats/src/main/java/com/netflix/conductor/contribs/queue/nats/JetStreamObservableQueue.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -48,6 +49,7 @@ public class JetStreamObservableQueue implements ObservableQueue {
     private final Lock mu = new ReentrantLock();
     private final String queueType;
     private final String subject;
+    private final String streamName;
     private final String queueUri;
     private final JetStreamProperties properties;
     private final Scheduler scheduler;
@@ -78,7 +80,7 @@ public class JetStreamObservableQueue implements ObservableQueue {
             this.subject = getQueuePrefix(conductorProperties, properties) + queueUri;
             queueGroup = null;
         }
-
+        this.streamName = streamNameFromSubject(this.subject);
         this.queueType = queueType;
         this.properties = properties;
         this.scheduler = scheduler;
@@ -95,6 +97,13 @@ public class JetStreamObservableQueue implements ObservableQueue {
         return StringUtils.isBlank(properties.getListenerQueuePrefix())
                 ? conductorProperties.getAppId() + "_jsm_notify_" + stack
                 : properties.getListenerQueuePrefix();
+    }
+
+    private static String streamNameFromSubject(String subject) {
+        return subject.replace(".", "_")
+                .replace("*", "ANY")
+                .replace(">", "ALL")
+                .toUpperCase(Locale.ROOT);
     }
 
     @Override
@@ -247,7 +256,8 @@ public class JetStreamObservableQueue implements ObservableQueue {
     private void createStream(JetStreamManagement jsm) {
         StreamConfiguration streamConfig =
                 StreamConfiguration.builder()
-                        .name(subject)
+                        .name(streamName)
+                        .subjects(subject)
                         .replicas(properties.getReplicas())
                         .retentionPolicy(RetentionPolicy.Limits)
                         .maxBytes(properties.getStreamMaxBytes())
@@ -293,7 +303,7 @@ public class JetStreamObservableQueue implements ObservableQueue {
                         .build();
 
         try {
-            jsm.addOrUpdateConsumer(subject, consumerConfig);
+            jsm.addOrUpdateConsumer(streamName, consumerConfig);
             return consumerConfig;
         } catch (IOException | JetStreamApiException e) {
             throw new NatsException("Failed to add/update consumer", e);
@@ -305,7 +315,7 @@ public class JetStreamObservableQueue implements ObservableQueue {
             JetStream js = nc.jetStream();
 
             PushSubscribeOptions pso =
-                    PushSubscribeOptions.builder().configuration(consumerConfig).stream(subject)
+                    PushSubscribeOptions.builder().configuration(consumerConfig).stream(streamName)
                             .bind(true)
                             .build();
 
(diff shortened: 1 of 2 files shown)


nthmost-orkes:
Closing in favor of tracking issue #798. Thank you @rajeshwar-nu for porting the postpone logic!

rajeshwar-nu:
@nthmost-orkes this PR is fixing #798 , I didn't understand why do we need to close this?

nthmost-orkes:
Sorry, was too agressive on sweeping stale PRs!  Reopened and revewing now.

nthmost-orkes:
The build failure is a compilation error in `WorkflowSweeper.java`. The sub-workflow repair block added after the `finally` at the end of `sweep()` references `task`, but that variable is out of scope there — it's defined inside the `isTaskRepairable` forEach lambda earlier in the method.

The block starting at line ~258:
```java
if (task.getTaskType().equals(TaskType.TASK_TYPE_SUB_WORKFLOW)
        && task.getStatus() == TaskModel.Status.IN_PROGRESS) {
```

...appears to have been meant to go inside the task-repair loop body (inside the `forEach` at line ~199), but instead it was placed after the `finally` block closes, where `task` doesn't exist.

Fix: move that block into the `forEach` lambda where tasks are being repaired, or refactor `repairSubWorkflowTask` to be called from within the loop.

rajeshwar-nu:
@nthmost-orkes Fixed the failing tests, ready for review again

rajeshwar-nu:
Fixes https://github.com/conductor-oss/conductor/issues/831
Fixes https://github.com/conductor-oss/conductor/issues/832
Fixes https://github.com/conductor-oss/conductor/issues/833
Fixes https://github.com/conductor-oss/conductor/issues/834

nthmost-orkes:
Hi @rajeshwar-nu -- thanks for rolling with the code organization I've been trying to do!  Reviewing this now, should be able to get this integrated soon.

nthmost-orkes:
Good improvements overall — the sequential sweep, idle sleep, smart lock backoff, sub-workflow repair, and the `computePostpone` rewrite are all solid. The WAIT task condition bug fix (inverted `== 0` check) and scanning all tasks for the minimum postpone (instead of breaking on the first match) are genuine correctness wins.

One critical issue and a couple of minor ones:

---

### 🚨 Critical: `isTaskRepairable` filter was dropped but not replaced

`isTaskRepairable` is still defined (lines 103–120) but the `.filter(isTaskRepairable)` was removed from the `workflow.getTasks().forEach()` repair loop. Now **all tasks** — including COMPLETED, FAILED, CANCELED, TIMED_OUT — are iterated.

For each terminal task:
1. `QueueUtils.getQueueName(task)` returns the worker queue name (e.g. `MY_TASK_TYPE`)
2. `queueDAO.containsMessage(queueName, task.getTaskId())` returns `false` (terminal tasks have been removed from their queues)
3. The code logs a spurious *"Going to repair the task"* warning, increments the `recordQueueMessageRepushFromRepairService` metric, and — if the second `containsMessage` check also returns false — **pushes the completed task back to the worker queue**

This could cause already-completed simple tasks to be re-executed by workers, and will definitely produce metric noise / log spam for every sweep of any workflow that has completed tasks. The `isTaskRepairable` filter should be re-applied (`.stream().filter(isTaskRepairable).forEach(...)`) or the forEach body should guard with `if (task.getStatus().isTerminal()) return;`.

---

### ⚠️ `computePostpone` returns 0 instead of `workflowOffsetTimeout` when no eligible tasks

When a workflow has no active tasks (all terminal), the new code returns `Duration.ofSeconds(0)`. The old code returned `workflowOffsetTimeout`. In `WorkflowExecutorOps`, since `0 != timeout`, `setUnack(0)` is called — re-queuing the workflow for immediate re-sweep. In normal flow this gets caught by `forceSetLastTaskAsNotExecuted`, but it's a tighter spin than before. Probably fine, just worth being aware of.

---

### ℹ️ `isTaskRepairable` is dead code

Once the critical issue above is resolved by re-applying the filter, `isTaskRepairable` will be in use again. If you instead choose to inline the guard, the predicate field should be deleted.

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -194,28 +209,65 @@ public void sweep(String workflowId) {
...
+                                                    "Going to repair the task {} / {}, with status {}, workflow = {}, timeout = {}, now-wait = {}",
+                                                    task.getTaskId(),
+                                                    task.getReferenceTaskName(),
+                                                    task.getStatus(),
+                                                    workflowId,
+                                                    task.getWaitTimeout(),
+                                                    (now - task.getWaitTimeout()));
+                                            Monitors.recordQueueMessageRepushFromRepairService(
+                                                    task.getTaskDefName());
+                                            if (!queueDAO.containsMessage(
```
nthmost-orkes, line 232: Is this second `containsMessage` check redundant? We are already inside the outer `if (\!queueDAO.containsMessage(...))` at line 218, so the message should be guaranteed missing here. Could there be a race where it reappears between the two checks, or is it safe to remove?

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -246,6 +298,25 @@ public void sweep(String workflowId) {
         }
     }
 
+    private void repairSubWorkflowTask(TaskModel task, WorkflowModel subWorkflow) {
+        switch (subWorkflow.getStatus()) {
```
nthmost-orkes, line 304: Would it be worth adding a `default` case here with a log warning? Probably unreachable in practice, but could help catch unexpected sub-workflow states if new statuses are added later.

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
rajeshwar-nu: added a small code comment, explaining its need to reduce risk of duplicate processing from a concurrent sweep process on same workflow

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
rajeshwar-nu: done, good one

Issue #733, opened 2026-01-25
Add Azure Workload Identity Authentication to Azure Blob


vivianzheng404:
Add Azure Workload Identity Authentication to Azure Blob. Workload Identity is the production standard to authenticate conductor server pod with azure blob. 

This is useful for companies and developers whose cloud infrastructure runs on Microsoft Azure.

---
PR https://github.com/conductor-oss/conductor/pull/734/files

PR #734, opened 2026-01-25
Add Azure Workload Identity Authentication to Azure Blob

vivianzheng404:
Pull Request type
----
- [ ] Feature

Changes in this PR
----

Add Azure Workload Identity Authentication to Azure Blob

Issue #733


steph409:
we would also like to use azure blob, thanks for preparing the PR! I think it would be good to use the defaultcredential, which works for workloadidentity but also systemassigned identities

Issue #727, opened 2026-01-22
[FEATURE] Support multiple workflow status listeners


akhilpathivada:
### Description
Currently, Conductor allows only a single workflow status listener type to be configured via `conductor.workflow-status-listener.type`. This means choosing between **kafka**, **workflow_publisher**, **queue_publisher**, or **stub** - but not multiple simultaneously.

### Current Limitation
Only ONE of these can be active at a time:
```
conductor.workflow-status-listener.type=workflow_publisher
conductor.workflow-status-listener.type=kafka
conductor.workflow-status-listener.type=queue_publisher
```
### Use Case
We'd like to publish workflow events to multiple destinations simultaneously. For example:

1. Kafka (analytics/monitoring)
2. Queues (internal automations)
3. Webhooks (external integrations)

Why not just one?

1. Different consumption patterns (streaming vs queuing vs HTTP)
2. Independent failure domains (Kafka down ≠ queue-based automations broken)
3. Different retention/retry requirements
4. Legacy systems (queue-based) + new systems (Kafka-based) coexisting

> It's basically event broadcasting (similar to how AWS EventBridge works) - not duplication, just different consumers needing the same event.

### Questions
Before proposing an implementation, we wanted to understand:

1. Is the single-listener design intentional? Are there architectural reasons for this constraint?
2. Would a pattern similar to Event Queue Providers (currently used in Condcutor)  OR a composite listener pattern (similar to how other frameworks handle multiple listeners) be acceptable?
3. Are there any known concerns or downsides we should be aware of?

### Recommended Approaches
**Option-1: Enable/Disable Pattern (Mirrors Event Queue Providers)**
This approach follows Conductor's existing architecture for Event Queue Providers, where multiple providers coexist and are auto-discovered by Spring.
Configuration:
```
# Enable multiple listeners simultaneously
conductor.workflow-status-listener.kafka.enabled=true
conductor.workflow-status-listener.workflow-publisher.enabled=true
conductor.workflow-status-listener.queue-publisher.enabled=true

# Each listener has its own configuration namespace
conductor.workflow-status-listener.kafka.topic=workflow-events
conductor.status-notifier.notification.url=http://webhook-endpoint
conductor.workflow-status-listener.queue-publisher.successQueue=_callbackSuccessQueue
```

How it would work:

- Each listener configuration class uses `@ConditionalOnProperty` with its own enable flag (e.g., `kafka.enabled`, `workflow-publisher.enabled`)
[...]

PR #720, opened 2026-01-20
feat: Add configurable workflow status subscription for HTTP webhook publisher

akhilpathivada:
Resolves #710 

### Summary
Enables configuration-based control over which workflow status changes trigger HTTP webhook events, matching the existing task status publishing behavior.

### Problem
The HTTP webhook publisher (`conductor.workflow-status-listener.type=workflow_publisher`) currently hardcodes publishing only **COMPLETED** and **TERMINATED** workflow events, with no way to configure which statuses to receive. This is inconsistent with the task publisher which allows full configurability via subscribed-task-statuses.

### Solution
Added **subscribedWorkflowStatuses** configuration property to enable selective workflow event publishing:

```
# New configuration property
conductor.status-notifier.notification.subscribed-workflow-statuses=RUNNING,COMPLETED,TERMINATED,PAUSED

# Valid values: RUNNING, COMPLETED, FAILED, TIMED_OUT, TERMINATED, PAUSED, RESUMED, RESTARTED, RETRIED, RERAN, FINALIZED
```

### Changes

1. **StatusNotifierNotificationProperties**: Added `subscribedWorkflowStatuses` field with getters/setters
2. **StatusChangePublisherConfiguration**: Inject workflow status config and pass to publisher
3. **StatusChangePublisher**: Implement conditional checks in all 9 workflow lifecycle methods
4. **StatusChangePublisherTest**: Added comprehensive unit tests (17 test cases)
5. **config.properties**: Added configuration example with documentation

### Testing
✅ All 9 workflow status methods tested with various configurations
✅ Null and empty list handling
✅ Build successful with no linter errors

PR #740, opened 2026-01-28
feat: Support multiple WF status listeners via composite pattern

akhilpathivada:
Resolves https://github.com/conductor-oss/conductor/issues/727

### Summary

Implements support for multiple workflow status listeners using the Composite Pattern, enabling simultaneous publishing of workflow events to multiple destinations (Kafka, queues, webhooks, archive).

### Motivation

Previously, Conductor only allowed a single workflow status listener type to be configured via `conductor.workflow-status-listener.type`. Users had to choose between `kafka`, `queue_publisher`, `workflow_publisher`, or `archive` - but could not use multiple simultaneously.

This limitation prevented organizations from broadcasting the same workflow events to different systems for different purposes:

- Kafka: Real-time analytics and monitoring dashboards
- Queues: Internal workflow orchestration and cleanup jobs
- Webhooks: External system integrations (ticketing systems)

Each system has different requirements (stream processing vs guaranteed delivery vs fire-and-forget), different failure domains, and different retention policies.

### Changes

1. **CompositeWorkflowStatusListener**: Delegates workflow events to all configured listeners in parallel
2. **CompositeWorkflowStatusListenerConfiguration**: Spring configuration with factory method for dynamic listener creation
3. **CompositeWorkflowStatusListenerProperties**: Configuration properties for composite listener
4. Comprehensive unit tests with coverage for all workflow event types and failure scenarios

### Configuration Example
```
# Enable composite mode
conductor.workflow-status-listener.type=composite

# List the listeners to enable (comma-separated)
conductor.workflow-status-listener.composite.types=kafka,workflow_publisher,queue_publisher

# Each listener retains its existing configuration namespace

# Kafka configuration
conductor.workflow-status-listener.kafka.producer[bootstrap.servers]=kafka:29092
conductor.workflow-status-listener.kafka.default-topic=workflow-events

# Workflow publisher (webhook) configuration
conductor.status-notifier.notification.url=http://webhook-endpoint:8080/workflow-events

# Queue publisher configuration
conductor.workflow-status-listener.queue-publisher.successQueue=_callbackSuccessQueue
conductor.workflow-status-listener.queue-publisher.failureQueue=_callbackFailureQueue
```
### Testing
✅ Unit tests for all workflow event types 
✅ Failure isolation tests (single and multiple listener failures)
✅ Edge cases (empty list, single listener)
✅ Verification that all delegates are invoked correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 
 <h1 align="center" style="border-bottom: none">
-    Conductor - {agentic, durable, scalable} Workflow Engine
+    Conductor - Internet scale Workflow Engine
 </h1>
 
 
@@ -18,7 +18,9 @@
 [![Conductor Slack](https://img.shields.io/badge/Slack-Join%20the%20Community-blueviolet?logo=slack)](https://join.slack.com/t/orkes-conductor/shared_invite/zt-2vdbx239s-Eacdyqya9giNLHfrCavfaA)
 [![Conductor OSS](https://img.shields.io/badge/Conductor%20OSS-Visit%20Site-blue)](https://conductor-oss.org)
 
-Conductor is an open-source, durable workflow engine built at [Netflix](https://netflixtechblog.com/netflix-conductor-a-microservices-orchestrator-2e8d4771bf40) for orchestrating microservices, AI agents, and event-driven workflows at internet scale. Actively maintained by [Orkes](https://orkes.io) and a growing [community](https://join.slack.com/t/orkes-conductor/shared_invite/zt-2vdbx239s-Eacdyqya9giNLHfrCavfaA).
+#### Orchestrating distributed systems means wrestling with failures, retries, and state recovery. Conductor handles all of that so you don't have to.
+
+Conductor is an open-source, durable workflow engine built at [Netflix](https://netflixtechblog.com/netflix-conductor-a-microservices-orchestrator-2e8d4771bf40) for orchestrating microservices, AI agents, and durable workflows at internet scale. Trusted in production at Netflix, Tesla, LinkedIn, and J.P. Morgan. Actively maintained by [Orkes](https://orkes.io) and a growing [community](https://join.slack.com/t/orkes-conductor/shared_invite/zt-2vdbx239s-Eacdyqya9giNLHfrCavfaA).
 
 [![conductor_oss_getting_started](https://github.com/user-attachments/assets/6153aa58-8ad1-4ec5-93d1-38ba1b83e3f4)](https://youtu.be/4azDdDlx27M)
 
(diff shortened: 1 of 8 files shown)


akhilpathivada:
I'll mark this as ready for review once https://github.com/conductor-oss/conductor/pull/720 is merged, as this PR is dependent on those changes.

akhilpathivada:
@v1r3n could you please help reviewing this?

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
```diff
@@ -0,0 +1,202 @@
...
+    CompositeWorkflowStatusListenerProperties.class,
+    StatusNotifierNotificationProperties.class,
+    KafkaWorkflowStatusPublisherProperties.class,
+    ConductorQueueStatusPublisherProperties.class,
+    ArchivingWorkflowListenerProperties.class
+})
+@ConditionalOnProperty(name = "conductor.workflow-status-listener.type", havingValue = "composite")
+public class CompositeWorkflowStatusListenerConfiguration {
+
+    private static final Logger LOGGER =
```
nthmost-orkes, line 48: This `RestClientManager` bean is registered unconditionally whenever `type=composite` is active — even if `workflow_publisher` is not in the composite types list. This means a user who configures `types=kafka,queue_publisher` will still have `RestClientManager` instantiated, potentially failing at startup if `conductor.status-notifier.notification.url` is not configured.

This also makes the `Optional<RestClientManager>` parameter in `compositeWorkflowStatusListener` misleading, since the bean is always present.

The fix is to move `RestClientManager` construction inside the `workflow_publisher` case in `createListener()`, instead of registering it as a top-level bean.

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListener.java
```diff
@@ -0,0 +1,133 @@
...
+                workflow, "onWorkflowPaused", listener -> listener.onWorkflowPaused(workflow));
+    }
+
+    @Override
+    public void onWorkflowResumed(WorkflowModel workflow) {
+        delegateToListeners(
+                workflow, "onWorkflowResumed", listener -> listener.onWorkflowResumed(workflow));
+    }
+
+    @Override
```
nthmost-orkes, line 120: `listeners.stream().parallel()` routes work through the JVM common ForkJoin pool, which is shared across the entire application. For the typical use case of 2–3 listeners, the overhead of the ForkJoin pool likely exceeds any parallelism benefit, and slow or blocking listeners could starve other pool users.

Consider either:
- Sequential iteration (`listeners.forEach(...)`) — simpler and sufficient when listeners are fast
- A dedicated `ExecutorService` with `CompletableFuture.allOf(...)` — if true parallel execution is important for latency

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
```diff
@@ -0,0 +1,202 @@
...
+                                            rcm,
+                                            executionDAOFacade,
+                                            notifierProperties
+                                                    .get()
+                                                    .getSubscribedWorkflowStatuses());
+                                })
+                        .orElseThrow(
+                                () ->
+                                        new IllegalStateException(
+                                                "Workflow publisher requested but notification properties not configured. "
```
nthmost-orkes: `notifierProperties.get()` will throw `NoSuchElementException` if the Optional is empty, bypassing the descriptive `orElseThrow` message below. Although `@EnableConfigurationProperties` makes this hard to trigger today, it is fragile. Consider `notifierProperties.map(StatusNotifierNotificationProperties::getSubscribedWorkflowStatuses).orElseThrow(...)` or restructuring to validate up-front.

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
```diff
@@ -0,0 +1,202 @@
...
+                                            props.getTtlDuration());
+
+                                    if (props.getTtlDuration().getSeconds() > 0) {
+                                        return new ArchivingWithTTLWorkflowStatusListener(
+                                                executionDAOFacade, props);
+                                    } else if (props.getWorkflowArchivalType()
+                                            == ArchivingWorkflowListenerProperties.ArchivalType
+                                                    .S3) {
+                                        return new ArchivingWorkflowToS3(executionDAOFacade, props);
+                                    } else {
```
nthmost-orkes: Returning `null` for an unknown type is silent degradation — a typo like `types=kafkaa` will result in that listener being silently dropped with only a WARN log, and no other indication of misconfiguration. Since this is evaluated at startup, consider throwing `IllegalArgumentException` here instead to fail fast and surface the error clearly.

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
akhilpathivada: Agree. I thought of `@EnableConfigurationProperties ` would always save from `notifierProperties.get()` being Empty - so I've written this way. I've restructured it now, according to the comments.

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListener.java
akhilpathivada: Fixed it. Went with option-1, since true parallel execution is not very critical for these kind of features.

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
akhilpathivada: Done. Thanks for calling this out - I've missed considering this, and throwing an expection!

workflow-event-listener/src/main/java/com/netflix/conductor/contribs/listener/composite/CompositeWorkflowStatusListenerConfiguration.java
akhilpathivada: Done

manan164:
Thanks for the contribution! A few things worth addressing before merging:

**Issue 1 — Sequential dispatch (docs say "parallel")**
The PR description and README say listeners run "in parallel" / "simultaneously", but the implementation uses `listeners.forEach(...)` which is strictly sequential. If Kafka is slow or blocked, every listener after it in the list is delayed by the same amount — worst case in exactly the failure scenario this feature is meant to isolate. Either switch to `CompletableFuture`-based parallel dispatch, or correct the docs to say "sequentially".

**Issue 2 — Configuration duplicates listener factory logic**
`CompositeWorkflowStatusListenerConfiguration` manually instantiates `KafkaWorkflowStatusPublisher`, `ConductorQueueStatusPublisher`, etc., duplicating logic that already lives in `KafkaWorkflowStatusPublisherConfiguration`, `ArchivingWorkflowListenerConfiguration`, and so on. This works today since constructor signatures match, but if a new dependency is added to any listener, it must be updated in two places. Consider delegating construction back to the existing config classes rather than reimplementing it here.

**Issue 3 — `Optional` injection doesn't guard what you think**
`CompositeWorkflowStatusListenerConfiguration` declares `@EnableConfigurationProperties` for all the properties types (Kafka, archive, queue, etc.), which means Spring always creates those beans when composite mode is active — the `Optional` parameters will never be `Optional.empty()`. The `orElseThrow` safety net is effectively dead code. In practice, if a user sets `types=kafka` but forgets to configure Kafka properties, they won't see your helpful error message — they'll get a cryptic failure from inside the Kafka producer with default/empty values. Consider validating that the required properties are actually populated (e.g., checking `bootstrapServers` is non-empty) rather than relying on `Optional` presence.

Issue #755, opened 2026-02-04
[FEATURE] Add TaskDef-level control for task status event publishing


akhilpathivada:
### Motivation

Task status event publishing is currently controlled globally via configuration. Workflows already support per-definition control through `workflowStatusListenerEnabled` in WorkflowDef, but tasks lack equivalent functionality in TaskDef.

### Proposed Solution
1. Add one field to TaskDef:
```
@ProtoField(id = 23)
private boolean taskStatusListenerEnabled = true;  // Default: true
```
2. Add "IfEnabled" wrapper methods to TaskStatusListener (mirrors existing WorkflowStatusListener pattern):
```
default void onTaskScheduledIfEnabled(TaskModel task) {
    TaskDef taskDef = task.getTaskDefinition();
    if (taskDef != null && taskDef.isTaskStatusListenerEnabled()) {
        onTaskScheduled(task);
    }
}
// ... similar for onTaskCompleted, onTaskFailed, etc.
```
Update WorkflowExecutorOps to call `onTaskScheduledIfEnabled(`) instead of `onTaskScheduled()` etc.

### Task Coverage
This feature works for tasks that have a TaskDef:

1. SIMPLE tasks - Always have TaskDef (full support)
2. System tasks - Only if you register a TaskDef for them
3. Operators - Not applicable (control flow only)

**Note:** System tasks (HTTP, KAFKA_PUBLISH, etc.) work without TaskDef by default. Register a TaskDef to control their events.

### Backward Compatibility

Fully backward compatible - no breaking changes:
- Default value is `true` → existing behavior preserved
- Tasks without TaskDef → continue publishing events as before
- Existing TaskDefs → automatically get `taskStatusListenerEnabled: true`
- No migration needed → adopt incrementally when needed

### Usage Example
```
{
  "name": "process_payment",
  "taskStatusListenerEnabled": false, // or true
  "retryCount": 1,
  "timeoutSeconds": 30
}
```

We'd appreciate your feedback before starting the implementation.

akhilpathivada:
@v1r3n We'd like to hear your feedback on this proposal.

PR #756, opened 2026-02-04
feat: Add TaskDef-level control for task status event publishing

akhilpathivada:
Resolves https://github.com/conductor-oss/conductor/issues/755

### Summary

Adds `taskStatusListenerEnabled` field to `TaskDef` for per-task control over status event publishing.

### Implementation

**Core Changes:**

1. `TaskDef.java` - Added `taskStatusListenerEnabled` boolean field (default: `true`)
2. `TaskStatusListener.java` - Added `onTask*IfEnabled()` wrapper methods for all 9 status callbacks
3. `WorkflowExecutorOps.java` - Updated to use wrapper methods
4. `taskdef.proto` + `AbstractProtoMapper.java` - Added gRPC/protobuf support

**Tests:**
- `TaskStatusListenerTest.java` - Unit tests covering enabled, disabled, and null TaskDef scenarios

### Backward Compatibility

Fully backward compatible:
- Default value is `true` 
- Tasks without registered TaskDef default to enabled via `.orElse(true)`

### Example Usage

```
{
  "name": "high_frequency_poll_task",
  "taskStatusListenerEnabled": false
}
```


v1r3n:
Hi @akhilpathivada we recently added new fields for this in WorkflowTask
https://github.com/conductor-oss/conductor/blob/main/common/src/main/java/com/netflix/conductor/common/metadata/workflow/WorkflowTask.java#L154
I suggest leveraging this to control at a task level.  We can certainly pull this up to TaskDef, as well so it retains hierarchal model.

akhilpathivada:
Hi @v1r3n Thanks for the feedback! I want to make sure I understand the hierarchical model you're suggesting.

I reviewed `onStateChange` in `WorkflowTask` - it's a `Map<String, List<StateChangeEvent>>` that appears designed for custom events, so I'm not sure how to leverage it for controlling `TaskStatusListener` events. Could you clarify:

1. How to leverage `onStateChange` for task-level control? If this is the approach, could you share what you have in mind - I'd like to understand your idea better.
2. Or should I add a `taskStatusListenerEnabled` field to `WorkflowTask` for task-level control, and then pull it up to `TaskDef` as well to retain the hierarchical model (`WorkflowTask` overrides `TaskDef`)?

v1r3n:
I think task level state changes will not mix well with this, approving this PR so we can continue with this.

v1r3n (approved):

PR #766, opened 2026-02-13
Fix: Prevent duplicate workflow termination notifications

fw-himanshu:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

This PR adds a guard clause to prevent duplicate workflow termination operations when a workflow is already in `TERMINATED` status.

**Problem:**
Workflows that were already in `TERMINATED` state were being terminated again, resulting in duplicate workflow termination notifications being sent repeatedly.

**Solution:**
Added an early return check in the `terminateWorkflow()` method to verify if the workflow is already terminated before proceeding with termination operations. If the workflow status is already `TERMINATED`, the method returns immediately without performing any additional termination actions or sending notifications.

**Changes:**
- Added status check for `WorkflowModel.Status.TERMINATED` in `WorkflowExecutorOps.terminateWorkflow()`
- Returns early if workflow is already terminated, preventing redundant operations and duplicate notifications
- Maintains existing behavior: COMPLETED workflows still throw `ConflictException` as expected

Issue #

Alternatives considered
----

1. **Notification deduplication at the messaging layer**: Instead of preventing the termination operation, we could deduplicate notifications at the notification service level. However, this approach would still perform unnecessary termination operations and might not prevent all side effects.

2. **Idempotency token**: Track termination requests with idempotency tokens to prevent duplicate processing. This adds complexity and requires additional state management without addressing the root cause of redundant termination calls.

3. **Current approach (chosen)**: Add early return for already-terminated workflows. This is the simplest solution that prevents redundant operations at the source, eliminating duplicate notifications and improving efficiency with minimal code changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,19 +24,21 @@ jobs:
           distribution: "zulu"
           java-version: "21"
       - name: Cache SonarCloud packages
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ~/.sonar/cache
           key: ${{ runner.os }}-sonar
           restore-keys: ${{ runner.os }}-sonar
       - name: Cache Gradle packages
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             ~/.gradle/caches
             ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: ${{ runner.os }}-gradle-
+      - name: Force Docker API Version
+        run: echo 'api.version=1.44' > ~/.docker-java.properties
       - name: Build with Gradle
         if: github.ref != 'refs/heads/main'
         env:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,7 +24,7 @@ jobs:
           distribution: 'zulu'
           java-version: '21'
       - name: Cache Gradle packages
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             ~/.gradle/caches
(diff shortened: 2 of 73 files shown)


Robban1980:
I don't see any tests for your change. 
How do you verify that all execution paths still work as expected with no side effects?

fw-himanshu:
> I don't see any tests for your change. How do you verify that all execution paths still work as expected with no side effects?

@Robban1980 Added some tests to check both the use cases.

nthmost-orkes:
## Review Notes

**Verdict: Low risk, good fix. Approving for merge.**

### What the change does

Adds an early return in `terminateWorkflow(String workflowId, String reason)` when the workflow is already in `TERMINATED` state, preventing redundant DB updates, queue removals, and duplicate notifications.

### Analysis

**The silent return vs. exception behavior is intentional and correct.** The existing `COMPLETED` guard throws a `ConflictException` because terminating a completed workflow is a logical error. Duplicate termination of an already-TERMINATED workflow is idempotent — not a conflict — so a silent return is the right semantic here.

**Other terminal states (`FAILED`, `TIMED_OUT`) are not guarded by this PR.** The inner `terminateWorkflow(WorkflowModel, ...)` overload already handles these at line 620 (`if (!workflow.getStatus().isTerminal())`), which skips re-setting status, but does not short-circuit the full termination logic (output update, queue cleanup, notifications). This is pre-existing behavior and out of scope for this fix.

**Sub-workflow impact:** If a parent workflow triggers `terminateWorkflow` on a child that is already TERMINATED, the call now returns cleanly instead of re-notifying. This is the desired behavior.

### Test coverage

Two new tests are solid:
- `testTerminateAlreadyTerminatedWorkflow` — verifies the guard fires correctly and no side effects occur
- `testTerminateWorkflowIdempotencyWithNoSideEffects` — verifies the double-call sequence (RUNNING → TERMINATED → second call) is idempotent

### Follow-up (out of scope here)

For completeness, `FAILED` and `TIMED_OUT` workflows could theoretically hit the same duplicate-notification issue via the public entry point. Worth a separate issue/PR if that scenario has been observed in practice.

Issue #768, opened 2026-02-13
Redundant queue ack in getLastPollTask() — wasted backend call on every single task poll


mohammaddanishali-bit:
**Summary**

While tracing the task polling flow step by step, I found that ExecutionService.getLastPollTask() performs a redundant ackTaskReceived() call that results in a wasted queue backend round-trip on every single-poll request.

⸻

**The Problem**
getLastPollTask() delegates to poll() and then acks the returned task:

`public Task getLastPollTask(String taskType, String workerId, String domain) {
    List<Task> tasks = poll(taskType, workerId, domain, POLL_COUNT_ONE, POLLING_TIMEOUT_IN_MS);
    // ...
    Task task = tasks.get(0);
    ackTaskReceived(task);   // ← redundant
    return task;
}`

But **poll()** already acks all returned tasks at the end of its execution:

`public List<Task> poll(String taskType, String workerId, String domain, int count, int timeout) {
    // ... pop from queue, validate, set IN_PROGRESS, persist ...
    tasks.forEach(this::ackTaskReceived);   // ← already acked here
    return tasks;
}`

So the same task gets acked twice — once inside poll(), then again inside getLastPollTask().

**Verification**

Added instrumentation logs and confirmed with a running instance (Redis queue + Postgres persistence):

`STEP-1 POLL:          taskType=process_payment, poppedFromQueue=1
STEP-2 FIRST-ACK:     acking task=98b08e16-... (inside poll())
STEP-3 REDUNDANT-ACK: acking task=98b08e16-... AGAIN (inside getLastPollTask())`

The double call is confirmed by the code flow itself — poll() acks via tasks.forEach(this::ackTaskReceived) at line 206, then getLastPollTask() calls ackTaskReceived(task) again at line 229 on the same task object.
Expected behavior by backend on the redundant second ack.

**Impact**

This affects every GET /api/tasks/poll/{tasktype} request — the most common poll endpoint used by workers. Each poll results in one wasted queue backend call:

### Backend Ack Behavior Comparison

| Backend              | First ack (inside `poll()`)        | Second ack (inside `getLastPollTask()`) |
|----------------------|------------------------------------|------------------------------------------|
| PostgresQueueDAO     | DELETE succeeds → true             | Row already deleted → false (wasted)     |
| MySQLQueueDAO        | DELETE succeeds → true             | Row already deleted → false (wasted)     |
[...]

PR #770, opened 2026-02-13
Remove redundant ackTaskReceived() call in getLastPollTask()

mohammaddanishali-bit:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

This PR fixes a bug where `ackTaskReceived()` was being called twice for the same task in the `getLastPollTask()` method of `ExecutionService`.

**The Problem:**
- `getLastPollTask()` internally calls `poll()`, which already acknowledges tasks via `tasks.forEach(this::ackTaskReceived)`
- After `poll()` returns, `getLastPollTask()` was redundantly calling `ackTaskReceived()` again on the same task
- This resulted in duplicate acknowledgments for every task polled through `getLastPollTask()`

**The Solution:**
- Removed the redundant `ackTaskReceived()` call in `getLastPollTask()` (line 222 of ExecutionService.java)
- The task acknowledgment now happens only once, inside the `poll()` method where it belongs

**Testing:**
- Added unit test `testGetLastPollTaskAcksOnlyOnce()` that verifies `queueDAO.ack()` is called exactly once when a task is successfully polled
- Added unit test `testGetLastPollTaskReturnsNullWhenEmpty()` that verifies no acknowledgment occurs when the queue is empty
- Both tests use Mockito to verify the exact number of acknowledgment calls

Issue #768

Alternatives considered
----

**Alternative 1:** Remove the ack call from `poll()` instead
- Rejected because `poll()` is a public method used by multiple callers, and removing the ack there would break other parts of the system that depend on tasks being acknowledged during polling

**Alternative 2:** Add a flag to `poll()` to control whether it should ack
- Rejected as it would complicate the API unnecessarily. The simpler solution is to remove the redundant call in the wrapper method

**Alternative 3:** Make `ackTaskReceived()` idempotent
- This would mask the underlying issue rather than fixing it. The root cause is the duplicate call, which should simply be removed

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
+++ b/core/src/main/java/com/netflix/conductor/service/ExecutionService.java
@@ -219,7 +219,7 @@ public class ExecutionService {
             return null;
         }
         Task task = tasks.get(0);
-        ackTaskReceived(task);
+
         LOGGER.debug(
                 "The Task {} being returned for /tasks/poll/{}?{}&{}",
                 task,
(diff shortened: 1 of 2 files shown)


manan164 (approved):

nthmost-orkes:
docker related test failure crashing the build again

PR #801, opened 2026-03-01
AI task tracking, module reorganization, and new REST endpoints

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
Summary                                                                                                                                                                                                       

  - AI token usage tracking: Thread Task through all LLM helper methods (chatComplete, generateImage, generateEmbeddings, generateAudio, generateVideo, checkVideoStatus) so taskId is captured in TokenUsageLog
  for per-task cost attribution
  - Robustness fixes in LLMHelper: Handle null/empty LLM responses in extractResponse, add error logging for JSON conversion failures, guard against null toolCalls list, and fix MIME type resolution when
  mimeType is null
  - FileSystemDocumentLoader: Restrict supports() to only match file:// URLs instead of unconditionally returning true
  - Module reorganization: Move NonRetryableException, TaskContext, AnnotatedSystemTaskWorker from core to common, and TokenUsageLog from ai to common for broader reuse
  - Version endpoint: Add VersionService and VersionResource (GET /api/version) to expose the server's implementation version
  - Workflow tasks endpoint: Add GET /api/workflow/{workflowId}/tasks with pagination (start, count) and optional status filtering
  - TaskMapperValidator interface: New validation contract for task mappers (getTaskType(), validate())
  - Legacy embeddings worker: Add deprecated LLM_GET_EMBEDDINGS worker alias that delegates to searchUsingEmbeddings for backward compatibility
  - gRPC schema: Add items field (field 33) to WorkflowTask proto and update AbstractProtoMapper accordingly
  - Build version: Set release version to 3.23.0-rc1

  Test plan

  - Verify AI task workers log taskId in token usage output
  - Confirm null/empty LLM responses are handled gracefully without exceptions
  - Test GET /api/version returns the server version string
  - Test GET /api/workflow/{id}/tasks with pagination and status filters
  - Validate gRPC WorkflowTask serialization round-trips the items field correctly
  - Ensure legacy LLM_GET_EMBEDDINGS task type still works
  - Verify FileSystemDocumentLoader no longer matches non-file:// URLs
  - Confirm moved classes resolve correctly from common module in dependent modules

Issue #

Alternatives considered
----

### DIFF
--- a/rest/src/main/java/com/netflix/conductor/rest/controllers/WorkflowResource.java
+++ b/rest/src/main/java/com/netflix/conductor/rest/controllers/WorkflowResource.java
@@ -233,9 +233,9 @@ public class WorkflowResource {
         }
 
         int totalHits = workflowFilteredTasks.size();
-        int lastIndex = start + count;
-        int toIndex = lastIndex > totalHits - 1 ? totalHits : lastIndex;
-        List<Task> requestedSubList = workflowFilteredTasks.subList(start, toIndex);
+        int fromIndex = Math.min(start, totalHits);
+        int toIndex = Math.min(start + count, totalHits);
+        List<Task> requestedSubList = workflowFilteredTasks.subList(fromIndex, toIndex);
         return new SearchResult<>(totalHits, requestedSubList);
     }
 


ai/src/main/java/org/conductoross/conductor/ai/LLMHelper.java
```diff
@@ -155,13 +160,15 @@ public LLMResponse generateImage(
     }
 
     public List<Float> generateEmbeddings(
+            Task task,
```
manan164, line 163: Do we need task argument here? It is not being used.

v1r3n: Yes for the consistency and also for Orkes where we have integrations, the task is used.

rest/src/main/java/com/netflix/conductor/rest/controllers/WorkflowResource.java
```diff
@@ -207,6 +209,36 @@ public List<Workflow> getWorkflows(
...
+                                            status.stream()
+                                                    .map(String::toUpperCase)
+                                                    .anyMatch(s -> t.getStatus().name().equals(s)))
+                            .collect(Collectors.toList());
+        }
+
+        int totalHits = workflowFilteredTasks.size();
+        int lastIndex = start + count;
+        int toIndex = lastIndex > totalHits - 1 ? totalHits : lastIndex;
+        List<Task> requestedSubList = workflowFilteredTasks.subList(start, toIndex);
```
manan164: We need a condition check between start and totalHits before invoking subList. totalHits must be > start.

PR #823, opened 2026-03-05
feat: optimize indexing throughput for high-volume bulk workloads

bradyyie:
## Summary

- Route `indexWorkflow()` through the bulk-buffered path instead of single-doc `IndexRequest` calls, eliminating per-document HTTP round-trips that were the primary throughput bottleneck
- Tune default configuration for sustained 200+ docs/sec with 300KB+ documents: `indexBatchSize` 1->100, `asyncMaxPoolSize` 12->24, `asyncWorkerQueueSize` 100->500, `indexReplicasCount` 0->1 (OS v2/v3)
- Add configurable `indexRefreshInterval` (default 5s) and `translogFlushThresholdSize` (default 1gb) properties, applied at index creation time in `addIndex()`, to reduce segment creation and flush frequency

### Problem
Workflow indexing used direct single-doc `client.index()` calls, bypassing the existing bulk buffer infrastructure. With ~300KB documents at 200+ docs/sec, this created:
- One HTTP round-trip per document (instead of batched bulk requests)
- Excessive segment fragmentation from 1s default refresh interval
- Frequent translog flushes at the default 512MB threshold

### Changes
Applied consistently across all three persistence modules (`es7-persistence`, `os-persistence-v2`, `os-persistence-v3`):

**DAO changes:**
- `indexWorkflow()` now calls `indexObject()` (bulk-buffered path) instead of direct `client.index()`
- Both `addIndex()` overloads (with/without mapping) include `refresh_interval` and `translog.flush_threshold_size` in index creation settings

**Properties changes:**
- `indexBatchSize`: 1 -> 100
- `asyncMaxPoolSize`: 12 -> 24
- `asyncWorkerQueueSize`: 100 -> 500
- `indexReplicasCount`: 0 -> 1 (OS v2/v3; ES7 was already 1)
- New: `indexRefreshInterval` (default `"5s"`)
- New: `translogFlushThresholdSize` (default `"1gb"`)
- Legacy `conductor.elasticsearch.*` property fallback supported in OS v2/v3

All settings remain configurable via Spring properties for environment-specific tuning.


nthmost-orkes (commented):
Thanks for this — routing `indexWorkflow()` through the bulk-buffered path is the right call, and doing it consistently across ES7, OS v2, and OS v3 is exactly the right scope. The default tuning (`indexBatchSize` 1→100, pool/queue bumps, `indexRefreshInterval`, `translogFlushThresholdSize`) makes sense for a high-throughput production setup.

One gap before this is ready: **ES8 has the same problem and the bulk infrastructure to fix it is already there.**

In `ElasticSearchRestDAOV8.indexWorkflow()` (line 317), there's the same direct synchronous `executeWithRetry(elasticSearchClient.index(...))` call that this PR replaces in the ES7 DAO. Every other write method in the ES8 DAO (`indexTask()`, `addMessage()`, `addEventExecution()`) already routes through `indexObject()` → `Es8BulkIngestionSupport.indexObject()` → `BulkIngester`. `indexWorkflow()` is the only one that doesn't. The fix is two lines — replace the `executeWithRetry` block with:

```java
Refresh refresh = properties.isWaitForIndexRefresh() ? Refresh.WaitFor : null;
indexObject(workflowIndexName, WORKFLOW_DOC_TYPE, workflowId, workflow, refresh);
```

ES8 already has `indexBatchSize` (default 1), `asyncMaxPoolSize` (default 12), `asyncWorkerQueueSize` (default 100), and `asyncBufferFlushTimeout` — they just need the same default bumps you applied to ES7/OS.

One thing to leave alone: ES8's `indexRefreshInterval` is already a 30s `Duration` applied via an index component template (not at index creation time). No need to replicate the `refresh_interval`/`translog` index-creation-settings pattern from ES7 — that part is already handled differently and correctly for ES8.

If you can extend this PR to cover ES8, it becomes a complete fix across all four persistence backends and I can approve and merge

Issue #825, opened 2026-03-05
Epic: HUMAN task stub implementation and correctness fixes


nthmost-orkes:
## Overview

Conductor OSS includes a minimal **HUMAN task stub** — the task starts `IN_PROGRESS` and waits indefinitely for an external system to update it. This is the correct behavior for OSS: it unblocks the workflow and lets any external system (ticketing, approval portal, etc.) drive task completion.

However, the current stub has correctness bugs that need fixing before the full enterprise-grade HUMAN task implementation can be backported from Orkes Conductor.

## Goals

- Fix the current stub so it behaves correctly (no queue pollution, proper API error codes, searchable in UI)
- Establish a clean foundation for the full Orkes implementation to converge onto
- **Not** in scope: assignment chains, SLA windows, form templates, claim/release workflows (those belong in the full implementation)

## Issues Tracked

| Issue | Description | Priority |
|-------|-------------|----------|
| #795 | Prevent HUMAN tasks from polluting the decider queue | High — CPU waste on every sweeper cycle |
| #783 | Fix 200 OK response on update-by-refName with invalid workflowId/taskRef | Medium — silent data loss |
| #704 | Add HUMAN task type to UI task type filter dropdown | Low — usability gap |

## Architecture Notes (Orkes Convergence)

The full enterprise HUMAN task in Orkes Conductor has:
- Assignment chains with SLA windows (PENDING → ASSIGNED → IN_PROGRESS → COMPLETED)
- Claim/release/skip/reassign REST endpoints (`/api/human/tasks/...`)
- PostgreSQL persistence table (`human_task_v2`) with JSONB and tsvector search
- UI form templates for task display
- State-change workflow triggers

The OSS stub in `Human.java` + `HumanTaskMapper.java` is intentionally simple. The convergence path is:
1. Fix correctness issues (this epic)
2. Port `OrkesHuman.java` state machine + service layer to OSS (future epic)
3. Add REST endpoints for HUMAN task management (future epic)
4. UI integration (future epic)

## Acceptance Criteria

- [ ] #795: Workflows blocked on HUMAN tasks are removed from the decider queue; re-queued on HUMAN task completion
- [ ] #783: `POST /api/tasks/{workflowId}/{taskRefName}/{status}` returns 404 when workflowId or taskRefName is not found
- [ ] #704: HUMAN appears in the UI task type filter dropdown alongside other system task types

mikebski:
[...]

PR #826, opened 2026-03-05
fix: prevent HUMAN tasks from churning the decider queue

### LINKED ISSUE #795: Prevent HUMAN tasks from polluting the decider queue
## Summary

Workflows with IN_PROGRESS HUMAN tasks remain in the decider queue, causing unnecessary processing cycles. HUMAN tasks are long-lived and don't need continuous polling — the workflow should be removed from the queue while waiting and re-queued when the HUMAN task completes.

## Current State

Investigation shows this was partially addressed in earlier commits but has **regressed** — the HUMAN task handling logic is currently missing from `ExecutorUtils.computePostpone()` in the new sweeper path.

## Desired Behavior

- Remove workflow from the decider queue when it's blocked on a HUMAN task
- Re-queue the workflow when the HUMAN task completes (push back to decide queue from `WorkflowExecutorOps`)
- Add a configurable flag (e.g., `humanTaskPreventsDeciderQueue`) in `ConductorProperties`

## Affected Files

- `core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java`
- `core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java`
- `core/src/main/java/org/conductoross/conductor/core/execution/ExecutorUtils.java`
- `core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java`

## Prior Art

Community PR #701 by @martinconstam implemented this cleanly with tests (76 additions). Note: a regression may have occurred when the new sweeper was introduced (PR #703).

nthmost-orkes:
## Summary

Fixes #795 (part of epic #825: HUMAN task stub implementation)

Two bugs caused workflows blocked on a HUMAN task to be re-swept far more frequently than intended:

1. **`ExecutorUtils.computePostpone()` had no HUMAN task branch** — `postponeDurationSeconds` stayed at `0`, so the method returned `workflowOffsetTimeoutSeconds` unchanged. Same issue existed for WAIT tasks with no timeout.

2. **`WorkflowExecutorOps.decide()` only called `setUnackTimeout` when `computePostpone()` returned a value *different* from the default `workflowOffsetTimeout`** — when they were equal (HUMAN, WAIT-no-timeout, SIMPLE with no response timeout), no explicit unack timeout was set. This left the queue backend's own default unack interval in control, which can be much shorter than `workflowOffsetTimeout`.

## Changes

- **`ExecutorUtils.computePostpone()`**: Added explicit branches for `TASK_TYPE_HUMAN` and WAIT tasks with no timeout, both returning `workflowOffsetTimeoutSeconds`. This is parity with the legacy sweeper's `unack()` implementation.

- **`WorkflowExecutorOps.decide()`**: Removed the `updatedOffset != timeout` guard. `setUnackTimeout` is now always called for non-terminal workflows so the decider queue position is always explicitly managed.

- **`ExecutorUtilsTest`**: New test class covering HUMAN, WAIT (with/without timeout), and default task scenarios.

## Re-queue on completion

When a HUMAN task is completed externally via the task update API, `updateTask()` already calls `decide()`, which re-queues the workflow with immediate priority. No additional change needed.

## Test plan

- [x] `ExecutorUtilsTest` passes (5 new tests)
- [x] Full `conductor-core` test suite passes

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
@@ -1192,19 +1192,18 @@ public class WorkflowExecutorOps implements WorkflowExecutor {
                 Duration timeout = properties.getWorkflowOffsetTimeout();
                 if (!workflow.getStatus().isTerminal()) {
                     Duration updatedOffset = computePostpone(workflow, timeout);
-                    if (updatedOffset.getSeconds() != timeout.getSeconds()) {
-                        // we have a new value, setUnack uses time in millis
-                        LOGGER.debug(
-                                "Pushing the workflow {} into decider queue by {} millis",
-                                workflow.getWorkflowId(),
-                                updatedOffset.getSeconds() * 1000);
-                        queueDAO.setUnackTimeout(
-                                DECIDER_QUEUE,
-                                workflow.getWorkflowId(),
-                                updatedOffset.getSeconds() * 1000);
-                    }
+                    // Always explicitly set the unack timeout so queue-backend defaults do not
+                    // determine re-sweep frequency (e.g. HUMAN/WAIT tasks would otherwise be
+                    // re-swept at the queue's own default unack interval, which may be very short).
+                    LOGGER.debug(
+                            "Pushing the workflow {} into decider queue by {} millis",
+                            workflow.getWorkflowId(),
+                            updatedOffset.getSeconds() * 1000);
+                    queueDAO.setUnackTimeout(
+                            DECIDER_QUEUE,
+                            workflow.getWorkflowId(),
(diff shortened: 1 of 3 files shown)


martinconstam:
FYI — I opened #1171 for the same issue (#795). It takes the remove-and-re-queue approach from the #795 acceptance criteria (remove the workflow from the decider queue while it is blocked on an `IN_PROGRESS` HUMAN task, then re-queue it when the task completes via `WorkflowExecutorOps#updateTask`) rather than only lengthening the postpone interval, so blocked workflows leave the queue entirely instead of being re-swept every offset. Happy to consolidate if you prefer one approach over the other.

nthmost-orkes:
superseded

PR #855, opened 2026-03-12
feat(postgres): make Flyway repairOnMigrate and validateOnMigrate configurable

### LINKED ISSUE #933: Flyway V9 checksum mismatch on upgrade due to modified V9__indexing_index_fix.sql
## Problem

When upgrading between Conductor versions, `V9__indexing_index_fix.sql` was modified (the old version had all SQL statements commented out, the new version has them enabled), causing a Flyway checksum mismatch that prevents startup:

```
Migration checksum mismatch for migration version 9
-> Applied to database : 1904511200
-> Resolved locally    : -975282508
```

Additionally, the Flyway instance in `PostgresConfiguration` is created manually via `Flyway.configure()`, bypassing Spring Boot's `FlywayAutoConfiguration`. This means standard `spring.flyway.*` properties (such as `repair-on-migrate`) have no effect, leaving no easy way to resolve the mismatch without manually updating the `flyway_schema_history` table.

## Workaround

PR #855 adds two configurable properties under `conductor.postgres.*`:

- **`flywayRepairOnMigrate`** (default: `false`) — automatically repairs the schema history on checksum mismatches.
- **`flywayValidateOnMigrate`** (default: `true`) — controls whether Flyway validates migration checksums before running.

Users upgrading from older versions can set:

```properties
conductor.postgres.flyway-repair-on-migrate=true
```

### Notes

- `repair()` runs on **every startup** when the flag is enabled, not just during an upgrade. It is idempotent, but users should disable it after the upgrade is complete.
- `flywayValidateOnMigrate=false` bypasses Flyway's checksum safety entirely and should be treated as a **last resort**, not a routine setting.

Ruiruiz30:
## Summary

The Flyway instance in `PostgresConfiguration` is created manually via `Flyway.configure()`, bypassing Spring Boot's `FlywayAutoConfiguration`. This means standard `spring.flyway.*` properties (such as `repair-on-migrate`) have no effect on Conductor's Flyway behavior.

This becomes a real problem when upgrading between Conductor versions. `V9__indexing_index_fix.sql` was modified after its initial release: the index `task_index_json_data_json_idx` was renamed to `task_index_json_data_fulltext_idx` (commit `26af4cb`). This changed the file's checksum, causing a mismatch for any instance that had already applied the original V9:

```
Migration checksum mismatch for migration version 9
-> Applied to database : 1904511200
-> Resolved locally    : -975282508
```

Currently the only workaround is to manually update the `flyway_schema_history` table in the database, which is not ideal for production environments.

Closes #933

## Changes

Add two configurable properties to `PostgresProperties` (under `conductor.postgres.*`):

- **`flywayRepairOnMigrate`** (default: `false`) — when set to `true`, Flyway will automatically repair the schema history on checksum mismatches during migrate. A `WARN` log is emitted at startup as a reminder to disable this once the issue is resolved.
- **`flywayValidateOnMigrate`** (default: `true`) — controls whether Flyway validates migration checksums before running. A `WARN` log is emitted when set to `false` to indicate that checksum safety is disabled.

These properties are wired into `PostgresConfiguration.flywayForPrimaryDb()`.

### Usage

Users upgrading from older Conductor versions can add this to their `config.properties`:

```properties
conductor.postgres.flyway-repair-on-migrate=true
```

## Backward Compatibility

- Both properties use safe defaults (`repairOnMigrate=false`, `validateOnMigrate=true`), matching the current hardcoded behavior.
- No impact on existing deployments.


nthmost-orkes:
Implementation looks correct — repair() runs inside the factory method before Spring calls migrate() via @Bean(initMethod = "migrate"), so the ordering is right, and the safe defaults mean no behavior change for existing deployments. Two small things worth addressing:

repair() runs on every startup when the flag is enabled, not just during an upgrade. It's idempotent so it won't cause problems, but users who turn it on expecting a one-time fix will leave it permanently enabled without realizing it. Worth calling that out in the property comment.

validateOnMigrate=false bypasses Flyway's checksum safety entirely — a corrupted or wrong migration won't be caught. It's a valid escape hatch but the comment should warn it's a last resort, not a routine setting.

Also worth filing an issue for this and linking it from the PR — helps other users track the V9 checksum mismatch problem and the fix.

Ruiruiz30:
Thanks for the review! Both suggestions have been addressed:

1. Updated the Javadoc for `flywayRepairOnMigrate` to clarify that repair runs on every startup while enabled, and users should disable it after the upgrade is complete.
2. Updated the Javadoc for `flywayValidateOnMigrate` to warn that setting it to `false` bypasses checksum safety entirely and should be used as a last resort only.

Also created issue #933 to track the V9 checksum mismatch problem, and linked it from this PR.

Ruiruiz30:
I've fixed the Spotless formatting issue in `PostgresProperties.java` — the Javadoc comment has been reflowed to match the expected line wrapping. Please re-run CI. 🙏

postgres-persistence/src/main/java/com/netflix/conductor/postgres/config/PostgresConfiguration.java
```diff
@@ -71,12 +71,20 @@ public Flyway flywayForPrimaryDb() {
...
-                .outOfOrder(true)
-                .baselineOnMigrate(true)
-                .load();
+        Flyway flyway =
+                config.configuration(Map.of("flyway.postgresql.transactional.lock", "false"))
+                        .schemas(properties.getSchema())
+                        .dataSource(dataSource)
+                        .outOfOrder(true)
+                        .baselineOnMigrate(true)
+                        .validateOnMigrate(properties.isFlywayValidateOnMigrate())
```
mp-orkes, line 85: Could we log a WARN when `flywayRepairOnMigrate=true` or `flywayValidateOnMigrate=false` are set?

Repair is meant to be a one-time fix. It should be disabled after the upgrade, but in practice people forget. A startup warning makes the risk a bit more visible and gives us a paper trail that they were informed.

Something like:

```shell
WARN: flywayRepairOnMigrate is enabled. Flyway will repair schema history on every startup. Disable this property once your migration is complete.
WARN: flywayValidateOnMigrate is disabled. Flyway will not validate migration checksums. This bypasses an important safety check and should only be used as a last resort.
```

mp-orkes (changes requested):
We need at least a `WARN` log for these properties. They give you flexibility but are also a footgun.

Why?

An upgraded instance with `repairOnMigrate=true` will have its checksum updated and start successfully, but the DDL statements in the new migration version never run. That instance silently ends up with a different schema than a fresh install.

---

Ideally migrations should never be modified, but it happened. 

Looking at the git log of [V9__indexing_index_fix.sql](https://github.com/conductor-oss/conductor/commits/main/postgres-persistence/src/main/resources/db/migration_postgres/V9__indexing_index_fix.sql) it seems the PR description is slightly inaccurate.

> _the old version had all SQL statements commented out, the new version has them enabled_

The original V9 was never commented out. The actual change between the two versions was just an index rename.

What am I missing here?

Ruiruiz30:
Thanks for the review, @mp-orkes.

**WARN logs** have been added for both properties — `flywayRepairOnMigrate=true` will print a `WARN` at startup reminding users to disable it once the checksum mismatch is resolved, and `flywayValidateOnMigrate=false` will print a `WARN` indicating that checksum safety is disabled.

**On the V9 description** — you're right, I got that wrong. Looking at the git history, the original V9 was never commented out. The actual change was an index rename: `task_index_json_data_json_idx` → `task_index_json_data_fulltext_idx` (commit `26af4cb`). That rename modified the file content and caused the checksum mismatch. I've updated the PR description to reflect this accurately.

nthmost-orkes:
Yes, footgun is real, though addressed w/ a WARN -- at this point @Ruiruiz30 you should address @mp-orkes directly and re-request review.

PR #880, opened 2026-03-18
fix: set subworkflowChanged=true in updateParentWorkflowTask to fix race with sweeper

nthmost-orkes:
## Summary

- **Root cause**: `WorkflowSweeper` background threads continuously poll `DECIDER_QUEUE` and call `decide()` → `adjustStateIfSubWorkflowChanged`, which resets `subworkflowChanged=false`. When a subworkflow completes later (`completeWorkflow` → `updateParentWorkflowTask`), the parent SUB_WORKFLOW task was updated but `subworkflowChanged` was never re-set to `true`, leaving the parent stuck.
- **Fix**: `updateParentWorkflowTask` now sets `subworkflowChanged=true` after updating the task. This is semantically correct — the subworkflow just transitioned to a terminal state, so the parent needs to re-evaluate.
- **Test coverage added**: Made `adjustStateIfSubWorkflowChanged` package-private (`@VisibleForTesting`) and added two unit tests covering the JOIN task boundary conditions:
  - `testAdjustStateForSubworkflowChangedDoesNotResetInProgressJoin` — verifies IN_PROGRESS JOIN tasks are NOT re-queued
  - `testAdjustStateForSubworkflowChangedResetsCanceledJoin` — verifies CANCELED JOIN tasks ARE reset to IN_PROGRESS and re-queued (the retry path)

## Failing test fixed

`SubWorkflowSpec > Test retrying a subworkflow where parent workflow timed out due to workflowTimeout`

Line 191 asserted `subworkflowChanged == true` after the retried subworkflow completed, but the sweeper had already reset the flag to `false` before completion.

## Test plan

- [ ] Run `SubWorkflowSpec` integration test — previously failing assertion at line 191 now passes
- [ ] Run `TestWorkflowExecutor#testUpdateParentWorkflowTask` — confirms `subworkflowChanged=true` is set
- [ ] Run `TestWorkflowExecutor#testAdjustStateForSubworkflowChangedDoesNotResetInProgressJoin` — new test
- [ ] Run `TestWorkflowExecutor#testAdjustStateForSubworkflowChangedResetsCanceledJoin` — new test


nthmost-orkes:
Closing — this isn't the right fix for the fork/join JOIN flake. #1206 established (and the maintainer confirmed) that the sweeper reopening a CANCELED JOIN to IN_PROGRESS is *expected* async behavior, not a product bug; the fix is to relax the too-early test snapshot. This PR does the opposite — it makes `adjustStateIfSubWorkflowChanged` reset CANCELED JOINs to IN_PROGRESS more deterministically (see its own `testAdjustStateForSubworkflowChangedResetsCanceledJoin`), which pushes *toward* the state the flaky assertions reject. If the separate parent-stuck-after-subworkflow-timeout scenario it mentions is a real standalone bug, it should be pursued as its own focused PR with that as the headline.

Issue #888, opened 2026-03-19
EPIC: Implement WAIT_FOR_WEBHOOK task type


nthmost-orkes:
## Overview

Add a functional `WAIT_FOR_WEBHOOK` system task to Conductor OSS, resolving a longstanding gap between the UI (which already renders webhook task configuration) and the backend (which has no implementation). Closes #528.

All work was completed in a single PR (#894) rather than the originally planned six incremental PRs.

---

## Status

✅ **Completed — see PR #894**

---

### New module: `webhook-task/`

```
webhook-task/
  src/main/java/org/conductoross/conductor/tasks/webhook/
    WaitForWebhookTask.java              # WorkflowSystemTask — start/execute/cancel
    WebhookTaskDAO.java                  # Interface: hash → taskId routing storage
    InMemoryWebhookTaskDAO.java          # Default single-node impl (@ConditionalOnMissingBean)
    WebhookConfig.java                   # Configuration model (ported from Orkes, no orgId)
    WebhookConfigDAO.java                # Interface: config + matcher index storage
    InMemoryWebhookConfigDAO.java        # Default single-node impl
    WebhookConfigService.java            # Config CRUD + matcher index computation
    WebhooksConfigResource.java          # CRUD /api/metadata/webhook
    WebhookHashingService.java           # Task-registration hash + inbound-event hash
    WebhookVerifier.java                 # Interface: request verification strategy
    HeaderBasedVerifier.java             # Default: match expected header key/value pairs
    IncomingWebhookEvent.java            # Inbound request capture model
    IncomingWebhookService.java          # Synchronous event processing + task completion
    IncomingWebhookResource.java         # POST /webhook/{id}, GET /webhook/{id}
    WebhookOrgContextProvider.java       # Interface for orgId isolation (no-op in OSS)
    config/WebhookAutoConfiguration.java # Spring auto-configuration
  src/main/java/com/netflix/conductor/core/execution/mapper/
    WaitForWebhookTaskMapper.java        # Task mapper (in conductor-core)
```

### How it works

**1. Task definition — pause a workflow and wait for a matching webhook:**
```json
{
  "name": "wait_for_payment",
  "taskReferenceName": "waitForPayment",
  "type": "WAIT_FOR_WEBHOOK",
  "inputParameters": {
    "matches": {
      "$['event']['type']": "payment.completed",
      "$['data']['orderId']": "${workflow.input.orderId}"
    }
  }
}
```

**2. Webhook configuration:**
```json
{
[...]

PR #894, opened 2026-03-19
feat: WAIT_FOR_WEBHOOK system task — full implementation (epic #888)

nthmost-orkes:
## Summary

**PR 1 of 6** for the `WAIT_FOR_WEBHOOK` epic (#888). This PR establishes the task-type foundation — enough to register and park a workflow on a webhook task, but **not yet to complete one**. Inbound delivery, config CRUD, hashing, and verification land in followup PRs (see below).

## What's in this PR

- **`TaskType.WAIT_FOR_WEBHOOK`** — new task-type enum value
- **`WaitForWebhookTaskMapper`** — maps a workflow task definition into a scheduled task instance; validates required `matches` input
- **`WaitForWebhookTask`** (`start` / `execute` / `cancel`) — extends `WorkflowSystemTask`; on `start`, registers the task with `WebhookTaskDAO` and parks the task in `IN_PROGRESS`. `execute` is a no-op (task completes via inbound delivery in a later PR). `cancel` removes the DAO registration.
- **`WebhookTaskDAO`** interface + **`InMemoryWebhookTaskDAO`** implementation — minimal contract (`put` / `pop` / `popAll` / `remove`) for routing inbound events to parked tasks. In-memory impl is for single-node / test use; Redis impl lands in PR 4.
- **`webhook-task` Gradle module** — new module wired into `server/build.gradle` and `settings.gradle`
- **Tests** — `WaitForWebhookTaskMapperTest`, `InMemoryWebhookTaskDAOTest`, `WaitForWebhookTaskTest`
- **Example workflow** — `order_payment_workflow.json` test fixture

13 files, +877 lines.

## What's **not** in this PR (deferred to followups)

| PR | Issue | Scope |
|----|-------|-------|
| PR 2 | #890 | `WebhookConfig` model + `WebhooksConfigResource` CRUD API |
| PR 3 | #889 | `WebhookHashingService` + `WebhookVerifier` (HEADER_BASED, etc.) |
| PR 4 | #892 | `IncomingWebhookService` + `POST /webhook/{id}` inbound endpoint; Redis DAO implementations |
| PR 5 | #891 | `workflowsToStart` — start new workflow instances from inbound events |
| PR 6 | #893 | End-to-end integration tests + user-facing docs |

Until PR 4 lands, a `WAIT_FOR_WEBHOOK` task will park `IN_PROGRESS` indefinitely — there is no inbound path yet. This PR is intentionally non-functional in isolation; it's structural plumbing that the next PRs build on.

[...]

PR #921, opened 2026-03-25
feat: precision pass — fix GET event dispatch and config-not-found (WAIT_FOR_WEBHOOK epic #888)

nthmost-orkes:
## Summary

Precision pass on the \`webhook-task\` module — fixes two Orkes convergence gaps plus six correctness issues identified in adversarial code review.

### Orkes convergence fixes

- **Fix GET event dispatch**: \`handlePing()\` now treats \`GET /webhook/{id}\` requests where the verifier returns \`null\` (not a recognized ping challenge) as real webhook events — dispatches to waiting \`WAIT_FOR_WEBHOOK\` tasks and triggers \`workflowsToStart\` inline. Previously these events were silently dropped. Matches Orkes Enterprise behavior.

- **Fix config-not-found to return 200**: \`handleWebhook()\` now returns \`""\` (HTTP 200) when no config is found for the given webhook ID, instead of throwing \`IllegalArgumentException\` (HTTP 400). Prevents retry storms when a webhook config is deleted while the provider still has the URL registered.

### Correctness fixes (adversarial review)

1. **GET ping verification gap** — \`handlePing\`'s null-ping branch now calls \`verifier.verify()\` before dispatching. Previously unauthenticated GET requests could trigger workflow dispatch without any verification.

2. **Transaction ordering** — removed redundant \`webhookTaskDAO.remove()\` in \`completeTask\`: \`popAll\` already removed the task ID atomically. The stale final remove could mask \`updateTask\` failures.

3. **Double-remove cleanup** — same: removed redundant removes in null-task and terminal-task guard clauses of \`completeTask\`.

4. **Early verifier validation** — \`WebhooksConfigResource\` now rejects configs at creation/update time if no verifier implementation is registered for the configured type (e.g. STRIPE, TWITTER). Prevents storing configs that silently fail at inbound-event time.

5. **Payload key collision warning** — \`parsePayload\` now logs WARN when a query parameter key overwrites a body field of the same name.

6. **Redis \`popAll\` atomicity** — \`RedisWebhookTaskDAO\` now overrides \`popAll\` with SMEMBERS + DEL, narrowing the race window vs. the default SMEMBERS + N×SREM implementation.

---

### Yellow: Hash normalization asymmetry (requires coordinated OSS + Orkes deploy)

**File:** \`webhook-task/.../WebhookHashingService.java\` and \`redis-persistence/.../RedisWebhookTaskDAO.java\`

**Issue:** Registration hash appends the raw expected value from the task definition; inbound hash appends the extracted actual value from the payload. These diverge when the expected and actual values differ only by case (e.g. `"PENDING"` vs `"pending"`). \`equalsIgnoreCase\` passes but the hashes don't match, so the task is never completed.

**Root cause:** Both OSS (\`WebhookHashingService.computeRegistrationHash\`) and Orkes Enterprise (\`RedisWebhookTaskService.computeHash\`) have the same bug — the asymmetry is in the hash inputs, not the comparison. The fix is to normalize both sides (e.g. \`toLowerCase\`) before appending to the hash.

**Why deferred:** Any normalization change produces different hash strings for existing registrations. Deployed \`WAIT_FOR_WEBHOOK\` tasks registered under old hashes will never match inbound events using new hashes. Requires a coordinated deployment: quiesce active WAIT_FOR_WEBHOOK tasks, deploy OSS + Orkes simultaneously, or use a migration window. Tracked in the Orkes counterpart in \`orkes-io/orkes-conductor\` at \`WebhookHashingService.computeJsonHash\` and \`RedisWebhookTaskService\` lines ~90–95.

---

### Red: Hash delimiter collision (requires data migration)

**File:** \`webhook-task/.../WebhookHashingService.java\`

**Issue:** The routing hash key uses \`;\` as the delimiter between components (\`workflowName;version;taskRefName;value1;value2\`). If any component itself contains \`;\`, the hash becomes ambiguous — two different keys could produce the same hash string, causing cross-routing between unrelated tasks.

**Root cause:** Delimiter is not escaped. A workflow named \`"order;processing"\` would produce the same base key as \`"order"\` with task ref \`"processing;...\`".

**Why deferred:** Changing the delimiter (or adding escaping) produces different hashes for all existing stored keys. All \`WEBHOOK_TASKS.<hash>\` entries in Redis/Postgres written with the old format become unreachable. Requires a data migration script to re-key all stored entries. Additionally, the same delimiter is used in \`orkes-io/orkes-conductor\`'s \`WebhookHashingService\` — any change must be coordinated across both codebases and all environments. The risk is low in practice (workflow names rarely contain \`;\`), but correctness requires a proper fix with migration tooling.

---

## Context

Part of epic #888 (WAIT_FOR_WEBHOOK). See also \`orkes-io/orkes-conductor#3535\` for the Orkes Enterprise adapter layer that bridges OSS interfaces to the Orkes Postgres/Redis infrastructure.

## Test plan

- [x] All 15 \`conductor-webhook-task\` unit tests pass
- [x] \`conductor-redis-persistence\` compiles cleanly
- [x] \`spotlessApply\` applied with no violations

### DIFF
--- a/redis-persistence/src/main/java/com/netflix/conductor/redis/dao/RedisWebhookTaskDAO.java
+++ b/redis-persistence/src/main/java/com/netflix/conductor/redis/dao/RedisWebhookTaskDAO.java
@@ -13,6 +13,7 @@
 package com.netflix.conductor.redis.dao;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import org.conductoross.conductor.common.webhook.WebhookTaskDAO;
@@ -72,5 +73,30 @@ public class RedisWebhookTaskDAO extends BaseDynoDAO implements WebhookTaskDAO {
     @Override
     public void remove(String hash, String taskId) {
         jedisProxy.srem(nsKey(WEBHOOK_TASKS, hash), taskId);
+    }
+
+    /**
+     * Atomically claims all waiting task IDs for a routing hash using SMEMBERS + DEL.
+     *
+     * <p>DEL is atomic in Redis — it removes the entire key in one round-trip. The race window
+     * between SMEMBERS and DEL is narrow (two sequential commands) vs. the default implementation
+     * (SMEMBERS + N individual SREMs). For true single-command atomicity, replace with a Lua {@code
+     * SMEMBERS+DEL} script via {@code jedisProxy.evalsha()} once a {@code SCRIPT LOAD} hook is
+     * available.
+     *
+     * <p>If the set was already deleted between SMEMBERS and DEL (concurrent inbound event), DEL is
+     * a no-op and the same task IDs may be returned twice. {@link
+     * IncomingWebhookService#completeTask} handles this gracefully — it checks terminal status
+     * before completing.
+     */
+    @Override
+    public List<String> popAll(String hash) {
+        String key = nsKey(WEBHOOK_TASKS, hash);
+        List<String> members = new ArrayList<>(jedisProxy.smembers(key));
+        if (members.isEmpty()) {
+            return Collections.emptyList();
+        }
+        jedisProxy.del(key);
+        return members;
     }
 }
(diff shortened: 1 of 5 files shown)


nthmost-orkes:
Folded into #894 — all fixes from the precision pass and adversarial review are now documented in the #894 description. Closing this draft to keep the review in one place.

PR #973, opened 2026-04-04
Improve SUB_WORKFLOW reliability, recovery, and scalability

rajeshwar-nu:
## Summary

This PR improves `SUB_WORKFLOW` reliability, recovery, and scalability for large fanout and nested fanout workloads.

It addresses a failure mode where parent workflows can persist `SUB_WORKFLOW` tasks but fail to reliably create, reattach, or recover child workflow linkage after transient persistence/queue instability or long synchronous child-start paths.

The new model makes child launch retry-safe with deterministic child workflow ids and moves child start work onto the async system-task path, reducing pressure on the parent workflow decision loop.

## Problem

`SUB_WORKFLOW` launch previously coupled too much work to parent workflow scheduling.

For workflows like:

```text
dynamic fork/join -> SUB_WORKFLOW -> dynamic fork/join -> SUB_WORKFLOW
```

large fanout can create many child workflows while parent workflow lock is held. If Redis, queue, or persistence has a transient failure during that loop, parent can be left with `SUB_WORKFLOW` tasks that are scheduled but not cleanly attached or recoverable.

Main failure risks:

- parent decision loop blocked by synchronous child workflow creation
- lock lease expiry while parent still launching many children
- partial child launch followed by failed parent task update
- retry path lacking stable child identity
- stale index reads incorrectly influencing child existence checks
- parent/child attachment lag causing tests and UI reads to observe intermediate states
- generic postpone timing causing slow revisit for unresolved active `SUB_WORKFLOW` tasks

### Async SUB_WORKFLOW launch

`SUB_WORKFLOW` tasks are queued and executed through async system-task execution instead of launching child workflows inline during parent scheduling.

This reduces heavy child startup work on the parent decision path and makes dynamic fanout with nested sub-workflows less sensitive to parent lock lease duration.

### Deterministic child workflow identity

Child workflow id is generated deterministically from:

```text
parentWorkflowId + parentTaskId + retryCount
```

This gives each parent `SUB_WORKFLOW` task attempt a stable child workflow id. If launch is retried after a transient failure, retry targets the same child id instead of creating duplicate/orphan child workflows.

### Idempotent child start / reattach

Child launch now uses idempotent start semantics:

- if child workflow exists in execution store, reuse and attach it
- if child workflow does not exist in execution store, create it
- if execution-store lookup fails transiently, fail transiently instead of falling back to stale index state

Index-only presence no longer suppresses child creation.

### Faster parent attachment

Parent `SUB_WORKFLOW` task gets `subWorkflowId` attached once child workflow record exists. It does not wait for child workflow inline expansion to finish before persisting linkage.

This shortens parent attachment gap and gives UI/API reads eventual consistency with clearer semantics:

- `subWorkflowId` can appear before child first task is scheduled
- child workflow record exists before id is attached
- child first task may become visible shortly after child evaluation runs

### Recovery behavior

Recovery path now handles scheduled `SUB_WORKFLOW` tasks without child id by retrying deterministic launch. Failed launch also leaves explicit launch error state on task output instead of ambiguous blank scheduled state.

Retry/rerun/restart paths clear stale subworkflow runtime state before re-execution so new attempts use new deterministic child id for new retry count.

### SUB_WORKFLOW revisit timing

Active `SUB_WORKFLOW` tasks now get dedicated postpone handling. `SCHEDULED` and `IN_PROGRESS` `SUB_WORKFLOW` tasks use workflow offset timeout candidate, capped by max postpone timeout.

This avoids applying generic worker-task postpone behavior to orchestration tasks and makes unresolved child linkage/status refresh faster.

### Parent propagation and join recovery

Parent update paths for retry/rerun/restart now push affected parents for evaluation and reopen relevant join tasks when active fork branches exist.

This keeps hierarchical subworkflow retry/rerun/restart flows moving after child state changes.

## Reliability Impact

This PR improves reliability by:

- making child launch retry-safe
- avoiding duplicate/orphan child workflows on retry after partial launch
- removing stale index state from child existence decisions
- allowing scheduled unattached `SUB_WORKFLOW` tasks to retry launch
- clearing stale child ids/output when retrying subworkflow tasks
- reducing time unresolved `SUB_WORKFLOW` tasks wait before revisit

## Scalability Impact

This PR improves scalability by:

- removing bulk child creation from parent scheduling critical path
- reducing lock lease sensitivity for large dynamic fanout
- making nested dynamic fanout workloads less likely to exhaust parent decision time
- making child launch parallelizable by async system-task workers

## Compatibility / Behavior Notes

This PR intentionally changes `SUB_WORKFLOW` launch behavior:

- child launch now runs through async system-task execution
- parent `SUB_WORKFLOW` task can be `SCHEDULED` briefly before async launch starts it
- parent can expose `subWorkflowId` before child first task is scheduled
- deterministic child workflow ids are UUID-formatted but derived from parent workflow/task/retry identity
- `WorkflowExecutor.startWorkflowIdempotent(...)` returns `WorkflowModel`

No new persistence reservation APIs or reservation storage are required in this final approach.

Mixed-version rollout should be treated carefully because launch timing and retry semantics changed.

## Tests

Added/updated coverage for:

- deterministic subworkflow id generation
- idempotent child create-vs-reattach behavior
- execution-store-only child existence checks
- transient execution-store lookup failure propagation
- scheduled `SUB_WORKFLOW` retry without existing `subWorkflowId`
- stale subworkflow id/output cleanup on retry/rerun/restart
- async `SUB_WORKFLOW` system-task execution in integration and e2e tests
- hierarchical subworkflow retry/rerun/restart behavior
- active `SUB_WORKFLOW` postpone behavior
- join recovery after subworkflow retry/rerun/restart

## Alternatives Considered

| Option | Design | Pros | Cons | Decision |
|---|---|---|---|---|
| 1 | Deterministic child workflow id from `parentWorkflowId + parentTaskId + retryCount` | Retry-safe; no extra reservation storage; UUID format preserved; stable child identity across transient failures; no duplicate/orphan child creation for same parent task attempt | Changes child id semantics from random to derived; retry-count handling must be correct; callers must understand child id is stable per parent task attempt | Chosen. Best balance of reliability, simplicity, and storage compatibility. |
| 2 | Generate random UUID up front and persist it into parent `SUB_WORKFLOW` task before child exists | Keeps random UUID semantics; no reservation DAO/keyspace/table; retry can reuse persisted id if parent task update succeeds | UI/API can expose `subWorkflowId` before child workflow exists, causing temporary 404; if parent task update fails before child launch, retry may generate different id; attachment correctness depends on parent persistence happening first | Rejected. Simpler, but creates weaker failure behavior and visible broken-link window. |
| 3 | Generate random UUID inside async system-task worker on each launch attempt | Smallest code change; no reservation storage; child id only appears after worker creates child | Poor retry safety; failure after child create but before parent attach can create duplicate/orphan child workflows on retry; side effects can run more than once | Rejected. Unsafe under exact failure mode this PR targets. |
| 4 | Reserve random child workflow id in reservation storage per parent workflow/task | Retry-safe; keeps random UUID semantics; avoids exposing child id before child exists; can recover same id across parent task retries | Adds new DAO/storage/keyspace/table behavior; needs cleanup across cancel/delete/remove paths; more backend-specific complexity; more operational surface area | Rejected after implementation review. Reliability benefit similar to option 1, but much higher persistence complexity. |

Current PR uses option 1 because it provides stable retry identity without adding new persistence primitives. It also keeps child ids UUID-formatted, so existing storage backends can store them without schema/key changes.


rajeshwar-nu:
Hi @vishesh-orkes, @mp-orkes , @nthmost-orkes , can I please get some thoughts on this 🙏🏻 
Thanks

v1r3n:
@rajeshwar-nu can you address the failing tests otherwise this PR LGTM

rajeshwar-nu:
@v1r3n thanks for the review, tests are fixed now

v1r3n:
I see the sub-workflow ids are pre-fetched.  I think an alternative to it might be to accept workflow Id as part of the start workflow request.  The ids can be generated using IdGenerator attached to the parent workflow and then start the sub-workflow.  This avoids the id reservation logic and keeps things simpler.  This also means as a user you can pass the id as part of the start workflow request.  I would then go ahead and update IdGenerator to add a validate method that ensures that the user supplied id is in a valid format.

rajeshwar-nu:
> I see the sub-workflow ids are pre-fetched. I think an alternative to it might be to accept workflow Id as part of the start workflow request. The ids can be generated using IdGenerator attached to the parent workflow and then start the sub-workflow. This avoids the id reservation logic and keeps things simpler. This also means as a user you can pass the id as part of the start workflow request. I would then go ahead and update IdGenerator to add a validate method that ensures that the user supplied id is in a valid format.

Generating subworkflow_id in parent workflow does not scale well for dynamic fork-join with many forks (around 1k forks). The idea behind the reservation is to keep things async, including generation of subworkflow-id, scheduling and starting the subworkflow. The reservation can quickly be created, which allows huge dyn fork-join fan-outs to complete quickly (preventing the the leased lock from getting timeout). Otherwise, large no. of forks executing sequentially within the single lease of parent cannot complete in time, leaving the workflow vulnerable to get inconsistent updates by a subsequent decider call

v1r3n:
I think we can avoid having the id reservation.  Traditionally, ids are blocked and need to be accounted for, in case of conductor each id is a random UUID, so instead of doing reservation, we can just generate an id on demand and attach to a workflow (we will need to update workflow executor to be able to accept id as part of start request, which should only be available internally for sub-workflow flow and not for anything else).

This avoids the complexity, and if for some reason if the system fails, the next attempt will generate new id, we don't need to do reservation.

How about this approach?

v1r3n:
> > I see the sub-workflow ids are pre-fetched. I think an alternative to it might be to accept workflow Id as part of the start workflow request. The ids can be generated using IdGenerator attached to the parent workflow and then start the sub-workflow. This avoids the id reservation logic and keeps things simpler. This also means as a user you can pass the id as part of the start workflow request. I would then go ahead and update IdGenerator to add a validate method that ensures that the user supplied id is in a valid format.
> 
> Generating subworkflow_id in parent workflow does not scale well for dynamic fork-join with many forks (around 1k forks). The idea behind the reservation is to keep things async, including generation of subworkflow-id, scheduling and starting the subworkflow. The reservation can quickly be created, which allows huge dyn fork-join fan-outs to complete quickly (preventing the the leased lock from getting timeout). Otherwise, large no. of forks executing sequentially within the single lease of parent cannot complete in time, leaving the workflow vulnerable to get inconsistent updates by a subsequent decider call

what about generating them in parallel? we are talking about generating 10K UUIDs.

rajeshwar-nu:
@v1r3n 
So I see this as a choice between one of the 4 options

1. Have a deterministic subworkflowId (unqiue per parentId, parentTaskId) -> this ensure resilience against failures and retries. Tradeoff -> the format of sub-workflow-id needs to change to accomodate deterministic nature of it
2. Generate existing format random uuid using IDGenerator and persist within the SUBWORKFLOW task -> creates a period of UI showing 404 on clicking the subworkflow-id, until the async system task-worker creates the workflow, which can be long for a busy conductor
3. Generate random uuid for subworkflow-id, when handling the async start of subworkflow in asyn system task-worker -> poorest failure resilience (this is something I observed when progressing towards solution 2). Any retries on failure will create new subworkflows, while old subworkflows will become orphaned and likely execute to produce side-effects
4. Reserve a subworkflow-id deterministically per (parentId,parentTaskId) -> ensure failure reslience, does not create duplicate subworkflow for a particular parentSubworkflowTask id on retries

The current PR implements solution 4


| Option | Design | Resilience | UI 404 Window | Complexity | Tradeoff |
|---|---|---|---|---|---|
| 1 | Deterministic ID from `parentWorkflowId + parentTaskId` | High | No, if attached only after child exists | Medium | Changes ID generation semantics; must validate ownership on reuse; retry/rerun/restart need careful tests |
| 2 | Random ID persisted in `SUB_WORKFLOW` task before child exists | High | Yes, can be long on busy async queue | Low/Medium | Simpler than reservation, but frontend/API may expose child ID before child workflow exists |
| 3 | Random ID generated inside async worker per attempt | Low | No | Low | Failure after child create but before parent attach creates duplicate/orphan children with side effects |
| 4 | Current PR: random ID stored in reservation store | High | No | High | Safe, but adds reservation DAO/keyspace/table/migration/cleanup across all persistence backends |

rajeshwar-nu:
@v1r3n I changed the approach here. Instead of using a reservation store, the flow is now:

1. When a SUB_WORKFLOW task is scheduled, Conductor persists it as SCHEDULED.
2. The SUB_WORKFLOW task is pushed to the system task queue for async processing.
3. The async system task worker picks up the task and starts the child workflow through an idempotent start path.
4. The child workflow id is deterministic for the parent task attempt: parentWorkflowId + parentTaskId + retryCount. It is still formatted as a UUID, so storage/backend compatibility is preserved.
5. Because the child id is deterministic, retries/restarts for the same parent task attempt target the same child workflow id. If the child already exists, the idempotent starter reuses it; if it does not exist, it creates it.
6. The parent SUB_WORKFLOW task is only attached to the child workflow id after the child workflow record exists, avoiding a long UI/API window where the parent points to a non-existent child.

v1r3n (approved):

PR #996, opened 2026-04-08
feat: wildcard search and sub-workflow filtering

v1r3n:
## Summary

- **Wildcard search**: `workflowType=order*` now works across all backends (Postgres, SQLite, ES7, ES8, OpenSearch). Postgres/SQLite translate `*` to SQL `LIKE %`. ES7/OpenSearch already support Lucene wildcards natively. ES8 now uses `wildcard` query instead of `term` when `*` is detected.
- **Sub-workflow filtering**: `parentWorkflowId` is now an indexed, queryable field. Top-level workflows store `""`, sub-workflows store the parent's ID. Query with `parentWorkflowId=""` to exclude sub-workflows.
- **UI toggle**: "Exclude sub-workflows" switch added to the workflow search BasicSearch filter bar.

### Backend changes
- `WorkflowSummary`: added `parentWorkflowId` field (`@ProtoField(id=22)`) + proto/gRPC mapping
- Flyway migrations: Postgres V14, SQLite V2 — adds `parent_workflow_id` column + index
- ES7/ES8/OpenSearch: added `parentWorkflowId` keyword mapping
- `PostgresIndexQueryBuilder` / `SqliteIndexQueryBuilder`: wildcard `*` in `=` operator emits `LIKE` with `%` substitution; `parent_workflow_id` added to valid fields
- `PostgresIndexDAO` / `SqliteIndexDAO`: persist `parentWorkflowId` in INSERT/UPSERT
- ES8 `NameValue`: detect `*` in EQUALS values and use `wildcard` query instead of `term`

### UI changes
- `WorkflowSearch.tsx`: new `excludeSubWorkflows` URL query state
- `BasicSearch.tsx`: toggle switch + `parentWorkflowId=""` clause in `buildQuery()`

### DIFF
--- a/sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
+++ b/sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
@@ -97,7 +97,9 @@ public class SqliteIndexQueryBuilder {
             } else {
                 if (attribute.endsWith("_time")) {
                     return attribute + " " + operator + " datetime(?)";
-                } else if (operator.equals("=") && values.size() == 1 && values.get(0).contains("*")) {
+                } else if (operator.equals("=")
+                        && values.size() == 1
+                        && values.get(0).contains("*")) {
                     return "lower(" + attribute + ") LIKE lower(?)";
                 } else {
                     return attribute + " " + operator + " ?";
--- a/sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
+++ b/sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
@@ -48,7 +48,8 @@ public class SqliteIndexQueryBuilder {
         "task_type",
         "task_def_name",
         "update_time",
-        "json_data"
+        "json_data",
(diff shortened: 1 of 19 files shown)


sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
```diff
@@ -116,7 +121,11 @@ public void addParameter(Query q) throws SQLException {
                     q.addParameter(value);
                 }
             } else {
-                q.addParameter(values.get(0));
+                String val = values.get(0);
+                if (val.contains("*")) {
+                    val = val.replace("*", "%");
```
manan164, line 126: Should we put all special chars here, something like val = val.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_").replace("*", "%");? This will take care of cases like, workflowType=order_proc* would match orderXprocV1 since _ can match single char in sql.

manan164 (approved):

sqlite-persistence/src/main/java/com/netflix/conductor/sqlite/util/SqliteIndexQueryBuilder.java
v1r3n: I think we can do a separate PR for that?

PR #997, opened 2026-04-08
Fix standalone and sentinel Redis command retries

### LINKED ISSUE #938: [BUG] Redis Sentinel path ignores conductor.redis.maxRetryAttempts, so transient failover errors are not retried
**Describe the bug**

When Conductor is configured with `redis_sentinel`, the shared Redis property `conductor.redis.maxRetryAttempts` appears to have no effect.

This is confusing operationally, because the property suggests that transient Redis failures will be retried, but the Sentinel path still surfaces failover-time exceptions directly to the application during master changes.

In practice, during a Sentinel failover window, transient errors such as connection failures and sometimes `READONLY` failures can bubble up from Redis-backed DAO/queue operations even when `conductor.redis.maxRetryAttempts` is set.

**Details**

Conductor version: current `main`
Persistence implementation: `redis_sentinel`
Queue implementation: `redis_sentinel`
Lock: `redis`
Workflow definition: N/A
Task definition: N/A
Event handler definition: N/A

Relevant configuration example:

```properties
conductor.db.type=redis_sentinel
conductor.queue.type=redis_sentinel
conductor.redis.hosts=<sentinel-1>:26379:<rack>;<sentinel-2>:26379:<rack>;<sentinel-3>:26379:<rack>
conductor.redis.clusterName=mymaster
conductor.redis.maxRetryAttempts=3
```

Relevant code paths on `main`:

- `redis-persistence/src/main/java/com/netflix/conductor/redis/config/RedisProperties.java`
- `redis-persistence/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java`

`RedisProperties` exposes `maxRetryAttempts`, but `RedisSentinelConfiguration` creates a `JedisSentinelPool` directly and does not appear to use that setting or an equivalent retry policy.

**To Reproduce**

1. Configure Conductor to use `redis_sentinel`.
2. Set `conductor.redis.maxRetryAttempts` to a non-zero value, for example `3`.
3. Start a workload that performs normal Redis-backed queue / execution / metadata operations.
4. Trigger a Sentinel failover by stopping the current Redis master or forcing Sentinel to promote a replica.
5. Observe that transient Redis failures are still surfaced during the failover window instead of being retried by the Sentinel client path.

**Expected behavior**

If a shared Redis retry property is exposed, it should either:

- apply to the `redis_sentinel` path as well, with clear bounded behavior for transient failover errors, or
- be documented/deprecated as not applicable to Sentinel so operators are not given a false sense of resilience.

At minimum, the current behavior should be made explicit in docs/config naming.

**Additional context**

This looks like a gap between configuration surface area and actual behavior:

- `conductor.redis.maxRetryAttempts` exists as a general Redis property
- the Sentinel implementation uses `JedisSentinelPool`.
- the Sentinel path does not appear to consume `maxRetryAttempts` or any equivalent retry/backoff settings

A possible fix would be one of:

1. Add explicit Sentinel retry handling for transient failover exceptions with bounded attempts/backoff.
2. Introduce Sentinel-specific retry properties instead of relying on a shared property name.
3. Deprecate or document `conductor.redis.maxRetryAttempts` as not applying to `redis_sentinel` if that is the intended behavior.

One caution: generic retries around Redis writes should be limited to clearly transient failover/network exceptions and kept bounded, since not every Redis operation is safely retryable.

uri-simspace:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
This PR adds bounded retries to the shared `JedisCommands` bean used by Redis standalone and Redis Sentinel.

Root cause:
`conductor.redis.maxRetryAttempts` was defined, but it was not applied to the non-cluster `UnifiedJedisCommands` path. During Redis Sentinel failover, transient errors such as connection resets, `READONLY`, `MASTERDOWN`, `TRYAGAIN`, and `LOADING` could bubble straight back to Conductor instead of being retried for a short bounded window.

What changed:
- add a `RetryingJedisCommands` decorator driven by `conductor.redis.maxRetryAttempts`
- wrap the shared `JedisCommands` bean for `redis_standalone` and `redis_sentinel`
- retry only transient failover errors
- add unit coverage for retry classification and retries on command paths like `evalsha`
- update configuration tests for the new wrapped `JedisCommands` bean

Impact:
- improves Redis DAO resiliency during standalone and Sentinel failover windows
- improves queue-related Redis command paths that use the shared `JedisCommands` seam
- leaves Redis cluster behavior unchanged in this change set

Validation:
- `./gradlew spotlessApply --no-daemon`
- `./gradlew :conductor-redis-configuration:test :conductor-redis-persistence:test --no-daemon`

Issue #
Fixes #938

Alternatives considered
----
Retrying at the DAO layer was rejected because many DAO methods are multi-step and could duplicate side effects when retried wholesale. Restricting the fix to queue-specific code was also not enough, because the same missing retry behavior affects the broader OSS Redis command path outside the queue implementation.

### DIFF
--- a/redis-configuration/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java
+++ b/redis-configuration/src/main/java/com/netflix/conductor/redis/config/RedisSentinelConfiguration.java
@@ -27,6 +27,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 import com.netflix.conductor.redis.jedis.JedisCommands;
+import com.netflix.conductor.redis.jedis.RetryingJedisCommands;
 import com.netflix.conductor.redis.jedis.UnifiedJedisCommands;
 import com.netflix.dyno.connectionpool.Host;
 
@@ -46,8 +47,9 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 public class RedisSentinelConfiguration extends RedisConfiguration {
 
     @Bean
-    public JedisCommands getJedisCommands(UnifiedJedis unifiedJedis) {
-        return new UnifiedJedisCommands(unifiedJedis);
+    public JedisCommands getJedisCommands(
+            UnifiedJedis unifiedJedis, RedisProperties redisProperties) {
+        return RetryingJedisCommands.wrap(new UnifiedJedisCommands(unifiedJedis), redisProperties);
     }
 
     @Override
--- a/redis-configuration/src/main/java/com/netflix/conductor/redis/config/RedisStandaloneConfiguration.java
+++ b/redis-configuration/src/main/java/com/netflix/conductor/redis/config/RedisStandaloneConfiguration.java
@@ -21,6 +21,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.stereotype.Component;
 
 import com.netflix.conductor.redis.jedis.JedisCommands;
+import com.netflix.conductor.redis.jedis.RetryingJedisCommands;
 import com.netflix.conductor.redis.jedis.UnifiedJedisCommands;
 import com.netflix.dyno.connectionpool.Host;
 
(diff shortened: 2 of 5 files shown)


uri-simspace:
One scope note for reviewers: this PR improves the shared OSS Redis command path by adding bounded retries for transient failover errors in the standalone and Sentinel `JedisCommands` wiring. That gives us meaningful coverage in Conductor itself, including DAO operations and queue-related Redis calls that go through the shared client seam. There is still likely follow-up work needed in `orkes-conductor-queues` for full queue failover behavior, especially around queue-specific operations such as polling/script execution and any internal retry or exception-handling logic there. I’m calling that out explicitly because it is related to the same failure surface, but it is outside the scope of this repository and this PR.

v1r3n (approved):

PR #1022, opened 2026-04-16
feat(file-storage) Initial implementation for file storage feature

### LINKED ISSUE #1024: [FEATURE]: Support configurable object storage targets for AI-generated image/video outputs
## Describe the Feature Request
AI image and video generation tasks currently persist generated media to local disk by default. This creates friction for deployments where application instances are ephemeral, stateless, or horizontally scaled, and where generated media needs to be stored in shared object storage instead of node-local filesystems.

It would be helpful for Conductor to support writing AI-generated image/video outputs to configurable object storage locations, including S3-compatible backends, rather than relying only on local disk storage.

## Describe Preferred Solution
Allow AI media generation tasks to store generated outputs in a configurable remote object storage target, such as AWS S3 or other S3-compatible storage systems.

From a user perspective, the desired behavior is:
- generated image/video outputs do not have to be persisted on local disk
- object storage can be used as the primary destination for generated media
- the feature works for common self-hosted and cloud deployments where shared object storage is the standard persistence layer

### Alternatives

Current alternatives are not ideal:
- writing media to local disk and handling a separate upload step outside Conductor
- using shared network filesystems instead of object storage
- building custom post-processing around AI tasks to move generated files after task completion

### Additional context

_No response_

kowser-orkes:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

# Summary

Adds a file storage subsystem to Conductor OSS that lets workers pass large binary files between tasks via presigned
URLs, without routing file bytes through the Conductor server.

- **REST API** (`/api/files`) — create file, get upload/download URLs, confirm upload, initiate/complete multipart upload
- **Workflow-scoped access** — download requests are gated by workflow family (self + ancestors + descendants); a
  configurable `defaultWorkflowId` bypass for shared reference files
- **4 storage backends** — S3, Azure Blob, GCS, Local 
  - feature-flagged, disabled by default via `conductor.file-storage.enabled`
- **5 DAO implementations** — Postgres (`V15`), MySQL (`V9`), SQLite (`V3`), Redis, Cassandra
- **File handle IDs** use `conductor://file/<fileId>` scheme
  - conversion between bare `fileId` (URL path params) and `fileHandleId` (JSON bodies) is handled by `FileIdToFileHandleIdConverter`
- **Multipart upload** support: S3 uses per-part presigned URLs; GCS/Azure use a single resumable URL from initiate

# Key design decisions

- File bytes never pass through the Conductor server — all transfers go directly between the SDK client and the storage backend via presigned URLs
- Files are strictly owned by their creator workflow — download access requires the caller's workflow to be in the file's workflow family (self + ancestors + descendants), resolved by `WorkflowFamilyResolver`
- `storagePath` is server-internal and never exposed in any API response

# Test plan

- [ ] `FileStorageServiceImplTest` — unit tests for all service methods incl. access control
- [ ] `WorkflowFamilyResolverTest` — ancestor/descendant resolution
- [ ] `FileModelConverterTest` — DTO ↔ model conversion
- [ ] `FileResourceTest` — controller layer
- [ ] `LocalFileStorageTest` — local backend
- [ ] Integration tests for S3 (LocalStack), Azure Blob (Azurite), GCS (fake-gcs-server)
- [ ] E2E test suite in `e2e/` against MinIO

# References

- File Storage Design Doc: `docs/file-storage/design.md`

Alternatives considered
----


rest/src/main/java/org/conductoross/conductor/controllers/FileResource.java
```diff
@@ -0,0 +1,90 @@
...
+        return fileStorageService.getUploadUrl(fileId);
+    }
+
+    @PostMapping("/{fileId}/upload-complete")
+    @Operation(summary = "Confirm file upload completion")
+    public FileUploadCompleteResponse confirmUpload(@PathVariable String fileId) {
+        return fileStorageService.confirmUpload(fileId);
+    }
+
+    @GetMapping("/{fileId}/download-url")
```
v1r3n, line 60: this should be at the workflow id level as `/{workflowId}/{fileId}/download-url`

rest/src/main/java/org/conductoross/conductor/controllers/FileResource.java
```diff
@@ -0,0 +1,96 @@
...
+    @Operation(summary = "Confirm file upload completion")
+    public FileUploadCompleteResponse confirmUpload(@PathVariable("fileId") String fileId) {
+        return fileStorageService.confirmUpload(fileId);
+    }
+
+    @GetMapping("/{workflowId}/{fileId}/download-url")
+    @Operation(summary = "Get presigned download URL")
+    public FileDownloadUrlResponse getDownloadUrl(
+            @PathVariable("workflowId") String workflowId, @PathVariable("fileId") String fileId) {
+        return fileStorageService.getDownloadUrl(fileId);
```
v1r3n: the implementation does not use workflowId.

v1r3n (approved):

PR #1047, opened 2026-04-27
fix: prevent rerun from failing parent workflows via sweeper race

nthmost-orkes:
## Summary

**Root cause**: A race condition in `WorkflowExecutorOps.rerunWF()` caused parent workflows to be permanently failed after a child workflow was rerun.

`updateAndPushParents()` was called before `executionDAOFacade.updateWorkflow()`, meaning the parent was pushed to the DECIDER_QUEUE while the child was **still FAILED in the DAO**. The background sweeper would immediately pick up the parent, call `SubWorkflow.execute()`, read the child as FAILED (terminal), set the parent's SUB_WORKFLOW task to FAILED, and cascade-fail the parent — permanently, with no recovery.

**Fix** (`WorkflowExecutorOps.java`): move `updateAndPushParents()` to *after* `executionDAOFacade.updateWorkflow()` in both branches of `rerunWF()`. The child is now always visible as RUNNING in the DAO before the parent is pushed for sweeper evaluation.

**Why it surfaced now**: the ordering bug has been present since `6e74001ec` (Sep 2021). It became reliably flaky only after `99ecc5f12` (Dec 2025) replaced the old polling-based sweeper with a new `WorkflowSweeper` that runs dedicated executor threads in a tight loop — combined with `expediteLazyWorkflowEvaluation()` pushing the parent with delay=0, the sweeper now wins the race often enough to break CI consistently.

**Additional hardening** (test files):
- `SimpleWorkflowSpec`: bumped `await().atMost(5s → 30s)` for queue-size stabilization after `decide()`.
- `SubWorkflowRerunSpec`: added `await().atMost(10s)` before parent-workflow status checks, consistent with the pattern in `HierarchicalForkJoinSubworkflowRestartSpec`.

## Test plan

- [ ] CI `test-harness` passes on both parallel runs
- [ ] `build` (spotless) passes
- [ ] No logic changes to workflow rerun behavior — only the ordering of DAO write vs. parent notification


nthmost-orkes:
Closing as superseded. The production change here is a no-op: `updateAndPushParents` already sets `subworkflowChanged=true`, sets the parent SUB_WORKFLOW task to IN_PROGRESS, and pushes every parent up the chain to the decider queue (WorkflowExecutorOps lines 315-317), so the extra `updateAndPushParents` call adds nothing. The flaky rerun/restart specs this targeted were de-flaked by #1206, which took the correct test-relaxation approach for the same sweeper race. Nothing actionable remains here.

PR #1065, opened 2026-05-04
fix: respect Spring property hierarchy for external config

### LINKED ISSUE #547: Ability for conductor to pickup config values as environment variables

Hello, I'm trying to incorporate the standalone conductor into my K8s deployment. Built the Docker image following: https://conductor-oss.github.io/conductor/devguide/running/docker.html#standalone-server-image.

I figured out that i need to pass the Redis host name at build time, since it is picked up from the `config.properties` file. It will be great if we can pass the values in `config.properties` at runtime as environment variables.

### Alternatives

I actually made a small change to the `docker/server/bin/startup.sh` file and added the ability to pick up environment variables like this:

before: `java ${JAVA_OPTS} -jar -DCONDUCTOR_CONFIG_FILE=$config_file conductor-server.jar 2>&1 | tee -a /app/logs/server.log`

after: `java ${JAVA_OPTS} -jar -DCONDUCTOR_CONFIG_FILE=$config_file ${CONDUCTOR_OPTS} conductor-server.jar 2>&1 | tee -a /app/logs/server.log`

`CONDUCTOR_OPTS` can now be set by k8s or Docker at runtime like this:

Docker example: `docker run -it --rm -e "CONDUCTOR_OPTS=-DCONDUCTOR.REDIS.HOSTS=172.17.0.1:6379:us-east-1c" conductor:server`.

Let me know if this is a good way to do it. 

### Additional context

_No response_

vovaparker:
## Summary

Register `CONDUCTOR_CONFIG_FILE` as a Spring [config-data location][spring-docs]
instead of pushing every entry into JVM system properties. This restores the
documented Spring Boot property hierarchy: env vars and JVM args now correctly
override values from the config file, while the file still overrides Spring's
built-in defaults.

The previous behavior blocked the standard PaaS deployment pattern (Railway,
Heroku, K8s, ECS) where credentials and connection strings are injected via
env vars. Operators currently work around it by passing `-D...` JVM flags
via `JAVA_OPTS`.

The same fix is applied to `server-lite` (`CONDUCTOR_LITE_CONFIG_FILE`),
which had identical code.

Resolves #547.

## Fix

```java
System.setProperty("spring.config.additional-location", "optional:file:" + configFile);
```

Spring then loads the file as part of its standard property resolution. The
resulting precedence (low → high):

1. Default properties
2. `@PropertySource` annotations
3. Config data (`application.properties`, etc.) ← `CONDUCTOR_CONFIG_FILE` lands here
4. OS environment variables
5. Java System properties
6. Command-line arguments

The `optional:` prefix preserves existing behavior of tolerating a missing
file rather than failing startup.

## Verification

A test config file defines `spring.datasource.url=jdbc:postgresql://from-file:5432/postgres`.
An optional env var sets `SPRING_DATASOURCE_URL=jdbc:postgresql://from-env:5432/postgres`.
Conductor's startup connection attempt fails with
`java.net.UnknownHostException: <host>` — the `<host>` proves which
`spring.datasource.url` value Spring resolved.

| # | Branch  | Env var    | File value  | Resolved host | Outcome                                    |
|---|---------|------------|-------------|---------------|--------------------------------------------|
| 1 | `main`  | `from-env` | `from-file` | `from-file`   | Reproduces #547 — env var ignored          |
| 2 | this PR | `from-env` | `from-file` | `from-env`    | Env var overrides file ✅                   |
| 3 | this PR | (unset)    | `from-file` | `from-file`   | File still overrides classpath defaults ✅ |

## Backward compatibility

**Env-var precedence change.** Users currently relying on `CONDUCTOR_CONFIG_FILE`
to override env-var-set values will see those env vars take effect after this change. 
Migration: set the value via env var (which now works as expected)
or via `-D` JVM args (which still take precedence over both file and env).
The previous behavior contradicted the documented Spring Boot property
hierarchy; this change brings Conductor in line with it.

**Workflow placeholder resolution.** Workflow `${VAR}` placeholders are
resolved by `EnvUtils`, which reads `System.getProperty` and `System.getenv`.
Before this change, every key in `CONDUCTOR_CONFIG_FILE` was reachable that
way — including sensitive keys like `spring.datasource.password`. 
After this change, file-defined values are no longer reachable; 
placeholders for them will be left unsubstituted in task inputs.

[spring-docs]: https://docs.spring.io/spring-boot/reference/features/external-config.html

### DIFF
--- a/server-lite/src/main/java/org/conductoross/conductor/Conductor.java
+++ b/server-lite/src/main/java/org/conductoross/conductor/Conductor.java
@@ -12,9 +12,6 @@
  */
 package org.conductoross.conductor;
 
-import java.io.IOException;
-import java.util.Properties;
-
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -23,7 +20,6 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.autoconfigure.mongo.MongoAutoConfiguration;
 import org.springframework.context.annotation.ComponentScan;
-import org.springframework.core.io.FileSystemResource;
 
 // Prevents from the datasource beans to be loaded, AS they are needed only for specific databases.
 // In case that SQL database is selected this class will be imported back in the appropriate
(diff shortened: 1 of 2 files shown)


nthmost-orkes:
Thank you for the detailed write-up and the verification table — it's clear you put real thought into this, and issue #547 is a genuine pain point for PaaS deployments.

Unfortunately we can't merge this as-is. The backward-compatibility section you flagged is a production-breaking change for a significant portion of our users: Conductor's `EnvUtils` resolves `${VAR}` placeholders in task inputs by calling `System.getProperty` and `System.getenv`. Under the current behavior, every key in `CONDUCTOR_CONFIG_FILE` is available that way. After this change, file-defined values are no longer in the JVM system properties, so any workflow using `${spring.datasource.password}` or similar config-file keys as task input placeholders will silently receive the literal `${...}` string instead of the value. That's a silent data-corruption class of regression.

The right path forward is likely a two-phase approach: fix the Spring property hierarchy (your change) and simultaneously update `EnvUtils` to also source values from Spring's `Environment` bean, so placeholder resolution isn't broken. If you'd like to take that on we'd welcome a revised PR.

Closing for now — thanks again for the investigation.

PR #1068, opened 2026-05-04
resolve ${...} expressions in subWorkflowParam.workflowDefinition at …

v1r3n:
…runtime

Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
Problem

  SubWorkflowTaskMapper added workflowDefinition to the params map after calling getTaskInputV2, so ${ref.output.field} expressions were never resolved. The value ended up as a literal string in workflowInput rather
  than as the resolved Map in inputData["subWorkflowDefinition"] where SubWorkflow.start() reads it.

Issue #

Alternatives considered
----


bradyyie (approved):
LGTM !

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
```diff
@@ -1803,6 +1802,7 @@ private boolean rerunWF(
```
nthmost-orkes, line 1802: Can you add a sentence to the PR description explaining this ordering change? Moving `updateAndPushParents()` to after `decide()` changes observable behavior for parent workflows — it would help reviewers (and future git-blame readers) understand what was wrong with the original order and confirm there's no race condition introduced here.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
```diff
@@ -47,19 +49,52 @@ public SubWorkflow(ObjectMapper objectMapper) {
...
-        Integer resolvedVersion = version == 0 ? null : version;
+
+        // Null-safe version read: version may be absent when an inline workflowDefinition is
+        // supplied (the mapper skips the MetadataDAO lookup in that case).
+        // Use Number.intValue() — the same JSON round-trip issue that affects priority (Integer
+        // stored, Double retrieved) applies here: parseInt("2.0") would throw and silently fall
+        // back to null (= latest), causing the wrong sub-workflow version to be executed.
+        Integer resolvedVersion = null;
+        Object versionObj = input.get("subWorkflowVersion");
+        if (versionObj instanceof Number) {
```
nthmost-orkes, line 75: Question: is priority propagation working end-to-end here? The mapper sets `subWorkflowTask.setWorkflowPriority()` (a task-level attribute) but this code reads from `input.get("priority")` (task inputData). Those are different fields — the mapper never calls `addInput("priority", ...)`, so `priorityObj` will always be null unless the user explicitly passes `priority` in `inputParameters`. Should the mapper also call `subWorkflowTask.addInput("priority", priority)` so this path actually fires? Or should `start()` read `task.getWorkflowPriority()` directly?

nthmost-orkes (commented):
Things i'd like -- mostly documentation for clear guidance of humans and machines:

**1. `rerunWF` ordering change needs a description**
The two moved `updateAndPushParents()` calls in `WorkflowExecutorOps` are a behavioral change but the PR description doesn't mention them. Can you add a sentence or two explaining what was wrong with the original order and what the correct semantics are? (See inline comment.)

**2. Parameter table in docs is incomplete**
The updated `sub-workflow-task.md` doesn't list `idempotencyKey`, `idempotencyStrategy`, or `priority` as valid `subWorkflowParam` fields, but the code now propagates all three. Please add rows for them.

**3. `_systemMetadata.dynamic` deserves at least a code comment**
The flag injected into `wfInput` is invisible to users and undocumented. Even a short comment noting the purpose and that it's Conductor-internal would help future engineers (and LLM assistants) understand the contract.

**4. Version null comment suggestion**
See inline comment in `SubWorkflow.java` — one line would make the null-as-latest semantics explicit at the call site.

**5. Question about priority propagation**
See inline comment in `SubWorkflow.java`.

nthmost-orkes:
**Suggested one-liner for `SubWorkflow.start()`** (around the `startWorkflowInput.setVersion(resolvedVersion)` call):

```java
// null = use latest version; set explicitly when resolvedVersion > 0
startWorkflowInput.setVersion(resolvedVersion);
```

Makes the null-as-latest semantic visible at the call site without requiring readers to trace back through the version-resolution logic above.

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
nthmost-orkes: (Or... let me know how it *doesn't* change observable behavior?)

core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
v1r3n: Yes this is coming from Orkes where we already use.

v1r3n:
> **Suggested one-liner for `SubWorkflow.start()`** (around the `startWorkflowInput.setVersion(resolvedVersion)` call):
> 
> ```java
> // null = use latest version; set explicitly when resolvedVersion > 0
> startWorkflowInput.setVersion(resolvedVersion);
> ```
> 
> Makes the null-as-latest semantic visible at the call site without requiring readers to trace back through the version-resolution logic above.

Yes, this is the current behavior with Conductor and sub-workflows in OSS enforced versions.  We don't do that in Orkes.  so making them both at 1:1 now with this change.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java
manan164: I have same concern as @nthmost-orkes , I am writing tests for this.

manan164: it is working fine.

manan164 (approved):

core/src/main/java/com/netflix/conductor/core/execution/mapper/SubWorkflowTaskMapper.java
```diff
@@ -82,13 +87,14 @@ public List<TaskModel> getMappedTasks(TaskMapperContext taskMapperContext) {
         subWorkflowTask.addInput("subWorkflowVersion", subWorkflowVersion);
         subWorkflowTask.addInput("subWorkflowTaskToDomain", subWorkflowTaskToDomain);
         subWorkflowTask.addInput("subWorkflowDefinition", subWorkflowDefinition);
+        subWorkflowTask.addInput("idempotencyKey", resolvedParams.get("idempotencyKey"));
```
mp-orkes, line 90: OSS does not enforce `idempotencyKey` or `idempotencyStrategy`. Is this intentional?

v1r3n: Yes so we can start doing that.

PR #1079, opened 2026-05-11
Fix WorkflowSweeper self-deadlock with non-reentrant locks (#1058)

### LINKED ISSUE #1058: WorkflowSweeper self-deadlocks with non-reentrant locks (Postgres), permanently stalling workflows after SUB_WORKFLOW complete
## Description

When `conductor.workflow-execution-lock.type` is set to `postgres` (or any non-reentrant lock
implementation), workflows containing a `SUB_WORKFLOW` task stall permanently after the
sub-workflow completes. The parent workflow remains in `RUNNING` status indefinitely, and the
sweeper logs repeat every ~30 seconds:

```
INFO WorkflowSweeper : Running sweeper for workflow <id>
INFO WorkflowSweeper : can't get a lock on <id>, will try after 30 seconds
```

### Root Cause

`WorkflowSweeper.sweep()` acquires the execution lock at line 162, then calls
`workflowExecutor.decide(workflowId)` which calls `executionLockService.acquireLock()` **again**
internally (`WorkflowExecutorOps.decide()`, line 1089). With a non-reentrant lock (Postgres
advisory locks are non-reentrant by design), the second acquisition always returns `false`.
`decide()` then returns `null`, which `sweep()` interprets as external lock contention and
re-queues the workflow with a backoff delay. This repeats forever.

```java
// WorkflowSweeper.sweep() — acquires lock here:
if (!executionLockService.acquireLock(workflowId)) { ... }

// Then calls decide() which ALSO calls acquireLock() internally:
workflow = workflowExecutor.decide(workflowId);  // always returns null with non-reentrant lock

// null is misread as contention -> re-queue with backoff -> infinite loop
if (workflow == null) {
    queueDAO.push(DECIDER_QUEUE, workflowId, 0, backoffSeconds);
    return;
}
```

The `SUB_WORKFLOW` task is the most common trigger because its completion does not re-enqueue the
parent workflow directly — it relies on the next sweep cycle to detect the completed sub-workflow
and advance the parent. That sweep cycle then hits this deadlock.

## Steps to Reproduce

1. Configure Conductor with `conductor.workflow-execution-lock.type: postgres` and
   `conductor.app.workflow-execution-lock-enabled: true`
2. Define a workflow containing a `SUB_WORKFLOW` task followed by any subsequent task
3. Start the workflow and let the sub-workflow complete
4. Observe the parent workflow remains in `RUNNING` indefinitely

The following unit test reproduces the bug and can be added to `WorkflowSweeperTest`:

```java
/**
 * Reproduction for: sweep() acquires the execution lock, then calls decide(workflowId) which
 * also calls acquireLock() internally. With a non-reentrant lock (e.g. Postgres advisory lock)
 * the second acquisition always fails, decide() returns null, and the sweeper re-queues the
 * workflow with a backoff delay — leaving it permanently stuck even though the sub-workflow has
 * already completed.
 *
 * <p>Expected (correct) behaviour: decide() is called and the workflow advances.
 * Actual (buggy) behaviour: decide() returns null because the sweeper already holds the lock;
 * the workflow is pushed back to the decider queue with a 30-second delay indefinitely.
 */
@Test
public void sweepDoesNotStallWhenSubWorkflowCompletedAndLockIsNonReentrant() {
    TaskModel subWorkflowTask =
            newTask(
                    "sub-workflow-task",
                    TaskType.TASK_TYPE_SUB_WORKFLOW,
                    TaskModel.Status.IN_PROGRESS);
    subWorkflowTask.setSubWorkflowId("sub-workflow-id");
    WorkflowModel workflow = newWorkflow(List.of(subWorkflowTask));

    WorkflowModel subWorkflow = new WorkflowModel();
    subWorkflow.setStatus(WorkflowModel.Status.COMPLETED);
    subWorkflow.setOutput(Map.of("result", "ok"));

    // The sweeper acquires the lock successfully on the first call.
    // The second call simulates a non-reentrant lock: decide(workflowId) calls acquireLock()
    // internally and cannot acquire a lock that the sweeper already holds.
    when(executionLockService.acquireLock(WORKFLOW_ID))
            .thenReturn(true)   // sweeper's own pre-acquire
            .thenReturn(false); // decide()'s internal acquire — non-reentrant lock fails

    when(workflowExecutor.getWorkflow(WORKFLOW_ID, true)).thenReturn(workflow);
    // decide(workflowId) returns null when it cannot acquire the lock
    when(workflowExecutor.decide(WORKFLOW_ID)).thenReturn(null);

    workflowSweeper.sweep(WORKFLOW_ID);

    // Bug: the workflow is pushed back to the decider queue with a backoff delay,
    // causing it to stall permanently even though the sub-workflow is already done.
    verify(queueDAO, never()).push(anyString(), anyString(), anyInt(), anyLong());
}
```

This test **fails** on the current `main` branch, confirming the bug.

## Suggested Fix

Remove the redundant pre-acquisition of the lock from `WorkflowSweeper.sweep()`. The
`decide(workflowId)` method already handles its own locking — the sweeper's pre-acquire is not
only redundant but actively causes the deadlock with non-reentrant lock backends.

```java
// WorkflowSweeper.sweep() — REMOVE these lines:
    log.error("Couldn't acquire lock to sweep workflow {}", workflowId);
    return;
}
// ...
} finally {
    executionLockService.releaseLock(workflowId);  // also remove
}
```

With this change, `sweep()` calls `decide(workflowId)` directly, which acquires and releases the
lock itself. The `workflow == null` path remains correct for handling genuine external contention
(another thread holding the lock), since `decide()` still returns `null` in that case.

The `executionLockService` field and constructor parameter in `WorkflowSweeper` can then be
removed entirely as they are no longer used.

## Environment

- Conductor version: `3.23.0` (also reproduced on current `main`)
- Lock type: `postgres` (`conductor.workflow-execution-lock.type: postgres`)
- Trigger: any workflow with a `SUB_WORKFLOW` task

## Workaround

Disable the execution lock for single-node deployments:

```yaml
conductor:
  app:
    workflow-execution-lock-enabled: false
  workflow-execution-lock:
    type: noop_lock
```

nthmost-orkes:
## Summary

- Remove the redundant execution-lock pre-acquire from `WorkflowSweeper.sweep()`. `WorkflowExecutorOps.decide(String)` already acquires/releases the same lock internally, so the sweeper's pre-acquire is both redundant and actively harmful with non-reentrant lock backends.
- Drop the now-unused `ExecutionLockService` field and constructor parameter.
- Update existing `WorkflowSweeperTest` cases to stop mocking/verifying the removed lock calls, and add a regression test for the SUB_WORKFLOW deadlock scenario.

Fixes conductor-oss/conductor#1058.

## Root cause

`WorkflowSweeper.sweep()` did this:

```java
if (!executionLockService.acquireLock(workflowId)) { return; }   // sweeper's own pre-acquire
try {
    workflow = workflowExecutor.decide(workflowId);              // calls acquireLock() AGAIN internally
    if (workflow == null) {
        // misread as external contention -> backoff push
        queueDAO.push(DECIDER_QUEUE, workflowId, 0, backoffSeconds);
        return;
    }
    ...
} finally {
    executionLockService.releaseLock(workflowId);
}
```

Postgres advisory locks are non-reentrant by design. The inner `acquireLock` inside `decide()` therefore always returned `false`, `decide()` returned `null`, and the sweeper pushed the workflow back to the decider queue with a backoff delay — forever.

`SUB_WORKFLOW` is the common trigger because the child's completion does not re-enqueue the parent directly; the parent relies on the next sweep cycle to detect completion and advance — and that sweep cycle is exactly where the deadlock fires.

## User impact

First-time users running a Conductor stack on Postgres with the lock enabled — including anyone following a Postgres-backed quickstart — would see workflows with a `SUB_WORKFLOW` task get stuck in `RUNNING` indefinitely once the sub-workflow completed, with the sweeper logging `can't get a lock on <id>, will try after 30 seconds` on a loop. After this fix, those workflows advance to terminal in the next sweep cycle as expected.

## Credit

Diagnosis, failing unit test, and a working docker-compose reproduction by @NicolasBissig in #1058 and https://github.com/NicolasBissig/conductor-oss-lock-bug-repro — thank you.

## Test plan

- [x] `./gradlew :conductor-core:compileJava :conductor-core:compileTestJava` — clean
- [x] `./gradlew :conductor-core:test --tests org.conductoross.conductor.core.execution.WorkflowSweeperTest` — all 5 tests pass (4 existing + 1 new regression test)
- [ ] Run @NicolasBissig's docker-compose repro against a build with this fix and confirm the parent workflow completes within seconds instead of stalling

### DIFF
--- a/core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
+++ b/core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
@@ -40,7 +40,6 @@ import com.netflix.conductor.dao.QueueDAO;
 import com.netflix.conductor.metrics.Monitors;
 import com.netflix.conductor.model.TaskModel;
 import com.netflix.conductor.model.WorkflowModel;
-import com.netflix.conductor.service.ExecutionLockService;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
@@ -65,7 +64,6 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
     private final ConductorProperties properties;
     private final ObjectMapper objectMapper;
     private SystemTaskRegistry systemTaskRegistry;
-    private final ExecutionLockService executionLockService;
     private final Clock clock = Clock.systemDefaultZone();
     private AtomicBoolean stop = new AtomicBoolean(false);
 
@@ -77,8 +75,7 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
             ConductorProperties properties,
             SweeperProperties sweeperProperties,
             SystemTaskRegistry systemTaskRegistry,
-            ObjectMapper objectMapper,
-            ExecutionLockService executionLockService) {
+            ObjectMapper objectMapper) {
         this.queueDAO = queueDAO;
         this.executionDAO = executionDAO;
         this.sweeperProperties = sweeperProperties;
@@ -88,7 +85,6 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
         this.properties = properties;
         this.systemTaskRegistry = systemTaskRegistry;
         this.objectMapper = objectMapper;
-        this.executionLockService = executionLockService;
         log.info("Initializing sweeper with {} threads", properties.getSweeperThreadCount());
         for (int i = 0; i < properties.getSweeperThreadCount(); i++) {
             sweeperExecutor.execute(this::pollAndSweep);
@@ -159,10 +155,6 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
     }
 
     public void sweep(String workflowId) {
-        if (!executionLockService.acquireLock(workflowId)) {
-            log.error("Couldn't acquire lock to sweep workflow {}", workflowId);
-            return;
-        }
         log.info("Running sweeper for workflow {}", workflowId);
 
         try {
@@ -295,8 +287,6 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
             queueDAO.remove(DECIDER_QUEUE, workflowId);
         } catch (Throwable e) {
             log.error("Error running sweep for {}, error = {}", workflowId, e.getMessage(), e);
-        } finally {
-            executionLockService.releaseLock(workflowId);
         }
     }
 
--- a/core/src/test/java/org/conductoross/conductor/core/execution/WorkflowSweeperTest.java
+++ b/core/src/test/java/org/conductoross/conductor/core/execution/WorkflowSweeperTest.java
@@ -180,14 +180,13 @@ public class WorkflowSweeperTest {
      *
      * <p>Previously, sweep() pre-acquired the execution lock and then called
      * workflowExecutor.decide(workflowId), which acquires the same lock internally. With a
-     * non-reentrant lock backend (Postgres advisory locks), the inner acquisition always
-     * returned false, decide() returned null, and the sweeper re-queued the workflow with a
-     * backoff delay — stalling workflows containing a SUB_WORKFLOW task forever after the
-     * sub-workflow completed.
+     * non-reentrant lock backend (Postgres advisory locks), the inner acquisition always returned
+     * false, decide() returned null, and the sweeper re-queued the workflow with a backoff delay —
+     * stalling workflows containing a SUB_WORKFLOW task forever after the sub-workflow completed.
      *
-     * <p>After the fix, sweep() no longer pre-acquires the lock; decide() handles locking
-     * itself. A SUB_WORKFLOW parent whose child has completed must advance to terminal in a
-     * single sweep cycle, with no backoff push to the decider queue.
+     * <p>After the fix, sweep() no longer pre-acquires the lock; decide() handles locking itself. A
+     * SUB_WORKFLOW parent whose child has completed must advance to terminal in a single sweep
+     * cycle, with no backoff push to the decider queue.
      */
     @Test
     public void sweepAdvancesParentWhenSubWorkflowCompletedWithNonReentrantLock() {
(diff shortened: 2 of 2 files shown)


NicolasBissig:
### Validation against the docker-compose reproduction (issue #1058) ✅

I ran the fix from this branch against the [docker-compose reproduction](https://github.com/NicolasBissig/conductor-sub-workflow-lock-bug) to confirm the deadlock is resolved end-to-end.

**Setup:**
- Base image: `conductoross/conductor:3.22.3`
- `CONDUCTOR_WORKFLOW_EXECUTION_LOCK_TYPE: postgres` (the bug condition, left enabled)
- Patched by compiling `WorkflowSweeper.java` from this branch and replacing the class files in `conductor-core-3.22.3.jar` inside the fat JAR

**Result:**

| Scenario | Outcome | Time to complete |
|---|---|---|
| Unpatched `3.22.3` | ❌ `BUG CONFIRMED` — still `RUNNING` after 120 s | ∞ (never completes) |
| This branch | ✅ `COMPLETED` | ~5 s |

Conductor logs with the fix show no `can't get a lock` loop — the sweeper picks up the completed sub-workflow and advances the parent in a single sweep cycle, exactly as intended.

The fix is confirmed. The root cause diagnosis in the PR description is accurate: removing the redundant `acquireLock` call from `sweep()` is the correct and minimal fix — `decide()` already handles its own locking.

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -159,10 +155,6 @@ public CompletableFuture<Void> sweepAsync(String workflowId) {
     }
 
     public void sweep(String workflowId) {
-        if (!executionLockService.acquireLock(workflowId)) {
```
mp-orkes, line 162: deadlock fixed, races introduced. 

We need to make The Postgres lock implementation reentrant.

mp-orkes:
We're going to address #1058 at the lock layer instead: see #1088, which makes PostgresLockDAO reentrant per-thread. 

**The Lock interface Javadoc  already says "re-entrant"** — PostgresLockDAO is not honoring that contract, which is exactly what makes sweep() self-deadlock.

PR #1100, opened 2026-05-17
perf(inline): cut Inline task latency via Context pool, shared Engine, Source cache, native deep-copy

v1r3n:
## Summary

Removes four redundant costs paid on every Inline task evaluation in `core/src/main/java/com/netflix/conductor/core/events/ScriptEvaluator.java`:

- **Per-call GraalVM Context creation.** The pool existed but was disabled by default; flipping the default reuses 10 pre-built Contexts.
- **Duplicate Truffle compilation caches.** Introduced a static shared `Engine` so all pooled Contexts share compilation state instead of each holding their own.
- **Repeated parsing of identical scripts.** Added a bounded `Map<String, Source>` cache keyed by script text so the same expression isn't recompiled per call. Configurable via `CONDUCTOR_SCRIPT_SOURCE_CACHE_SIZE` (default 1024).
- **Jackson serialize/deserialize round-trip for input isolation.** Replaced in `JavascriptEvaluator` and `GraalJSEvaluator` with a native recursive Map/List/array clone. Preserves the input-isolation contract tested by `testDeepCopyProtection`.

Adds an operator-facing INFO log at startup so it's visible whether the optimizing Truffle runtime is engaged (`implementation=GraalVM`) or the interpreter fallback is in effect (`implementation=Default`). Also declares `org.graalvm.truffle:truffle-runtime` explicitly (it was transitive via `org.graalvm.js:js`, but explicit makes intent clear).

## Why now

Inline task evaluation showed up as a hot path. Investigation found multiple compounding inefficiencies stacked on top of each other rather than a single root cause. This PR addresses all of them at once and ships a benchmark to keep future regressions honest.

## A note on the JIT

GraalJS only JIT-compiles when running on a GraalVM JDK or with `-XX:+EnableJVMCI --upgrade-module-path=<graal-compiler.jar>` on stock OpenJDK. The Graal compiler jar that would enable that is GPLv2+CE — incompatible with this project's distribution model — so we are choosing to run in interpreter mode. The startup log makes that explicit; the optimizations in this PR target the interpreter path.

## Benchmark

`core/src/test/java/com/netflix/conductor/benchmark/` contains a standalone JS engine benchmark covering 12 graduated scripts: trivial boolean returns, simple arithmetic, deep property access, conditionals, `filter`/`map`/`reduce` over 50-1000 items, string manipulation, nested-loop aggregation, and a ~80-line orders-aggregation pipeline run over 500 and 5000 orders.

Run with:
```
./gradlew :conductor-core:benchmarkScripts
```

The benchmark compares GraalJS (production path), Mozilla Rhino (MPL 2.0), and Javet/V8 (Apache 2.0 + BSD-3). Findings on macOS arm64 / OpenJDK 21:

| Engine | Verdict |
|---|---|
| GraalJS interpreter (this PR) | Best or tied on 9/12 cases; ~12 us per simple Inline call |
| Mozilla Rhino | 1.5-3.6x faster on scripts that build many objects in a loop; ~30% slower on simple expressions |
| Javet (V8) | 5-50x slower on iteration; JNI marshalling of Java Map inputs/outputs dwarfs V8's JIT advantage |

The benchmark concluded that adding Rhino as an opt-in evaluator would help only narrow workloads and confuse the per-task selection UX. Javet was ruled out entirely. No new evaluator is added in this PR.

The benchmark stays in the tree as a reusable tool for future engine-related decisions.

## Test plan

- [x] `./gradlew :conductor-core:test --tests "*JavascriptEvaluatorTest" --tests "*GraalJSEvaluatorTest" --tests "*InlineTest" --tests "*DoWhileTest" --tests "*DecisionTaskMapperTest" --tests "*LambdaTaskMapperTest" --tests "*DoWhileTaskMapperTest" --rerun-tasks` — all 73 tests pass (0 failures, 0 errors)
- [x] `testDeepCopyProtection` continues to pass — confirms isolation contract preserved across the native deep-copy switch
- [x] `testArrayOperations` continues to pass — confirms `int[]` -> `List<Integer>` conversion still happens for JS array methods
- [x] `./gradlew :conductor-core:benchmarkScripts` runs end-to-end, all engines load and execute, results are stable across 5x sample-size runs
- [ ] Verify in a real deployment that startup log shows `implementation=...` line so operators can see runtime mode

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -48,12 +48,15 @@ dependencies {
     implementation "com.google.guava:guava:${revGuava}"
 
     //GraalVM dependencies for executing JavaScript and Python
-    // PINNED (#964): all five artifacts must use the same version (revGraalVM in dependencies.gradle)
+    // PINNED (#964): all GraalVM artifacts must use the same version (revGraalVM in dependencies.gradle)
     implementation("org.graalvm.polyglot:polyglot:${revGraalVM}")
     implementation("org.graalvm.js:js:${revGraalVM}")
     implementation("org.graalvm.js:js-scriptengine:${revGraalVM}")
     implementation("org.graalvm.polyglot:python:${revGraalVM}")
     implementation "org.graalvm.sdk:graal-sdk:${revGraalVM}"
+    // Optimizing Truffle runtime (UPL 1.0). Without it GraalJS runs in the Truffle
+    // interpreter, which is 10-100x slower than the JIT-compiled path.
+    implementation("org.graalvm.truffle:truffle-runtime:${revGraalVM}")
 
     // JAXB is not bundled with Java 11, dependencies added explicitly
     // These are needed by Apache BVAL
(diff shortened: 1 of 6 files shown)


manan164 (approved):

c4lm:
> engine-related decisions

I think it would be great if no new in-jvm engines* are added back in the future
The most likely future path will be an engine living outside the jvm IMO

c4lm (approved):

v1r3n:
> > engine-related decisions
> 
> I think it would be great if no new in-jvm engines* are added back in the future The most likely future path will be an engine living outside the jvm IMO

+1 the benchmark also proved that adding Rhino or V8 added no/little benefit.  The future path should be an engine living outside of JVM.

Issue #752, opened 2026-02-03
Epic: Migrate from com.netflix.conductor to org.conductoross.conductor namespace


nthmost-orkes:
## Motivation

Conductor OSS is no longer maintained by Netflix and is now an independent open-source project maintained by the community. The package namespace should reflect this ownership change:

- **Current (legacy)**: `com.netflix.conductor.*` - implies Netflix ownership
- **Target (modern)**: `org.conductoross.conductor.*` - reflects community ownership

This epic tracks the incremental migration of all modules to the new namespace.

## Current State

The codebase is in a **mixed state** with both namespaces:

### ✅ Fully Migrated (5 modules)

| Module | Status | Notes |
|---|---|---|
| `os-persistence-v2` | ✅ Complete | New module (PR #749) |
| `os-persistence-v3` | ✅ Complete | New module (PR #749) |
| `common` (partial) | 🔄 In Progress | Some new models use new namespace |
| `core` (partial) | 🔄 In Progress | New execution/task code uses new namespace |
| `rest` (partial) | 🔄 In Progress | Some controllers migrated |

### ❌ Not Yet Migrated (34 modules)

**Core Infrastructure** (7):
- [ ] `annotations`
- [ ] `annotations-processor`
- [ ] `server`
- [ ] `server-lite`
- [ ] `grpc`
- [ ] `grpc-server`
- [ ] `grpc-client`

**Persistence Backends** (6):
- [ ] `cassandra-persistence`
- [ ] `redis-persistence`
- [ ] `mysql-persistence`
- [ ] `postgres-persistence`
- [ ] `sqlite-persistence`
- [ ] `es7-persistence`

**Storage** (3):
- [ ] `awss3-storage`
- [ ] `azureblob-storage`
- [ ] `postgres-external-storage`

**Event Queues** (6):
- [ ] `awssqs-event-queue`
- [ ] `kafka-event-queue`
- [ ] `amqp`
- [ ] `nats`
- [ ] `nats-streaming`
- [ ] `kafka`

**Tasks & Extensions** (4):
- [ ] `http-task`
- [ ] `json-jq-task`
- [ ] `workflow-event-listener`
- [ ] `task-status-listener`

**Locking & Concurrency** (2):
- [ ] `redis-lock`
- [ ] `redis-concurrency-limit`

**Metrics** (1):
- [ ] `metrics`

**Testing** (2):
- [ ] `test-util`
- [ ] `test-harness`

**Deprecated Stubs** (3):
[...]

PR #1107, opened 2026-05-19
Publish interfaces

JamesHarris-orkes:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

https://github.com/conductor-oss/conductor/issues/1108
----

Add change publisher interfaces to workflow definitions, task definitions, and schedules.
----

This is to capture all changes to workflow definitions, task definitions and schedules. No alternative seems readily available.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -60,6 +60,22 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 'lts/*'
+
+      - name: Build ui-next and embed into server static resources
+        run: |
+          corepack enable
+          cd ui-next
+          pnpm install
+          NODE_OPTIONS=--max-old-space-size=4096 pnpm build
+          cd ..
+          mkdir -p server/src/main/resources/static
+          rm -rf server/src/main/resources/static/*
+          cp -r ui-next/dist/. server/src/main/resources/static/
+
       - name: Build Server JAR
         run: |
           export VERSION="${{github.ref_name}}"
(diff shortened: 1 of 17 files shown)


nthmost-orkes (commented):
Quick question on the enum types — `MetadataChangeType`, `MetadataResourceType`, and `ScheduleChangeType` are defined in the new interfaces but don't appear to be referenced anywhere yet (no usages in the listener methods, stubs, or tests). Are these placeholders for a follow-on PR that'll use them in an event payload, or is there a usage I'm missing?

JamesHarris-orkes:
> Quick question on the enum types — `MetadataChangeType`, `MetadataResourceType`, and `ScheduleChangeType` are defined in the new interfaces but don't appear to be referenced anywhere yet (no usages in the listener methods, stubs, or tests). Are these placeholders for a follow-on PR that'll use them in an event payload, or is there a usage I'm missing?

I am expecting they will be used by the implementations and could have been used for serialization. I will remove them from the interface.

nthmost-orkes:
One thing on placement: it looks like we could move these into the `org.conductoross` namespace, actually. `MetadataChangeListener` and `MetadataChangeListenerStub` are brand-new code, and `core` already has a bunch of `org.conductoross.conductor.core.*` packages (`execution`, `exception`, `storage`, `dao`, `service`…), so `org.conductoross.conductor.core.listener` would slot right in. It's only sitting in `com.netflix.conductor.core.listener` because it landed next to the older `TaskStatusListener`/`WorkflowStatusListener` — but those are legacy and stay put; new code is supposed to go conductoross-side. This is part of the broader push in #752.

The move is small: just the `package` line on the two new files plus import fixups in `ConductorCoreConfiguration`, `MetadataServiceImpl`, and `MetadataServiceTest`. Splitting the listener family across the two namespaces is fine — that's the expected shape during the migration.

(`ScheduleChangeListener` is correctly placed, to be clear — the whole `scheduler/core` module lives under `io.orkes.conductor.scheduler.*`, so it's matching its module, not the netflix namespace.)

JamesHarris-orkes:
Thanks @nthmost-orkes — both points are handled in `7c0a4913b`:

- **Unused enums**: removed `MetadataChangeType`, `MetadataResourceType`, and `ScheduleChangeType` from the interfaces. They'll come back with the implementation that actually needs them (likely for serialization) rather than sitting unused here.
- **Namespace**: moved `MetadataChangeListener` and `MetadataChangeListenerStub` to `org.conductoross.conductor.core.listener`, with import fixups in `ConductorCoreConfiguration`, `MetadataServiceImpl`, and `MetadataServiceTest`. Left `ScheduleChangeListener` in `io.orkes.conductor.scheduler.*` as you noted.

CI is green. Ready for another look when you have a moment.

nthmost-orkes (approved):
i think we're good!  Thanks for your patience working through this

PR #1116, opened 2026-05-21
Introduce a Self-Tuning Adaptive Event Scheduler that decouples event…

hchepey-clari:
Introduce a Self-Tuning Adaptive Event Scheduler that decouples event cardinality from thread count and provides max-min fairness across event types without any operator-visible configuration.

Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Problem
The current event framework spins one RxJava subscription per registered event, each bound to a fixed-size scheduler pool. Two consequences: Thread exhaustion at high cardinality. With ~100 event types and a default pool of 2 threads, queues registered after the pool fills are never serviced. Adding event N+1 is silently broken at runtime. Head-of-line blocking under skewed load. When one event type has 10,000 buffered messages and another has 10, the cold event waits behind the hot one because there is no fairness layer; the shared scheduler dispatches strictly FIFO across queues. 

Solution
Introduce a Self-Tuning Adaptive Event Scheduler that decouples event cardinality from thread count and provides max-min fairness across event types without any operator-visible configuration. The design has three loops, each with one responsibility: Loop	Responsibility	Mechanism
Poller	Pull messages from each queue at a rate the consumer can absorb	Passive ObservableQueue.poll(batchSize, timeout) sized by per-lane buffer headroom Dispatcher	Allocate worker slots fairly across lanes	Deficit Round Robin, quantum = 1 Workers	Execute handler logic	Virtual-thread executor (Java 21+) Per-lane elasticity comes from a Vegas-style SimpleLimiter (Netflix concurrency-limits) that adjusts the in-flight ceiling using Little's Law from observed latency. DRR holds fairness (each lane visited every round); Vegas holds capacity (each lane sized to its own service curve). Neither requires user input.

Issue #

Alternatives considered
----


hchepey-clari:
@nthmost-orkes @v1r3n - Is there anything I can help with so that we can get this PR merged ASAP?

nthmost-orkes:
Hi @hchepey-clari -- Since you put up this PR, we have introduced a policy of only allowing PRs that are attached to issues explaining the need and allowing others both within the community and within Orkes to add their perspectives first.

That's going to explain a lot of the delay in circling back on this, we just need to have places to discuss and stage feature adds over time.

Thanks for understanding.  Again, if you can file an issue clarifying your own motivation for this PR and how it would influence your usage of OSS that will be extremely useful.  Thanks!

PR #1122, opened 2026-05-22
feat: add scheduled workflow history cleaner

### LINKED ISSUE #1081: [FEATURE]: Add periodic batch job to clean up old terminal workflow history
## Describe the Feature Request

In long-running Conductor deployments, terminal workflows (`COMPLETED` / `FAILED` / `TERMINATED` / `TIMED_OUT`) accumulate indefinitely in both the execution store (e.g. MySQL `workflow` / `task` tables) and the search index (`conductor_workflow`, `conductor_task`). Over time this causes:

- Storage bloat and rising query latency on the execution DB
- Larger primary shards and degraded search performance on the index
- Higher operational cost for retaining data well past its usefulness window

Conductor today ships `ArchivingWorkflowStatusListener` (in `conductor-community`), which archives a workflow at the moment it reaches a terminal state. That is useful for the "hot path", but it does **not** address:

1. **Backfill** — workflows that completed *before* the listener was enabled, or that were missed due to event delivery failures, are never cleaned up.
2. **Retention policy** — operators typically want to keep workflows for N days after completion for debugging/audit and *then* drop them. An event listener can only archive immediately, not on a delay.
3. **Idempotency / retry** — a failed archive attempt at the listener has no second chance.

I'd like to propose a complementary, opt-in **scheduled batch component** that periodically deletes terminal workflows older than a configurable retention window, from both the execution store and the search index, using only existing Conductor abstractions.

Related issues / discussions:
- #679 — Audit logging for workflow deletions
- #596 — Delete Workflow Execution History via UI causing duplicated task execution
- #272 — Workflow removal does not remove all children
- Netflix/conductor#1315 — Best practice: cleanup of redis and elasticsearch
- Netflix/conductor#1294 — Elasticsearch rollover & workflow archival

## Describe Preferred Solution

Add a scheduled component (e.g. `WorkflowHistoryCleaner`) that periodically removes terminal workflows older than a configurable retention threshold, with the following properties:

**Built on existing abstractions** — no new persistence APIs required.
- Discover candidates via `IndexDAO#searchArchivableWorkflows(indexName, ttlDays)`
- Delete via `ExecutionDAOFacade#removeWorkflow(id, archive)`, which already cleans both the execution store (sync) and the search index (async)

**Pluggable distributed locking** — depends only on `com.netflix.conductor.core.sync.Lock` (already used by `WorkflowSweeper`), so any concrete implementation (Redis / JDBC / ZooKeeper / Local / Noop) works. A separate `Lock` bean is used so cleanup TTL is independent of the sweeper's lock.

**Single-instance-at-a-time execution** — `Lock.acquireLock` with `timeToTry=0`; runs are skipped (not queued) when another instance holds the lock.

**Configurable retention & cadence** — Spring `@Scheduled` cron + retention days + a per-day catch-up window for backfill.

**Graceful shutdown** — implements `LifecycleAwareComponent` and re-checks `isRunning()` between batches; in-flight runs abort cleanly on JVM shutdown.

**Disabled by default** — `conductor.workflow-history-cleanup.enabled=false` so existing deployments are unaffected.

**Safe iteration** — bounded `maxIterationsPerDay`, a recent-id cache to avoid spinning on async index-delete lag, and a configurable pause between iterations to limit DB pressure.

### Proposed properties (illustrative)

| Key | Default | Purpose |
|---|---|---|
| `conductor.workflow-history-cleanup.enabled` | `false` | Master switch |
| `conductor.workflow-history-cleanup.cron` | `0 0 * * * *` | Schedule |
| `conductor.workflow-history-cleanup.retention-days` | `30` | Days after terminal state before deletion |
| `conductor.workflow-history-cleanup.catch-up-days` | `7` | Sliding window to also process older buckets per run (backfill) |
| `conductor.workflow-history-cleanup.lock-lease-time` | `2h` | Cleanup-lock TTL |
| `conductor.workflow-history-cleanup.max-iterations-per-day` | `100` | Safety cap per (day × run) |
| `conductor.workflow-history-cleanup.batch-pause` | `1s` | Pause between iterations |

## Describe Alternatives

- **Index-level TTL only (e.g. OpenSearch ISM / Elasticsearch ILM)** — only addresses the search index; the execution store keeps growing.
- **Manual cleanup scripts** — error-prone, not coordinated across instances, no observability.
- **Extending `ArchivingWorkflowStatusListener`** — event-driven; cannot backfill historical data or honor a delayed retention window.
- **External cron + API calls** — possible, but ties the operator to a specific scheduler, reimplements distributed locking, and loses access to Conductor's internal metrics and lifecycle hooks.

### Additional context

_No response_

youngxpepp:
## Summary

Closes #1081.

Adds `WorkflowHistoryCleaner`, a Spring `@Scheduled` component that periodically deletes terminal workflows (`COMPLETED`/`FAILED`/`TERMINATED`/`TIMED_OUT`) older than the configured retention threshold.

- Pages candidates with `IndexDAO#searchArchivableWorkflows` and delegates the actual removal to `ExecutionDAOFacade#removeWorkflow`, reusing the existing execution-store / index ordering guarantees.
- Single-instance execution is enforced through the pluggable `com.netflix.conductor.core.sync.Lock` abstraction with **non-blocking `tryLock` semantics** — contenders exit immediately instead of queueing on the lock.
- Cooperative shutdown via `LifecycleAwareComponent`: any in-flight run checks `isRunning()` between iterations and at workflow boundaries.
- Safety rails: bounded iterations per day, fixed-size LRU of recently-processed ids (mitigates async index-delete lag), and configurable batch pauses.
- Disabled by default to keep existing deployments unaffected; opt in with `conductor.workflow-history-cleanup.enabled=true`.

Rolling indices (`conductor_task_log_*`, `conductor_event_*`, `conductor_message_*`) are out of scope — manage their retention through index-level lifecycle policies (e.g. OpenSearch ISM).

## Key Properties

| Property | Default | Description |
|---|---|---|
| `conductor.workflow-history-cleanup.enabled` | `false` | Master switch |
| `conductor.workflow-history-cleanup.cron` | `0 0 * * * *` | Schedule (top of the hour) |
| `conductor.workflow-history-cleanup.zone` | `UTC` | Cron time zone |
| `conductor.workflow-history-cleanup.retention-days` | `30` | Lower bound of the deletion window |
| `conductor.workflow-history-cleanup.catch-up-days` | `7` | Number of day-buckets walked per run |
| `conductor.workflow-history-cleanup.max-iterations-per-day` | `200` | Safety cap on the iteration loop |
| `conductor.workflow-history-cleanup.processed-cache-size` | `5000` | LRU size for recently-processed ids |
| `conductor.workflow-history-cleanup.batch-pause` | `500ms` | Pause between iterations |
| `conductor.workflow-history-cleanup.index-refresh-wait` | `2s` | Wait when the same ids keep coming back |
| `conductor.workflow-history-cleanup.lock-id` | `workflow-history-cleanup` | Lock identifier |
| `conductor.workflow-history-cleanup.lock-lease-time` | `2h` | Lock TTL |
| `conductor.workflow-history-cleanup.workflow-index-name` | (derived from `index-prefix`) | Override only if needed |

`SchedulerConfiguration` thread pool size bumped from 3 to 4 to accommodate the new scheduled job.

## Test plan
- [x] `./gradlew :conductor-core:test` — 18 new test cases in `TestWorkflowHistoryCleaner` cover lock contention, async index-delete lag, mid-run shutdown, exception handling, iteration caps, and null-result edge cases.
- [x] `./gradlew :conductor-core:spotlessCheck`
- [x] `./gradlew :conductor-core:compileJava :conductor-core:compileTestJava`
- [ ] Manual smoke test against a real Conductor deployment with OpenSearch (reviewer to validate end-to-end behavior)

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/SchedulerConfiguration.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/SchedulerConfiguration.java
@@ -67,7 +67,7 @@ public class SchedulerConfiguration implements SchedulingConfigurer {
     @Override
     public void configureTasks(ScheduledTaskRegistrar taskRegistrar) {
         ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
-        threadPoolTaskScheduler.setPoolSize(3); // equal to the number of scheduled jobs
+        threadPoolTaskScheduler.setPoolSize(4); // equal to the number of scheduled jobs
         threadPoolTaskScheduler.setThreadNamePrefix("scheduled-task-pool-");
         threadPoolTaskScheduler.initialize();
         taskRegistrar.setTaskScheduler(threadPoolTaskScheduler);
--- a/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowHistoryCleaner.java
+++ b/core/src/main/java/com/netflix/conductor/core/reconciliation/WorkflowHistoryCleaner.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright 2026 Conductor Authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.core.reconciliation;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.core.env.Environment;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+import com.netflix.conductor.core.LifecycleAwareComponent;
+import com.netflix.conductor.core.dal.ExecutionDAOFacade;
+import com.netflix.conductor.core.sync.Lock;
+import com.netflix.conductor.dao.IndexDAO;
+import com.netflix.conductor.metrics.Monitors;
+
+/**
+ * Periodically removes terminal workflows (and their tasks) that are older than the configured
+ * retention window.
+ *
+ * <p>The batch is intentionally thin: it pages over candidate ids using {@link
+ * IndexDAO#searchArchivableWorkflows(String, long)} and delegates the actual removal to {@link
+ * ExecutionDAOFacade#removeWorkflow(String, boolean)}, which already coordinates the execution
+ * store and the search index in the right order.
+ *
+ * <p>Distributed safety is provided by the pluggable {@link Lock} abstraction so only one instance
+ * runs per scheduling tick — failed acquisitions back off immediately rather than queueing. The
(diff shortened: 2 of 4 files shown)


youngxpepp:
@akhilpathivada Could you take a look when you get a chance? This PR addresses #1081 — based on your earlier comment on the issue, I'd appreciate your review on the initial scope (global TTL only; field-value-based retention left as a follow-up).

vivianzheng404:
Can someone review this PR? This feature is awesome! We need it too!

PR #1148, opened 2026-06-03
test(test-harness): fix flaky parent-JOIN assertions in HierarchicalForkJoinSubworkflowRetrySpec

JamesHarris-orkes:
## Problem

`HierarchicalForkJoinSubworkflowRetrySpec` is flaky in CI. The `test-harness` job intermittently fails on assertions like:

```
Condition not satisfied:
tasks[3].status == Task.Status.CANCELED
|    |   |      |
|    |   IN_PROGRESS
```

i.e. a parent workflow's `JOIN` task is found `IN_PROGRESS` when the test expected the transient `CANCELED` state.

## Root cause

`WorkflowExecutorOps.retry(...)` updates parent workflows **asynchronously**: `updateAndPushParents` pushes each parent onto the decider queue via `expediteLazyWorkflowEvaluation` rather than deciding inline. The integration tests run with a live background sweeper (`conductor.app.sweeperThreadCount=1`), so a parent's `JOIN` — left `CANCELED` from the original failure — can be re-decided to `IN_PROGRESS` **before** the assertion executes.

This makes every parent-JOIN check performed immediately after `retry(...)` (and before the explicit `sweep(...)`) racy. The "retry on the root with resume flag" case is the one that tripped in CI, but all five retry tests in the spec share the identical race.

## Fix

Relax the transient parent-JOIN assertions to accept either state:

```groovy
tasks[3].status in [Task.Status.CANCELED, Task.Status.IN_PROGRESS]
```

This is test-only; no production code changes. Deterministic assertions are intentionally left strict:
- the terminal `FAILED` state asserted in `setup()` (JOIN is cancelled synchronously at termination), and
- the converged `IN_PROGRESS` state asserted after the explicit `sweep(...)` calls.

So the cascade behavior under test (retry propagating up the parent chain, JOINs converging to `IN_PROGRESS`, workflows completing) is still fully verified.

## Verification
- `./gradlew :conductor-test-harness:compileTestGroovy` → `BUILD SUCCESSFUL`.

### DIFF
--- a/test-harness/src/test/groovy/com/netflix/conductor/test/integration/HierarchicalForkJoinSubworkflowRetrySpec.groovy
+++ b/test-harness/src/test/groovy/com/netflix/conductor/test/integration/HierarchicalForkJoinSubworkflowRetrySpec.groovy
@@ -12,6 +12,8 @@
  */
 package com.netflix.conductor.test.integration
 
+import java.util.concurrent.TimeUnit
+
 import org.springframework.beans.factory.annotation.Autowired
 
 import com.netflix.conductor.common.metadata.tasks.Task
@@ -28,6 +30,8 @@ import static com.netflix.conductor.common.metadata.tasks.TaskType.TASK_TYPE_FOR
 import static com.netflix.conductor.common.metadata.tasks.TaskType.TASK_TYPE_JOIN
 import static com.netflix.conductor.common.metadata.tasks.TaskType.TASK_TYPE_SUB_WORKFLOW
 import static com.netflix.conductor.test.util.WorkflowTestUtil.verifyPolledAndAcknowledgedTask
+
+import static org.awaitility.Awaitility.await
 
 class HierarchicalForkJoinSubworkflowRetrySpec extends AbstractSpecification {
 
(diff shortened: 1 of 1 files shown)


JamesHarris-orkes:
Updated this PR to use the `await().atMost(10s)` polling style instead of relaxing the assertion to accept both states.

### Relationship to #1047
This is the **same family** of flake as #1047 — the background `WorkflowSweeper` (introduced in `99ecc5f12`) racing with a parent workflow that was pushed to `DECIDER_QUEUE` via `expediteLazyWorkflowEvaluation()` — but it is **not the same root cause**:

- **#1047** is a genuine **product bug** in `WorkflowExecutorOps.rerunWF()`: `updateAndPushParents()` ran *before* `executionDAOFacade.updateWorkflow()`, so the sweeper observed the child still FAILED and permanently cascade-failed the parent. Its fix is in production code.
- **This PR** is **test-only**. The `retry()` path has no ordering bug — the child is persisted RUNNING (`WorkflowExecutorOps.java` `updateWorkflow`/`updateTasks`) *before* `updateAndPushParents()` pushes the parents. The sweeper then correctly re-decides each parent's JOIN from `CANCELED` to `IN_PROGRESS`; the test was simply asserting the transient `CANCELED` state before that valid async transition.

The hardening here mirrors the `await()` pattern #1047 added to `SubWorkflowRerunSpec`, and the pre-existing pattern in `HierarchicalForkJoinSubworkflowRestartSpec`, so the three specs now handle the sweeper race consistently.

cc #1047

nthmost-orkes (approved):
Yep, the race is exactly as you've diagnosed: `retry()` hands parent re-decisions off to the sweeper queue asynchronously, so asserting the transient `CANCELED` state before a `sweep()` call is always a coin flip. Swapping in Awaitility to wait for the stable `IN_PROGRESS` is the right call, and it's consistent with how the same class of race was handled in the sibling specs — nice clean fix.

PR #1157, opened 2026-06-08
fix(webhook): accept any 2XX status code in webhook response (#966)

### LINKED ISSUE #966: Accept any 2XX code in webhook response

Accept any 2XX status code in the webhook response; currently, I believe only 200 and 201 are accepted

### Alternatives

_No response_

### Additional context

_No response_

mtalhasid:
## Problem
Webhook response validation in `RestClientManager.java` only accepted 
`HttpStatus.SC_OK` (200) and `HttpStatus.SC_ACCEPTED` (202), causing 
valid 2XX responses like 201, 204 etc. to be incorrectly treated as failures.

## Fix
Updated validation to accept any 2XX status code (200–299) instead of 
hardcoded values, making it consistent with how `HttpTask.java` and 
`HttpDocumentLoader.java` already handle HTTP responses in this codebase.

## Changes
- `task-status-listener/.../RestClientManager.java` — updated `executePost()` 
status check from `SC_OK || SC_ACCEPTED` to `>= 200 && < 300`

## Screenshot
Screenshot 2026-06-08 100136

Fixes #966


nthmost-orkes (dismissed):
Makes sense!

One thing to fix before this can merge: line 246 was the only use of `HttpStatus` in the file, so the import at line 26 is now unused:

```java
import org.apache.http.HttpStatus;
```

Spotless has `removeUnusedImports()` enabled in the root `build.gradle`, so `spotlessJavaCheck` will fail CI until that import is removed. Drop that line and you should be green.

(FWIW — in case anyone wonders why not enumerate the named constants like `SC_OK || SC_CREATED || SC_NO_CONTENT || ...`: the range check is deliberately better here. The intent is "the whole 2XX class," and an explicit list is brittle in exactly the way that caused this bug — list 200/202, forget 201, and valid responses get rejected. Enumerating just trades "forgot 201" for "forgot 207." The range can't have that gap.)

Once the import's gone this is good to go. Thanks for the fix!

talhadevelopes:
Pushed a fix removing the unused `HttpStatus` import as you flagged , `spotlessJavaCheck` passes locally now. Should be good to go, thanks for catching that!

Screenshot 2026-06-24 112626

nthmost-orkes:
> Pushed a fix removing the unused `HttpStatus` import as you flagged , `spotlessJavaCheck` passes locally now. Should be good to go, thanks for catching that!
> 

Build is still failing, so once that clears we can approve and merge this. Thanks!

talhadevelopes:
Synced this branch with main to pick up the recent hierarchical fork/join
de-flake fix (#1206), in case that was related to the test-harness failure
on this PR. Pushed,  just waiting on workflow approval to see the result.

nthmost-orkes (approved):
ty!

mtalhasid:
Hi @nthmost-orkes , just checking in, this has been approved and green for a while now. Is there anything still needed from my side, or is it just queued behind other work? Happy to rebase again if needed.

nthmost-orkes:
> Hi @nthmost-orkes , just checking in, this has been approved and green for a while now. Is there anything still needed from my side, or is it just queued behind other work? Happy to rebase again if needed.

Was just queued, thanks for your patience!

PR #1158, opened 2026-06-08
fix: SWITCH task no longer falls through to defaultCase when matched case is empty

### LINKED ISSUE #396: [BUG]: SWITCH task goes through defaultCase if other cases are empty
Hello all,

found a bug with task type SWITCH. If the defaultCase contains some task but the other cases do not, the execution will go through the defaultCase even if the SWITCH task correctly evaluated to one of the other branches.

A simple workflow to reproduce:

```
{
  "createTime": 1739485449420,
  "updateTime": 1739527406013,
  "name": "test_qc_diamond",
  "description": "Edit or extend this sample workflow. Set the workflow name to get started",
  "version": 17,
  "tasks": [
    {
      "name": "qc_action",
      "taskReferenceName": "qc_action",
      "inputParameters": {
        "qc_action_var": "submit"
      },
      "type": "SWITCH",
      "decisionCases": {
        "true": []
      },
      "defaultCase": [
        {
          "name": "terminate_workflow",
          "taskReferenceName": "no_submit_qc_fail",
          "inputParameters": {
            "terminationStatus": "FAILED",
            "terminationReason": "Task QC output was failed.",
            "workflowOutput": "terminated"
          },
          "type": "TERMINATE",
          "decisionCases": {},
          "defaultCase": [],
          "forkTasks": [],
          "startDelay": 0,
          "joinOn": [],
          "optional": false,
          "asyncComplete": false,
          "loopOver": [],
          "onStateChange": {},
          "permissive": false
        }
      ],
      "forkTasks": [],
      "startDelay": 0,
      "joinOn": [],
      "optional": false,
      "asyncComplete": false,
      "loopOver": [],
      "evaluatorType": "javascript",
      "expression": "$.qc_action_var == 'submit' ? true : false",
      "onStateChange": {},
      "permissive": false
    }
  ],
  "inputParameters": [],
  "outputParameters": {},
  "schemaVersion": 2,
  "restartable": true,
  "workflowStatusListenerEnabled": false,
  "ownerEmail": "example@email.com",
  "timeoutPolicy": "ALERT_ONLY",
  "timeoutSeconds": 0,
  "variables": {},
  "inputTemplate": {},
  "enforceSchema": true
}
```
Takes no inputs, just run it from workbench.

This workflow always goes through the `defaultCase` branch, thus resulting in a workflows status of FAILED, even though it should go through the "true" case and thus skip the termination task and go to the end of workflow having status of COMPLETED.

Please see that even the diagram is wrong:

Image

The evaluation is correctly `true`, but the termination task is ticked and the arrow is green both on the `true` branch and going from the termination task towards the end of workflow.

Interestingly, if I place at least one task inside the `true` branch, the workflow will correctly execute through that branch.

If this is intended behaviour, it is not documented. I consider it a bug.

I have also tried with the `value-param` evaluator, but the behavior was the same.

**Details**
Conductor version: 3.21.11 + 69 commits 
Queue implementation: Redis
Lock: Redis

**Expected behavior**
The workflow should go through the `true` branch.

nthmost-orkes:
## Summary

- When a `SWITCH` task evaluates to a case key that exists in `decisionCases` but has an empty task list (e.g. `"true": []`), the workflow was incorrectly executing the `defaultCase` tasks.
- Root cause: `SwitchTaskMapper` treated `selectedTasks.isEmpty()` the same as `selectedTasks == null`, collapsing the distinction between "no key matched" and "key matched, no tasks."
- Fix: remove the `isEmpty()` check so the fallback only triggers on `null` (no matching key).

## User impact

Users who intentionally defined an empty SWITCH branch (a "no-op" path) saw their workflow take the `defaultCase` path instead — potentially triggering termination tasks or other unintended steps even though the expression evaluated correctly.

## Changes

- `SwitchTaskMapper.java`: remove `|| selectedTasks.isEmpty()` from the defaultCase fallback condition
- `SwitchTaskMapperTest.java`: add `getMappedTasksWithEmptyMatchedCase` test that asserts only the SWITCH task itself is scheduled (not the defaultCase task) when the matched case has no tasks

Fixes #396

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/mapper/SwitchTaskMapper.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/mapper/SwitchTaskMapper.java
@@ -115,9 +115,9 @@ public class SwitchTaskMapper implements TaskMapper {
 
         // get the list of tasks based on the evaluated expression
         List<WorkflowTask> selectedTasks = workflowTask.getDecisionCases().get(evalResult);
-        // if the tasks returned are empty based on evaluated result, then get the default case if
-        // there is one
-        if (selectedTasks == null || selectedTasks.isEmpty()) {
+        // fall back to defaultCase only when no case key matched (null); an explicitly empty case
+        // list means the branch intentionally has no tasks and should not execute the default
+        if (selectedTasks == null) {
             selectedTasks = workflowTask.getDefaultCase();
         }
         // once there are selected tasks that need to proceeded as part of the switch, get the next
--- a/core/src/test/java/com/netflix/conductor/core/execution/mapper/SwitchTaskMapperTest.java
+++ b/core/src/test/java/com/netflix/conductor/core/execution/mapper/SwitchTaskMapperTest.java
@@ -244,6 +244,46 @@ public class SwitchTaskMapperTest {
     }
 
     @Test
+    public void getMappedTasksWithEmptyMatchedCase() {
+        // A case key matches but has no tasks — should NOT fall through to defaultCase.
+        WorkflowTask switchTask = new WorkflowTask();
+        switchTask.setType(TaskType.SWITCH.name());
+        switchTask.setName("Switch");
+        switchTask.setTaskReferenceName("switchTask");
+        switchTask.setDefaultCase(Collections.singletonList(task1));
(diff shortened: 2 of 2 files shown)


nthmost-orkes:
Closing to reopen with a clean commit history.

PR #745, opened 2026-01-30
Support Agentic flows

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] WHOSUSING.md
- [ ] Other (please describe):
**NOTE**: `./gradlew spotlessApply` has been run to fix format violations.
Changes in this PR
----
_Describe the new behavior from this PR, and why it's needed_
This PR introduces the **AI Module** (`conductor-ai`) to enable LLM-powered workflows and agents within Conductor.
**Key Features:**
*   **Vector Database Support:** Native integration with **MongoDB**, **Pinecone**, and **Postgres** for vector storage and retrieval.
*   **AI System Tasks:** New task mappers for core AI operations:
    *   `LLM_GENERATE_EMBEDDINGS`: Generate vector embeddings from text.
    *   `LLM_STORE_EMBEDDINGS`: Store embeddings in the configured vector DB.
    *   `LLM_SEARCH_INDEX`: Perform semantic search against vector indexes.
    *   `LLM_TEXT_COMPLETE`: Execute LLM text completion.
    *   `LLM_INDEX_TEXT`: High-level text indexing task.
*   **Annotated System Tasks:** Enhanced `AnnotatedSystemTaskWorker` infrastructure to simplify system task development.
*   **Task Context:** Introduced `TaskContext` to improve context propagation across task execution.
*   **Utilities:** Added `JsonSchemaValidator`, `StringTemplate`, and `TextUtils`.
These changes provide the foundational building blocks for creating AI Agents and orchestrating complex LLM interactions.
Issue #
Alternatives considered
----
_Describe alternative implementation you have considered_
N/A - Initial implementation of the AI module.

PR #1198, opened 2026-06-22
fix(ai): remove manual Accept-Encoding header so Mistral gzip responses decode

martinconstam:
## Problem

Any LLM task (`LLM_TEXT_COMPLETE`, `LLM_CHAT_COMPLETE`, …) using the **Mistral** provider always fails with:

```
Error while extracting response for type [org.springframework.ai.mistralai.api.MistralAiApi$ChatCompletion]
and content type [application/json] — caused by: Illegal character ((CTRL-CHAR, code 31)):
only regular white space (\r, \n, \t) is allowed between tokens
```

`code 31` is `0x1F` — the first byte of the gzip magic number (`0x1F 0x8B`). Jackson is being handed a gzip-compressed body that was never decompressed.

## Root cause

`MistralAI#createMistralAiApi` builds the `RestClient` on an `OkHttp3ClientHttpRequestFactory` and additionally sets the `Accept-Encoding` header by hand:

```java
RestClient.builder()
    .requestFactory(factory)
    .defaultHeader("Accept-Encoding", "gzip, deflate")
```

OkHttp only decompresses a gzip response **transparently when it adds the `Accept-Encoding` header itself** (via its `BridgeInterceptor`). Setting the header explicitly turns that automatic decompression off, so the gzipped body is passed straight to Jackson and parsing fails on the `0x1F` byte.

The header was originally added to work around spring-ai#372, but it is unnecessary — and now actively harmful — since the client uses the OkHttp request factory, which handles content encoding on its own.

## Fix

Remove the manual `Accept-Encoding` header and let OkHttp manage compression. Mistral responses then decode correctly.

## Tests

Adds `MistralAIGzipResponseTest`, a regression test using `MockWebServer` that serves a **gzip-encoded** chat completion and asserts it is decoded correctly. It fails against the old code (gzip body reaches Jackson undecoded → `code 31`) and passes with this change. The test needs no API key, so it runs in CI — unlike the existing `MistralAITest.IntegrationTests`, which are gated on `MISTRAL_API_KEY` and therefore skipped in CI (which is why this regression went unnoticed).

## Impact / scope

- Present since the initial AI module commit ("Support Agentic flows", #745); affects every release that ships the AI module (3.30.0 onward, including 3.31.0-rc.x).
- Mistral-provider specific; other providers are unaffected by this line.
- Verified against the live Mistral API and via the new test: responses parse once OkHttp manages the encoding.

### DIFF
--- a/ai/src/main/java/org/conductoross/conductor/ai/providers/mistral/MistralAI.java
+++ b/ai/src/main/java/org/conductoross/conductor/ai/providers/mistral/MistralAI.java
@@ -123,15 +123,16 @@ public class MistralAI implements AIModel {
                         : httpClient;
         var factory =
                 new org.springframework.http.client.OkHttp3ClientHttpRequestFactory(effective);
-        // Needs accept-encoding headers
-        // https://github.com/spring-projects/spring-ai/issues/372
+        // Do NOT set the Accept-Encoding header manually here. OkHttp only
+        // decompresses a gzip response transparently when it adds the
+        // Accept-Encoding header itself. Setting it explicitly turns off that
+        // automatic decompression, so the gzipped body reaches Jackson undecoded
+        // and parsing fails with "Illegal character ((CTRL-CHAR, code 31))".
+        // Letting OkHttp manage compression makes responses parse correctly.
         return MistralAiApi.builder()
                 .baseUrl(config.getBaseURL())
                 .apiKey(config.getApiKey())
-                .restClientBuilder(
-                        RestClient.builder()
-                                .requestFactory(factory)
-                                .defaultHeader("Accept-Encoding", "gzip, deflate"))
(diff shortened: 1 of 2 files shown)


v1r3n (approved):

martinconstam:
@v1r3n Thanks for the earlier approval! I've rebased this onto current `main` (it had drifted behind the AI OkHttp3 transport refactors) and fixed the one post-rebase compile issue in the test — the `ChatCompletion` package move from `ai.models` to `ai.model`. CI is green now.

This is still needed: the manual `Accept-Encoding: gzip, deflate` header on the Mistral provider is unchanged on `main` and still disables OkHttp's transparent gzip decompression, so gzipped responses reach Jackson undecoded (`Illegal character ((CTRL-CHAR, code 31))`). Would you be able to merge when you get a chance? Thanks!

PR #1204, opened 2026-06-24
Avoid starving system task workers

v1r3n:
### queue semaphores

Replace scheduled-executor polling with a tight loop per queue: poll immediately when tasks are available, sleep only on empty poll. All non-isolated queues share one thread pool so threads aren't wasted on idle task types, while each queue gets its own semaphore so one busy type cannot starve others.

Adds `conductor.app.taskWorkerConfigs.<TaskType>.{threadCount,permitCount}` for per-type overrides: a non-zero threadCount gives the type a dedicated pool; permitCount caps in-flight concurrency independently of thread count.

Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Issue #

Alternatives considered
----


shaileshpadave (approved):

manan164:
Default `taskWorkerConfigs` is empty, so all ~24 async types (incl. the 16 AI tasks) share one pool, each with permits == pool size. A batch of slow `LLM_TEXT_COMPLETE` calls can hold every shared thread — a 5ms JOIN then waits a full LLM-call duration in the FIFO.

Suggestions:
1. Ship defaults for known-slow types (`HTTP`, `LLM_*`, `GENERATE_*`, `CALL_MCP_TOOL`) — via `application.properties`, or let tasks self-declare (`@WorkerTask.threadCount()` already exists but is ignored server-side).
2. Default per-type permits to a fraction of the pool (e.g. ¾) so no single type can monopolize it.
3. At minimum, doc note: any type holding threads for seconds needs a `threadCount` entry.

core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorker.java
```diff
@@ -73,113 +85,285 @@ public SystemTaskWorker(
...
+        if (config != null) {
+            return config;
+        }
+        for (Map.Entry<String, ConductorProperties.TaskWorkerConfig> entry : configs.entrySet()) {
+            if (entry.getKey().equalsIgnoreCase(taskType)) {
+                return entry.getValue();
+            }
+        }
+        return null;
     }
```
NicholasDCole, line 365: **Consider extracting the config-resolution logic into its own collaborator (SRP).**

`SystemTaskWorker` now owns a lot of unrelated concerns: the poll loop, dispatch, backpressure, lifecycle draining, thread-pool construction — and, on top of that, the whole "given a queue name, what `ExecutionConfig` should it get?" decision tree (`createNonIsolatedExecutionConfig`, `createIsolatedExecutionConfig`, `findTaskWorkerConfig`, `warnIfOversubscribed`). That cluster is a cohesive, self-contained responsibility: `taskType -> ExecutionConfig`. It reads `properties.getTaskWorkerConfigs()`, does the case-insensitive fallback, applies the oversubscription warning, and picks dedicated-vs-shared pool — none of which needs the poll loop, the queue DAO, or the executor lifecycle.

Suggest pulling it into an `ExecutionConfigResolver` (or `TaskWorkerConfigFactory`) that takes `ConductorProperties` + the shared pool and exposes `ExecutionConfig resolve(String taskQueue)`. `getExecutionConfig` then becomes just the `computeIfAbsent` cache around a call to the resolver.

Main payoff is testability: the resolution rules (dedicated vs shared, permit defaulting, case-insensitive key match, oversubscription warning) can be unit-tested directly against the resolver, instead of constructing a full `SystemTaskWorker` with four mocks and reaching in via `getExecutionConfig` the way the current tests do. It also keeps the worker focused on poll/dispatch/lifecycle.

Not blocking — the logic is correct as written — but this is the natural seam if you want to keep the class from growing further.

core/src/main/java/com/netflix/conductor/core/execution/tasks/ExecutionConfig.java
```diff
@@ -24,14 +24,44 @@ class ExecutionConfig {
...
+        this.executorService = executorService;
+        this.semaphoreUtil = semaphoreUtil;
+        this.poolSize = -1;
+    }
+
+    private ExecutionConfig(ExecutorService executorService, int permits, int poolSize) {
+        this.executorService = executorService;
+        this.semaphoreUtil = new SemaphoreUtil(permits);
+        this.poolSize = poolSize;
+    }
```
NicholasDCole, line 54: **The overloaded constructors are ambiguous about what the `int` means — suggest named static factories.**

There are now five constructors, and the numeric argument means different things depending on the overload:

| Constructor | What the `int` means |
|---|---|
| `(int threadCount, String)` | threadCount, and it's *also* used as permits |
| `(int threadCount, String, int permits)` | threadCount + separate permits |
| `(ExecutorService, int permits)` | permits (not threads) |
| `(ExecutorService, SemaphoreUtil)` | — (test-only) |
| private `(ExecutorService, int, int)` | permits + poolSize |

The trap is that `(int, String)` treats the number as thread count while `(ExecutorService, int)` treats it as permits — same-looking call site, opposite meaning. A caller can pick the wrong overload and get silently different concurrency behavior.

Suggest replacing the public constructors with intention-revealing static factories and making the constructors private, e.g.:

```java
static ExecutionConfig isolated(int threadCount, String nameFormat)               // own pool, permits == threadCount
static ExecutionConfig dedicated(int threadCount, String nameFormat, int permits)  // own pool, explicit permits
static ExecutionConfig shared(ExecutorService sharedPool, int permits)             // shared pool, own semaphore
```

This also lets you drop the `poolSize = -1` sentinel: `shared(...)` is the only one that uses the shared pool, so `getPoolSize()` semantics ("dedicated size, or -1 if shared") become explicit at the call site rather than encoded in a magic number. The three call sites in `SystemTaskWorker` (`createIsolatedExecutionConfig`, the two branches of `createNonIsolatedExecutionConfig`) map one-to-one onto these three factories.

core/src/main/java/com/netflix/conductor/core/execution/tasks/ExecutionConfig.java
```diff
@@ -24,14 +24,44 @@ class ExecutionConfig {
...
+     * Non-isolated queues: share the given pool but each gets its own semaphore, so a slow/busy
+     * queue cannot exhaust a shared permit pool and starve other queues' polling.
+     */
+    ExecutionConfig(ExecutorService executorService, int permits) {
+        this(executorService, permits, -1);
+    }
 
-        this.semaphoreUtil = new SemaphoreUtil(threadCount);
+    /** Test-only: inject a pre-built semaphore (e.g. a mock). */
+    ExecutionConfig(ExecutorService executorService, SemaphoreUtil semaphoreUtil) {
```
NicholasDCole, line 44: **This constructor is test-only — mark it or move the seam.**

`ExecutionConfig(ExecutorService, SemaphoreUtil)` exists solely so tests can inject a mock/pre-built `SemaphoreUtil` (e.g. `testPermitAcquireFailureRequeuesPolledTasks`, `testPollAndExecuteLoopSurvivesUncaughtException`). It's part of the production constructor surface but never used in production, and it isn't annotated, so nothing signals that to a future caller — someone could pick it up and get a config with `poolSize = -1` and an externally-owned semaphore without realizing it's a test hatch.

Minimal fix: annotate it `@VisibleForTesting` (consistent with how the rest of this codebase marks test-only seams) and add a one-line comment. Better, if you adopt the static-factory refactor from the constructor comment: keep the factories as the only production entry points and expose this injection path as an explicitly-named `@VisibleForTesting static ExecutionConfig forTesting(ExecutorService, SemaphoreUtil)` so the test intent is unambiguous and it can't be mistaken for a real construction path.

ai/src/main/java/org/conductoross/conductor/ai/LLMHelper.java
```diff
@@ -674,10 +674,11 @@ String extractMethodFromInputParameters(Map<String, Object> inputParameters) {
...
         List<Media> media =
-                (rawMedia == null ? List.<String>of() : rawMedia).stream()
-                        .map(m -> getMedia(msg.getMimeType(), m))
-                        .filter(Objects::nonNull)
-                        .toList();
+                (rawMedia == null ? List.<String>of() : rawMedia)
+                        .stream()
+                                .map(m -> getMedia(msg.getMimeType(), m))
+                                .filter(Objects::nonNull)
+                                .toList();
```
NicholasDCole, line 681: Can we undo this change? This seems like just whitespacing

NicholasDCole (approved):

core/src/main/java/com/netflix/conductor/core/execution/tasks/ExecutionConfig.java
v1r3n: This is by design and to avoid making excessive changes.

ai/src/main/java/org/conductoross/conductor/ai/LLMHelper.java
v1r3n: Yes it broke spotless rules.

v1r3n: so required

NicholasDCole (commented):
I agree with @manan164 that a default thread pool for LLM requests makes sense
covering `LLM_*`

mp-orkes (approved):
LGTM but I would recommend reviewing #1369 first and if it makes sense merging that one first.

Why?

- There are going to be merge conflicts.
- #1369 addresses a correctness issue (a RC that generates duplicates). This PR is about throughput (it does matter but IMO correctness comes first)
- Without #1369 you will still have LLM_CHAT_COMPLETE, HTTP and other tasks that are long running and end up consuming N threads (for the N duplicate executions on sweeper runs)
- With #1369 the fact that there are no more duplicates mitigates the thread pool starvation.

PR #1212, opened 2026-06-24
fix(core): reset nested JOIN tasks transitively after sub-workflow restart

manan164:
## Summary

Three test flakiness sources fixed, all rooted in PR #973's async SUB_WORKFLOW launch:

**1. `NestedForkJoinSubWorkflowSpec` (core fix)**
`resetUnsuccessfulJoinTasksWithActiveBranches()` computed `activeReferenceTaskNames` as a one-shot snapshot. In nested fork-join, resetting `inner_join` to IN_PROGRESS in the same pass never made it visible to `outer_join`'s dependency check — `outer_join` stayed CANCELED → `DeciderService.retry()` → `TerminateWorkflowException(TERMINATED)`.
Fix: iterate until convergence — each reset JOIN is added to the active set, so outer JOINs are caught in the next pass.

**2. `SubWorkflowRestartSpec` (test fix)**
`queueDAO.pop()` can return empty when the background sweeper already consumed the task; Groovy `null[0]` → `asyncSystemTaskExecutor.execute(null)` → no-op → `subWorkflowId` stays null → `sweep(null)` → Jedis `zrem(null)` error.
Fix: guard the pop with an if-check and use `conditions.eventually` to wait for the ID.

**3. `WorkflowRerunTests` (e2e fix)**
10-second await for `subWorkflowId` attachment times out under CI load.
Fix: increase to 30 seconds.

## Test plan
- [ ] `NestedForkJoinSubWorkflowSpec` — all tests pass (run 10× in CI)
- [ ] `SubWorkflowRestartSpec` — no regressions
- [ ] `WorkflowRerunTests` — no timeout failures
- [ ] `SubWorkflowRetrySpec`, `SubWorkflowRerunSpec`, `HierarchicalForkJoin*` specs — no regressions
- [ ] `conductor-core` unit tests pass


test-harness/src/test/groovy/com/netflix/conductor/test/integration/NestedForkJoinSubWorkflowSpec.groovy
```diff
@@ -188,6 +188,7 @@ class NestedForkJoinSubWorkflowSpec extends AbstractSpecification {
     def "test restart on the sub workflow in a nested fork join workflow"() {
         when:
         workflowExecutor.restart(subworkflowId, false)
+        sweep(parentWorkflowId)
```
manan164, line 190: There seems to be race condition issue because of sweeper. Its better to sweep the parent workflow and then assert on the status.

test-harness/src/test/groovy/com/netflix/conductor/test/integration/SubWorkflowRerunSpec.groovy
```diff
@@ -88,8 +84,11 @@ class SubWorkflowRerunSpec extends AbstractSpecification {
         verifyPolledAndAcknowledgedTask(pollAndCompleteTask)
 
         when: "the subworkflow task should be in SCHEDULED state and is started by issuing a system task call"
-        List<String> polledTaskIds = queueDAO.pop(TASK_TYPE_SUB_WORKFLOW, 1, 200)
-        asyncSystemTaskExecutor.execute(subWorkflowTask, polledTaskIds[0])
+        def rootSubWfTask = workflowExecutionService.getExecutionStatus(rootWorkflowId, true)
```
manan164, line 87: Again race condition with sweeper. If the sweeper already has polled but not yet schedule then schedule it now.

v1r3n (approved):

PR #1214, opened 2026-06-24
Demo: deterministic integration test-harness (background sweeper off) — sweeper-off counterpart to #1212

nthmost-orkes:
## What this is

A **demonstration** of running the integration test-harness with the **background workflow sweeper disabled**, so progress is driven only by explicit `sweep()` calls and intermediate-state assertions become fully deterministic. This is the sweeper-**off** counterpart to #1212 (the sweeper-**on** approach), which has since merged to `main`. Put up for **discussion — not necessarily to merge**.

Rebased on current `main`, so it sits on top of #1212's product fix and sweeper-on spec fixes. Those turned out to be compatible with the sweeper-off harness as-is (they all advance via `sweep()`), so none of them needed changing.

## What changes

- **`AbstractSpecification`** — disable the background sweeper via the existing `conductor.app.sweeper.enabled=false` flag (**no production change**). With the `WorkflowSweeper` bean absent, the `sweep()` helper drives `WorkflowExecutor.decide()` directly — the same core call the sweeper's `pollAndSweep` loop makes. Manual `sweep()` becomes the only driver of deciding → deterministic.
- **`HierarchicalForkJoinSubworkflowRetrySpec`** — replace the `await()`-for-the-background-sweeper blocks with explicit `sweep()` calls (drop now-unused awaitility/TimeUnit imports).
- **`InlineSubWorkflowFromExpressionSpec`** — drive the DO_WHILE sub-workflow forward with explicit `sweep()` inside the polling blocks.
- **`WorkflowTestUtil.pollForTask`** — return `null` on timeout instead of throwing, so the handful of legacy best-effort setup polls (that the background sweeper used to satisfy) are no-ops again. Callers that actually need the task still assert on it via `verifyPolledAndAcknowledgedTask`, so this does not mask real failures.
- **`WorkflowSweeper` / `WorkflowSweeperTest`** — best-effort try/catch on the NotFound-recovery queue cleanup so a queue hiccup there can't escape `sweep()`. (This PR's original production hardening; orthogonal to the harness demo, kept.)

## Honest framing

Determinism here comes from **controlling `decide()` via explicit `sweep()`** — no background thread racing the assertions — *not* from the queue poll, which stays best-effort for task-queue visibility exactly as it already was via `await`. This trades production-fidelity (no live sweeper) for deterministic, strict, fast assertions — the opposite trade from #1212. The PR exists to make that trade concrete so the two approaches can be compared side by side.

## Validation

Full `:conductor-test-harness:test` suite green — **251 run, 0 failed, 31 skipped** — run repeatedly (locally, via testcontainers). No new tolerance bands on assertions.

## Orkes parity

Test-harness + the OSS-side `org.conductoross` `WorkflowSweeper` only. No Orkes parallel to mirror.

### DIFF
--- a/core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
+++ b/core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
@@ -292,7 +292,19 @@ public class WorkflowSweeper extends LifecycleAwareComponent {
             }
         } catch (NotFoundException nfe) {
             log.error("Error running sweep for {}, error = {}", workflowId, nfe.getMessage(), nfe);
-            queueDAO.remove(DECIDER_QUEUE, workflowId);
+            // Best-effort cleanup: the workflow is gone, so drop it from the decider queue.
+            // This must not throw — a queue-layer failure here (e.g. removing a message that
+            // was never enqueued for a just-restarted workflow) would otherwise escape sweep()
+            // and crash the sweeper thread.
+            try {
+                queueDAO.remove(DECIDER_QUEUE, workflowId);
+            } catch (Exception removeEx) {
+                log.warn(
+                        "Failed to remove {} from decider queue during NotFound recovery, error = {}",
+                        workflowId,
+                        removeEx.getMessage(),
+                        removeEx);
+            }
         } catch (Throwable e) {
             log.error("Error running sweep for {}, error = {}", workflowId, e.getMessage(), e);
         } finally {
--- a/core/src/test/java/org/conductoross/conductor/core/execution/WorkflowSweeperTest.java
+++ b/core/src/test/java/org/conductoross/conductor/core/execution/WorkflowSweeperTest.java
@@ -22,6 +22,7 @@ import org.junit.Test;
 
 import com.netflix.conductor.common.metadata.tasks.TaskType;
 import com.netflix.conductor.core.config.ConductorProperties;
+import com.netflix.conductor.core.exception.NotFoundException;
 import com.netflix.conductor.core.execution.WorkflowExecutor;
 import com.netflix.conductor.core.execution.tasks.SystemTaskRegistry;
 import com.netflix.conductor.core.execution.tasks.WorkflowSystemTask;
@@ -35,6 +36,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -183,6 +186,34 @@ public class WorkflowSweeperTest {
         verify(executionDAO).updateTask(subWorkflowTask);
         verify(workflowExecutor, times(2)).decide(WORKFLOW_ID);
         verify(queueDAO, never()).push(anyString(), anyString(), anyLong());
+        verify(executionLockService).releaseLock(WORKFLOW_ID);
+    }
+
+    @Test
+    public void sweepDoesNotPropagateWhenDeciderQueueCleanupFailsAfterNotFound() {
+        WorkflowModel workflow =
+                newWorkflow(
+                        List.of(
+                                newTask(
+                                        "simple-task",
+                                        TaskType.TASK_TYPE_SIMPLE,
+                                        TaskModel.Status.IN_PROGRESS)));
+
+        when(executionLockService.acquireLock(WORKFLOW_ID)).thenReturn(true);
+        when(workflowExecutor.getWorkflow(WORKFLOW_ID, true)).thenReturn(workflow);
+        // decide signals the workflow is gone (e.g. it raced a restart/terminate)
+        when(workflowExecutor.decide(WORKFLOW_ID))
+                .thenThrow(new NotFoundException("No such workflow found by id: " + WORKFLOW_ID));
+        // Simulate the recovery cleanup failing for ANY reason. This guards the catch
+        // block's behavior; it does not assert a specific queue-layer cause.
+        doThrow(new RuntimeException("queue cleanup failed"))
+                .when(queueDAO)
+                .remove(anyString(), anyString());
+
+        // must not propagate — a failed best-effort cleanup cannot be allowed to crash the sweeper
+        workflowSweeper.sweep(WORKFLOW_ID);
+
+        verify(queueDAO).remove(anyString(), eq(WORKFLOW_ID));
         verify(executionLockService).releaseLock(WORKFLOW_ID);
     }
 
--- a/test-harness/src/test/groovy/com/netflix/conductor/test/base/AbstractSpecification.groovy
+++ b/test-harness/src/test/groovy/com/netflix/conductor/test/base/AbstractSpecification.groovy
@@ -12,7 +12,6 @@
  */
 package com.netflix.conductor.test.base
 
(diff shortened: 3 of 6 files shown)


nthmost-orkes:
## CI just reproduced the race for us — and it confirms the diagnosis exactly

The `test-harness` check failed, and the failure mode is the strongest evidence yet that this is a harness race, not a sweeper bug.

**What failed** ([run 28137348270](https://github.com/conductor-oss/conductor/actions/runs/28137348270/job/83326970304)):

```
SubWorkflowRestartSpec > Test restart on the root in a 3-level subworkflow FAILED
  SubWorkflowRestartSpec.groovy:234 | NotFoundException: No such workflow found by id: null
251 tests run, 219 passed, 31 skipped, 1 failed
```

Two things changed versus the original failure, and both matter:

**1. The `IllegalArgumentException: "null is not a valid argument"` is gone.** The `sweep()` null-guard in this PR worked exactly as designed — `sweep(null)` now returns cleanly instead of crashing through `zrem(key, null)`. That's direct confirmation the sweep(null) → Jedis diagnosis was correct.

**2. The same null resurfaced one line later, *outside the sweeper entirely*.** The spec itself dereferences the null id:

```groovy
def newLeafWorkflowId = ...getTasks().get(1).subWorkflowId   // line 229 — NULL (the race)
sweep(newLeafWorkflowId)                                     // line 230 — now guarded, returns cleanly ✓
...
with(workflowExecutionService.getExecutionStatus(newLeafWorkflowId, true)) {  // line 234 — getExecutionStatus(null) → NotFoundException
```

`subWorkflowId` came back **null** again (the async `SUB_WORKFLOW` task lost the race), and the spec fed that null straight into `getExecutionStatus(null, true)` at line 234, which throws `NotFoundException` directly in the test — nothing to do with the sweeper.

## Bottom line

- **The race is real and just reproduced in CI** — `tasks[1].subWorkflowId` is intermittently null after restart.
- **It is unambiguously a harness bug, not a sweeper bug.** The sweeper guards in this PR are correct, but they can't make this spec pass on their own, because the spec reads and dereferences the same null id independently (line 234). The failure simply moved from the sweeper to the spec's next use of the null.
- **The only deterministic fix is in the spec:** wait for a non-null `subWorkflowId` before using it (both `sweep(...)` and `getExecutionStatus(...)`), in all three restart cases (root/mid/leaf).

## Proposed next step

The deterministic fix is to poll until `getTasks().get(1).subWorkflowId` is populated before lines 230/234 (e.g. `Awaitility` or a wait helper), in each of the three test cases. That touches a known-flaky hierarchical-subworkflow harness spec, so it should be coordinated rather than slipped in silently — happy to add it to this PR or split it out, whichever maintainers prefer.

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -159,6 +159,12 @@ public CompletableFuture<Void> sweepAsync(String workflowId) {
     }
 
     public void sweep(String workflowId) {
+        if (StringUtils.isBlank(workflowId)) {
```
manan164: Why would workflowId comes blank here?

nthmost-orkes: Good question, it shouldn't.

In production `sweep()` is only ever called from `pollAndSweep()`, where `workflowId` comes from `queueDAO.pop(DECIDER_QUEUE, ...)`. Those are Redis zset members, which are never null/blank, and that loop already has its own `catch(Throwable)`. So this branch is effectively dead code in prod.

Where it *does* go blank is the **test-harness**, which calls `workflowSweeper.sweep(...)` directly with a value read straight off a task:

```groovy
def newLeafWorkflowId = ...getTasks().get(1).subWorkflowId   // intermittently null (async race)
sweep(newLeafWorkflowId)
```

`subWorkflowId` is populated asynchronously by the `SUB_WORKFLOW` system task, so when the spec reads it too early it's `null` and `sweep(null)` is invoked. That's the failure this PR was chasing.

It's true we should remove this guard, i just need to figure out how not to generate `null` in the test

nthmost-orkes: You're right — I dug into it and **production can never reach this with a blank/null id**, so the entry guard was dead code. Removed it in 87bc796a8.

For the record, here's the full verification:
- No production code calls `sweep()`/`sweepAsync()` directly — only `pollAndSweep()`, which sweeps non-null Redis decider-queue members and already has its own `catch(Throwable)`.
- `StartWorkflowRequest` has no `workflowId` field, so external callers can't inject one. (`startWorkflow` does `Optional.ofNullable(input.getWorkflowId()).orElseGet(idGenerator::generate)`, which *would* let a caller-supplied empty string survive — but there's no public field to supply it.)
- Every internal `setWorkflowId(...)` in the start path uses a generated id (`SubWorkflow` → `generateSubWorkflowId`, failure-workflow path → `idGenerator.generate()`).

The only way `sweep()` ever saw a null was the test-harness calling it directly with a `subWorkflowId` read before the async `SUB_WORKFLOW` task populated it. I've kept the best-effort try/catch on the NotFound-recovery `remove` (general robustness), and the actual fix for the flake lives in the spec — see the plan in the PR thread.

nthmost-orkes:
## Two directions for the real fix — need to pick a lane

Dug into the fork/join flakes (repro'd the background-sweeper race deterministically, measured the JOIN states). It nets out to two philosophies — both validated locally (full test-harness green, repeatedly), both including #1212's product fix:

**A — sweeper off in the harness.** Specs drive every decide manually; strict assertions, no tolerance bands; fully deterministic. Downside: it removes a core production component and makes `sweep()` call `decide()` directly — so the integration tests stop exercising the real sweeper/concurrency, and likely wouldn't have caught the very bug #1212 fixes.

**B — sweeper on (#1212's direction).** Take #1212's transitive nested-JOIN-reset fix, and make the racy specs `await()` the *converged* state instead of asserting an exact pre-sweep transient. Stays production-like; the flakiness goes away because we stop asserting states the sweeper is mid-flight on. Trade: assertions become "eventual" rather than "exact step," plus the small stuff (blind-pop guard, poll-retry, a couple of `await` timeout bumps).

**My vote: B.** These are integration tests — fidelity to the running engine is the point, and A buys determinism by making the tests less like prod. A was worth building (it proved the specs *can* be made deterministic and mapped exactly which ones lean on the sweeper), but I'd ship B.

mp-orkes:
> ## Two directions for the real fix — need to pick a lane
> Dug into the fork/join flakes (repro'd the background-sweeper race deterministically, measured the JOIN states). It nets out to two philosophies — both validated locally (full test-harness green, repeatedly), both including #1212's product fix:
> 
> **A — sweeper off in the harness.** Specs drive every decide manually; strict assertions, no tolerance bands; fully deterministic. Downside: it removes a core production component and makes `sweep()` call `decide()` directly — so the integration tests stop exercising the real sweeper/concurrency, and likely wouldn't have caught the very bug #1212 fixes.
> 
> **B — sweeper on (#1212's direction).** Take #1212's transitive nested-JOIN-reset fix, and make the racy specs `await()` the _converged_ state instead of asserting an exact pre-sweep transient. Stays production-like; the flakiness goes away because we stop asserting states the sweeper is mid-flight on. Trade: assertions become "eventual" rather than "exact step," plus the small stuff (blind-pop guard, poll-retry, a couple of `await` timeout bumps).
> 
> **My vote: B.** These are integration tests — fidelity to the running engine is the point, and A buys determinism by making the tests less like prod. A was worth building (it proved the specs _can_ be made deterministic and mapped exactly which ones lean on the sweeper), but I'd ship B.

+1 on B

Keep the sweeper on. Assert on the eventual state.

nthmost-orkes:
Closing as we learned what we needed.

PR #1230, opened 2026-06-30
chore(github): add wrong-repo signals to issue templates and AGENTS.md

nthmost-orkes:
## Summary

Issues are regularly filed in this repo by users of unrelated products that share the Conductor name — likely including AI agents acting on their behalf, which would explain why the reports are well-formatted and detailed despite being in the wrong place.

This PR adds friction for those cases:

- **`bug_report.md` and `bug_report.yml`**: `[!WARNING]` callout listing concrete mismatch signals (OpenCode, `com.conductor.app` paths, `settings.toml`, desktop coding UI errors) so users can self-identify before filing
- **`AGENTS.md`**: "Repository Identity" section with the same signal list, written for LLM agents — positive signals (what this repo IS) followed by an explicit abort list for wrong-repo cases
- **`.github/wrong-repo-response.md`**: boilerplate close comment for manual triage

## What this does NOT do

This PR does not name or link to any other product. The templates assert what *this* repo is and flag the mismatch signals — anyone who is in the wrong place can find where they belong on their own.


nthmost-orkes:
Closing — naming a third-party product explicitly in our issue templates and AGENTS.md isn't the right approach. The templates should assert what *this* repo is without linking to or naming any other product. Will revise the triage doc to strip the explicit naming too.

PR #1235, opened 2026-06-30
feat(ui): generic task execution-panel plugin slot

vishesh-orkes:
## Summary

Adds a **generic plugin extension point for task execution panels** — extra tabs in the execution view's task detail (right) panel, contributed by plugins for specific task types. This is reusable infrastructure (not feature-specific); the enterprise UI uses it to add an LLM guardrail-executions viewer without modifying the core execution page.

## Changes
- `plugins/registry/types.ts` — `TaskExecutionPanelRegistration` (`id`, `label`, `taskTypes`, `component`) + `TaskExecutionPanelProps`; `ConductorPlugin.taskExecutionPanels?`; `PluginRegistry.getTaskExecutionPanels(taskType)`.
- `plugins/registry/registry.ts` — `getTaskExecutionPanels(taskType)` returns panels whose `taskTypes` include the given type.
- `pages/execution/state/constants.ts` — `PLUGIN_PANEL_TAB_BASE` (tab index base for plugin panels, above the built-ins).
- `pages/execution/RightPanel/RightPanel.tsx` — renders registered panels as extra tabs + their content for the selected task.

No behavior change for OSS builds (no panels registered → no extra tabs).

## Why
Mirrors the existing `dependencySections` extension pattern. Enterprise plugins (e.g. guardrails) register a panel via `taskExecutionPanels` instead of patching the execution page.


ui-next/src/pages/execution/state/constants.ts
```diff
@@ -5,3 +5,5 @@ export const OUTPUT_TAB = 2;
 export const LOGS_TAB = 3;
 export const JSON_TAB = 4;
 export const DEFINITION_TAB = 5;
+// Base tab index for plugin-registered task execution panels (well above built-ins).
+export const PLUGIN_PANEL_TAB_BASE = 100;
```
ajlai-orkes, line 9: is this to ensure it's always the last tab?

ajlai-orkes (approved):

ui-next/src/pages/execution/state/constants.ts
vishesh-orkes: Not exactly — the tabs render last because they're appended after the Definition tab in the JSX. PLUGIN_PANEL_TAB_BASE = 100 is about tab identity, not order: MUI matches the active tab by value, and the built-in tabs auto-index to 0–5, so plugin panels use 100 + i to guarantee their values never collide with the built-ins (and stay stable even if a built-in tab is added/removed or conditionally hidden).

PR #1237, opened 2026-07-01
fix: response timeout never fires for signal-wait WAIT tasks

### LINKED ISSUE #1146: WAIT for a signal task does not timeout with a task definition timeouts set
### Describe the bug

A WAIT task set to wait for a signal does not timeout with the timeout values set in a task definition.

Here is a task definition:

`{
  "createTime": 1780331441794,
  "updateTime": 1780333894546,
  "createdBy": "",
  "updatedBy": "",
  "name": "WAIT_1_MINUTE",
  "description": "Wait for a signal or timeout after 1 minute",
  "retryCount": 3,
  "timeoutSeconds": 60,
  "inputKeys": [],
  "outputKeys": [],
  "timeoutPolicy": "TIME_OUT_WF",
  "retryLogic": "FIXED",
  "retryDelaySeconds": 60,
  "responseTimeoutSeconds": 60,
  "inputTemplate": {},
  "rateLimitPerFrequency": 0,
  "rateLimitFrequencyInSeconds": 1,
  "ownerEmail": "",
  "backoffScaleFactor": 1,
  "totalTimeoutSeconds": 60,
  "enforceSchema": false
}`

Here is the task itself:
{
      "name": "WAIT_1_MINUTE",
      "taskReferenceName": "waitForSignal",
      "inputParameters": {},
      "type": "WAIT",
      "decisionCases": {},
      "defaultCase": [],
      "forkTasks": [],
      "startDelay": 0,
      "joinOn": [],
      "optional": false,
      "asyncComplete": false,
      "loopOver": [],
      "onStateChange": {},
      "permissive": false
    }

The task stays in IN_PROGRESS state forever.

Thanks.

nthmost-orkes:
## Root cause

`WaitTaskMapper.setCallbackAfter()` sets `task.callbackAfterSeconds = Integer.MAX_VALUE` for WAIT tasks that have no `duration`/`until` input (signal-wait mode). This sentinel tells `addTaskToQueue` to push the task to the system-task queue with a ~68-year delivery offset so `AsyncSystemTaskExecutor` never processes it — which is the desired behaviour for a task that waits for an external signal.

The problem is that `DeciderService.isResponseTimedOut` also uses `callbackAfterSeconds` when computing the tolerance window:

```java
long callbackTime = 1000L * task.getCallbackAfterSeconds();   // ~68 years in ms
long adjustedResponseTimeout = responseTimeout + callbackTime; // ~68 years
long noResponseTime = now - task.getUpdateTime();              // seconds/minutes
// noResponseTime < adjustedResponseTimeout → always true → timeout never fires
```

Any `responseTimeoutSeconds` configured on the task definition is therefore silently ignored for all signal-wait WAIT tasks.

## Fix

Cap `callbackAfterSeconds` at `Integer.MAX_VALUE` before converting to milliseconds inside `isResponseTimedOut`. The sentinel value means "no executor callback is expected" — the task is waiting for an external signal — so the callback-delay contribution to the response-timeout window should be zero, not 68 years.

```java
long callbackAfterSecs = task.getCallbackAfterSeconds();
long callbackTime =
        (callbackAfterSecs >= Integer.MAX_VALUE) ? 0 : 1000L * callbackAfterSecs;
```

`checkTaskTimeout` (wall-clock elapsed since `startTime`) and `checkTotalTimeout` (elapsed since `firstScheduledTime`) are unaffected — they don't use `callbackAfterSeconds` — and continue to fire via the sweeper's `decide()` cycle.

## User impact

First-time users and anyone configuring a WAIT task with `responseTimeoutSeconds` to act as a deadline would find the timeout silently ignored: the WAIT task stayed IN_PROGRESS long past the configured limit. After this fix, `responseTimeoutSeconds` is enforced correctly.

## Test

`TestDeciderService.testIsResponseTimedOut_signalWaitWithMaxCallbackAfterSeconds` — two assertions: timeout fires when elapsed > configured limit, does not fire when still within the window.

## Orkes parity

No Orkes parallel — this is a bug fix to an existing OSS-side calculation; no new classes or interfaces introduced.

## OSS-side additions

None.

Fixes conductor-oss/conductor#1146

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/DeciderService.java
@@ -909,7 +909,12 @@ public class DeciderService {
 
         // calculate pendingTime
         long now = System.currentTimeMillis();
-        long callbackTime = 1000L * task.getCallbackAfterSeconds();
+        long callbackAfterSecs = task.getCallbackAfterSeconds();
+        // Integer.MAX_VALUE is WaitTaskMapper's sentinel for "wait indefinitely for an external
+        // signal" (no duration/until). Adding ~68 years to the timeout window makes
+        // isResponseTimedOut permanently return false — treat it as zero callback delay instead.
+        long callbackTime =
+                (callbackAfterSecs >= Integer.MAX_VALUE) ? 0 : 1000L * callbackAfterSecs;
         long referenceTime =
                 task.getUpdateTime() > 0 ? task.getUpdateTime() : task.getScheduledTime();
         long pendingTime = now - (referenceTime + callbackTime);
--- a/core/src/test/java/com/netflix/conductor/core/execution/TestDeciderService.java
+++ b/core/src/test/java/com/netflix/conductor/core/execution/TestDeciderService.java
@@ -1548,6 +1548,31 @@ public class TestDeciderService {
     }
 
     @Test
+    public void testIsResponseTimedOut_signalWaitWithMaxCallbackAfterSeconds() {
+        // Signal-wait WAIT tasks (no duration/until) use Integer.MAX_VALUE as a sentinel meaning
+        // "never re-queue to AsyncSystemTaskExecutor". Before the fix, this made
+        // adjustedResponseTimeout ~68 years, permanently disabling response-timeout enforcement.
+        TaskDef taskDef = new TaskDef();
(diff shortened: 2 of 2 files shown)


nthmost-orkes:
Closing per Viren's feedback — the current behavior is WAI. Will discuss before proposing any changes to WAIT task timeout semantics.

Issue #1152, opened 2026-06-06
EPIC: Environment Variables (and Secrets) for Conductor OSS


nthmost-orkes:
---

## Overview

Bring Orkes' **environment variables** feature to Conductor OSS at full feature parity for the non-enterprise surface, plus a companion **secrets** capability built on the same resolution path. Today an OSS workflow author can reference OS/JVM env vars (`${ENV_VAR}`) and runtime workflow variables (`${workflow.variables.X}`), but there is **no way to define a named, server-stored environment variable** and reference it as `${workflow.env.MY_VAR}` the way Orkes supports — even though the SDKs expose `/api/environment` methods that 404 against an OSS server.

This Epic closes that gap, removes the multi-tenant / RBAC / cloud-provider machinery that has no place in OSS, and documents the feature so the long-standing "I thought I could do this in OSS" confusion ends.

## The confusion this fixes

"Environment variable" means **three different things** in the Conductor world, and the docs never disambiguate them. This is the root cause of the scattered issues:

| Meaning | How it's set | Example | Status in OSS |
|---|---|---|---|
| **1. Server config env vars** | OS/JVM env or `config.properties` | `CONDUCTOR_OPTS=-DconductOR...`, `conductor.redis.hosts` | Works (#547, #353) |
| **2. SDK client config env vars** | OS env read by the SDK `ApiClient` | `CONDUCTOR_SERVER_URL`, `CONDUCTOR_AUTH_KEY` | Works (#353, #514) |
| **3. Workflow env-var metadata** | Server-stored, CRUD via REST, referenced in task inputs | `${workflow.env.DB_URL}` | **Missing — this Epic** |

Meanings 1 and 2 are config plumbing and already work. Meaning 3 is the Orkes feature people actually ask for when they say "set environment variables in Conductor," and it does not exist in OSS. The docs deliverable below makes this distinction explicit so users stop conflating them.

## Approach

[...]

PR #1251, opened 2026-07-05
Secrets & environment: workflow references (${workflow.secrets/env}) + task-declared runtimeMetadata

mp-orkes:
This PR delivers **two related features** built on the same env-backed secret/environment providers (`SecretsDAO` / `EnvironmentDAO`). Feature 2 (originally #1255, squash-merged into this branch) **reuses** Feature 1's DAOs.

---

# Feature 1 — Workflow-level secret/env references (`${workflow.secrets.*}` / `${workflow.env.*}`)

## What
Adds two managed reference namespaces to workflow task inputs, backed by process environment variables behind pluggable DAO interfaces:

- `${workflow.env.<NAME>}` — resolved **eagerly** from `CONDUCTOR_ENV_<NAME>`.
- `${workflow.secrets.<NAME>}` — resolved **deferred** from `CONDUCTOR_SECRET_<NAME>`.

Plus read-only discovery endpoints: `GET /api/environment` (name+value) and `GET /api/secrets` (names only — never values).

## Why
Workflow definitions have no first-class way to reference a secret or a named environment value. The only mechanism today is bare `${VAR}`, which resolves *any* process env var from any workflow. This adds a safe, operator-controlled prefixed channel, and a seam (`SecretsDAO` / `EnvironmentDAO`) for real backends later. Bare `${VAR}` behavior is unchanged.

## How it works (non-obvious)
- **Secrets never persist.** `${workflow.secrets.*}` is left literal in the stored task input and substituted only at hand-off (worker poll + async system-task execution), onto a fresh copy — so the value never reaches the datastore, UI, or logs. Env vars resolve eagerly (not treated as sensitive).
- **Prefix = safety boundary.** Only vars an operator named under `CONDUCTOR_SECRET_` / `CONDUCTOR_ENV_` are reachable; `PATH`, credentials, etc. are not.
- **Providers** are selected by `conductor.secrets.type` / `conductor.environment.type` (`env` default, `noop` off-switch), gated by `@ConditionalOnProperty`; prefixes configurable via `conductor.{secrets,environment}.env.prefix`.

## How to test
Set env values (`CONDUCTOR_SECRET_FOO=bar`, or `-Dconductor.secret...`/`System.setProperty` in tests). Covered by unit tests (DAOs, `ParametersUtils`) and an end-to-end integration spec asserting the secret stays literal in stored input, env resolves eagerly, and the secret resolves on poll.

## Known limitations (intentional for this initial version)
- **No secret masking.** Resolved secret values are not masked in task/workflow **output**. `${workflow.secrets.X}` is kept literal in the *stored task input* and resolved only at execution hand-off, so the secret value is never persisted into the input. However, if a reference flows into a task's output (e.g. an `INLINE` task that returns it, or an HTTP task echoing a response), the resolved value will appear in the persisted/returned output. This is deliberate for this first version — output value masking is deferred to a follow-up.
- **External payload storage.** Secrets are not resolved for task inputs that were offloaded to external payload storage. Fails safe: the reference passes through unresolved and a warning is logged.
- **Malformed JSON sub-path.** `${workflow.secrets.NAME.path}` where the secret value is not valid JSON resolves to `null` with a warning.

Substitution covers worker-polled tasks, async system tasks (e.g. HTTP), and synchronous system tasks that run in the decide loop (INLINE, SET_VARIABLE, JSON_JQ_TRANSFORM, WAIT, HUMAN). End-to-end integration tests cover the worker-poll, HTTP, INLINE, SET_VARIABLE, JSON_JQ_TRANSFORM, and sub-workflow paths — each asserts the resolved secret is used at execution while the persisted task input stays literal.

---

# Feature 2 — Task-declared runtime metadata (`runtimeMetadata`)

Merged from #1255 (squash-merged into this branch). See that PR for full details.


core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -146,12 +152,29 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
...
+                Map<String, Object> literalInput = task.getInputData();
+                // Secrets substitution only sees task.getInputData(); when input has been
+                // offloaded to external payload storage, getInputData()/setInputData() operate on
+                // a different field and this substitution silently becomes a no-op.
+                if (task.getExternalInputPayloadStoragePath() != null) {
+                    LOGGER.warn(
+                            "Task {} has externalized input; ${{workflow.secrets.*}} references are not resolved for external payload storage",
+                            task.getTaskId());
+                }
+                task.setInputData(parametersUtils.substituteSecrets(literalInput));
```
mp-orkes, line 166: secret substitution

core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
```diff
@@ -1871,6 +1872,47 @@ boolean scheduleTask(WorkflowModel workflow, List<TaskModel> tasks) {
...
+     * <p>{@code ${workflow.secrets.X}} references are deferred at schedule time and are normally
+     * only resolved at task hand-off (worker poll or async system task execution). Synchronous
+     * system tasks execute inline in the decide loop and never go through those hand-off points, so
+     * without this substitution they would see the literal, unresolved secret reference.
+     */
+    private void startSyncSystemTaskWithSecrets(
+            WorkflowSystemTask systemTask, WorkflowModel workflow, TaskModel task) {
+        Map<String, Object> literalInput = task.getInputData();
+        try {
+            task.setInputData(parametersUtils.substituteSecrets(literalInput));
```
mp-orkes, line 1888: secret substitution

c4lm:
> ${workflow.secrets.X} is kept literal in the stored task input

Will this also be addressed in the follow up - do you think we should have parity with orkes-conductor here?
e.g. "SomeKeyerlyu": "<ins>***</ins>" in task **input** when fetching workflow with a task which relied on a secret

mp-orkes:
> > ${workflow.secrets.X} is kept literal in the stored task input
> 
> Will this also be addressed in the follow up - do you think we should have parity with orkes-conductor here? e.g. "SomeKeyerlyu": "***" in task **input** when fetching workflow with a task which relied on a secret

@c4lm 

This is a good question and definitely open to discussion.

With this implementation there is no secret masking. I want to do that as a follow up.

E.g.: You have a workflow that uses `${workflow.secrets.SECRET}` as input. You won't see the secret in the input. But, if for example, the inline task uses the secret value as output, there is no masking.

SECRET


Screenshot 2026-07-06 at 16 10 58

---
Orkes Conductor would mask both as "***".

Screenshot 2026-07-06 at 16 14 53

Screenshot 2026-07-06 at 16 15 01

c4lm:
I get that - I am asking whether you intend to preserve the placeholders vs *** when masking

mp-orkes:
> I get that - I am asking whether you intend to preserve the placeholders vs *** when masking

I want to preserve the placeholders. No masking in this PR.

If we want to mask things (as we do in Orkes Conductor) it's a follow-up PR.

mp-orkes:
### Relationship to #1152 (EPIC: Environment Variables and Secrets for Conductor OSS)

**Part of #1152** — this is an initial, **env-var-backed** phase; it does **not** implement the full EPIC, so it does not close it.

**Delivered toward the EPIC:**
- `${workflow.env.X}` / `${workflow.secrets.X}` resolution, incl. nested JSONPath and the `$${...}` escape (EPIC approach step 3).
- Read-only discovery: `GET /api/environment`, `GET /api/environment/{key}`, `GET /api/secrets`.
- A pluggable `EnvironmentDAO` / `SecretsDAO` seam that DB-backed impls can later plug into.

**Not in this phase (still open on the EPIC):**
- Different storage model — values come from process env vars under a prefix (`CONDUCTOR_SECRET_*` / `CONDUCTOR_ENV_*`), not server-stored CRUD. No `env_variables`/`secrets` tables; no postgres/mysql/sqlite/redis/cassandra DAO impls.
- Read-only REST — no `PUT`/`DELETE` (create/update/delete), no `GET /api/secrets/{key}` value read, no `/exists`.
- No secret value masking (secrets expose names only here).
- No upstream-split-then-lift port (`org.conductoross.*` translation, `Secret` model, byte-for-byte Orkes parity).
- Docs deliverable (three-meanings disambiguation + verified curl surface).

This PR also includes the `runtimeMetadata` feature (#1255), which is beyond the EPIC's scope.

v1r3n (approved):

common/src/main/java/com/netflix/conductor/common/metadata/tasks/Task.java
```diff
@@ -212,6 +213,15 @@ public boolean isRetriable() {
...
     private String parentTaskId;
 
+    /**
+     * Resolved secret/environment name to value map, injected at poll time from the task
+     * definition's declared {@code runtimeMetadata} names. Wire-only (REST/JSON): never persisted
+     * on {@code TaskModel}, not given a {@code @ProtoField} id, and intentionally excluded from
+     * {@link #toString()}, {@link #equals(Object)}, {@link #hashCode()}, and {@link #copy()}.
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private Map<String, String> runtimeMetadata = new HashMap<>();
```
mp-orkes, line 223: Are we good with the name `runtimeMetadata`? alternatives?

@nthmost @c4lm @bradyyie @manan164 @v1r3n

nthmost-orkes: Maybe `runtimeValues` ?

That's the best i can come up with... honestly, `runtimeMetadata` feels better here the more i think about it.

v1r3n: `runtimeMetadata` is good to me.  Lets merge.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -180,7 +188,22 @@ public List<Task> poll(
...
+                Task task = taskModel.toTask();
+                // Secrets substitution only sees taskModel.getInputData(); when input has been
+                // offloaded to external payload storage, getInputData()/setInputData() operate on
+                // a different field and this substitution silently becomes a no-op.
+                if (taskModel.getExternalInputPayloadStoragePath() != null) {
+                    LOGGER.warn(
+                            "Task {} has externalized input; ${{workflow.secrets.*}} references are not resolved for external payload storage",
+                            taskModel.getTaskId());
+                }
+                task.setInputData(parametersUtils.substituteSecrets(task.getInputData()));
```
nthmost-orkes: If `substituteSecrets` or `runtimeMetadataResolver.resolve()` throws here, the catch at line 207 re-enqueues the task — but the DB record was already written as `IN_PROGRESS` at line 189. On the next poll the task status is not terminal, so the loop retries, hits the same exception, and re-enqueues again. Could these two calls have their own try/catch so a resolution failure does not leave the task stranded until `responseTimeoutSeconds` fires?

core/src/main/java/com/netflix/conductor/core/env/EnvVarLookup.java
```diff
@@ -0,0 +1,55 @@
...
+            value = System.getProperty(full);
+        }
+        return value;
+    }
+
+    /**
+     * Returns all env vars / system properties whose key starts with {@code prefix}, prefix
+     * stripped.
+     */
+    public static Map<String, String> allWithPrefix(String prefix) {
```
nthmost-orkes, line 38: `lookup()` checks `System.getenv()` first and returns early if found, giving env vars higher priority than system properties. `allWithPrefix()` adds env vars first and then overwrites with system properties, reversing that order. If the same key exists in both, `GET /api/environment/{key}` and `GET /api/environment` would return different values — was one ordering intended to win, or should `allWithPrefix()` process system properties first so both methods agree?

test-harness/src/test/groovy/com/netflix/conductor/test/integration/TaskDeclaredSecretsSpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        wfId != null
+        polled.runtimeMetadata['API_KEY'] == 'key-123'
+        polled.runtimeMetadata['REGION'] == 'us-east-1'
+        !polled.runtimeMetadata.containsKey('MISSING_ONE')
+        polled.runtimeMetadata.size() == 2
+
+        and: "the resolved values are NOT persisted — a re-fetch of the task has no secrets"
+        workflowExecutionService.getTask(polled.taskId).runtimeMetadata.isEmpty()
```
nthmost-orkes: This assertion may always pass regardless of the implementation: `TaskModel` has no `runtimeMetadata` field, so `toTask()` via `BeanUtils.copyProperties` never populates it, and `Task.runtimeMetadata` initialises to an empty `HashMap` by default. To actually verify resolved values are not persisted, the test would need to read the raw `TaskModel` from the datastore rather than the `Task` DTO returned by `getTask()`.

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
```diff
@@ -330,6 +359,110 @@ public Map<String, Object> getWorkflowInput(
         return inputParams;
     }
 
+    private Object resolveEnvVariable(String ref) {
```
nthmost-orkes, line 362: `resolveEnvVariable` and `resolveSecretRef` share the same `resolveWithOptionalJsonPath` code path, and there is coverage for secrets JSONPath (e.g. `${workflow.secrets.TOKEN.sub.field}`). The env variant (`${workflow.env.CONFIG.host}` where CONFIG holds a JSON blob) has no test — a regression in the dot-split or `JsonPath.parse` call would be invisible on this branch even as all secrets tests pass.

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
```diff
@@ -330,6 +359,110 @@ public Map<String, Object> getWorkflowInput(
...
+                Object substituted = substituteSecretsValue(original);
+                if (substituted != original) {
+                    if (result == null) {
+                        result = new HashMap<>(src);
+                    }
+                    result.put(e.getKey(), substituted);
+                }
+            }
+            return result == null ? src : result;
+        } else if (value instanceof List) {
```
nthmost-orkes, line 397: The List branch of `substituteSecretsValue` is only exercised by a test whose list elements contain no `${workflow.secrets.*}` references. If the copy-on-write logic here were broken, secrets embedded in list-valued task inputs (e.g. `["Bearer ${workflow.secrets.TOKEN}"]`) would arrive at the worker unresolved with no test catching it.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -180,7 +188,22 @@ public List<Task> poll(
...
+                taskModel
+                        .getTaskDefinition()
```
nthmost-orkes: `taskDef` was already resolved at line 159 and is used at lines 164 and 190, but this optional chain re-invokes `taskModel.getTaskDefinition()` a third time. Was this intentional (e.g. to guard against a mutable optional), or could this just be `if (taskDef != null) task.setRuntimeMetadata(runtimeMetadataResolver.resolve(taskDef.getRuntimeMetadata()))`?

core/src/test/java/com/netflix/conductor/service/ExecutionServiceTest.java
```diff
@@ -335,4 +347,59 @@ public void testGetLastPollTaskReturnsNullWhenEmpty() {
...
+        when(executionDAOFacade.getTaskModel(taskId)).thenReturn(taskModel);
+        when(runtimeMetadataResolver.resolve(any()))
+                .thenReturn(Map.of("API_KEY", "token-value-123"));
+
+        List<Task> polled = executionService.poll(taskType, "worker", null, 1, 100);
+
+        assertEquals(1, polled.size());
+        Task returnedTask = polled.get(0);
+        assertEquals("token-value-123", returnedTask.getRuntimeMetadata().get("API_KEY"));
+        verify(runtimeMetadataResolver).resolve(any());
```
nthmost-orkes: `verify(runtimeMetadataResolver).resolve(any())` confirms the call happened but not what it received. A regression where the wrong TaskDef is fetched, or `getRuntimeMetadata()` returns empty, would keep this test green while workers receive no injected secrets — `verify(runtimeMetadataResolver).resolve(List.of("API_KEY"))` would catch that.

core/src/test/java/com/netflix/conductor/core/env/EnvVariableEnvironmentDAOTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @After
+    public void cleanup() {
+        System.clearProperty("CONDUCTOR_ENV_REGION");
+        System.clearProperty("CONDUCTOR_ENV_ZONE");
+    }
+
+    @Test
+    public void testGetReadsPrefixedProperty() {
+        System.setProperty("CONDUCTOR_ENV_REGION", "us-east-1");
```
nthmost-orkes, line 36: JUnit cannot set real OS env vars, so all three tests go through `System.setProperty()` and never exercise the `System.getenv()` fast-path in `EnvVarLookup.lookup()`. A regression that silently stopped reading actual env vars would be invisible to this suite — not sure there is a clean way to cover this without introducing a seam, but worth noting the gap.

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
mp-orkes: Good catch. I think there is more to this than just a try/catch.

As it is:

1. A missing secret/env value does NOT throw.
2. SecretsDAO/EnvironmentDAO would have to throw.
3. Or an unexpected bug.

Today this try/catch block would be defensive. 

---

However, IF we have a `SecretsDAO` backed by a data store (e.g. AZKV, AWS Parameter store) there could be an API exception. What should happen at that point?

1. Should the task be marked as `FAILED`? 
2. Or should we just return with the unresolved secret? 

If we do (2) this would most likely cause an issue. For example, a worker instead of actually getting a valid token it would get `${workflow.secrets.TOKEN}`, causing a `401` downstream.

core/src/main/java/com/netflix/conductor/core/env/EnvVarLookup.java
mp-orkes: good catch - I'll fix it

test-harness/src/test/groovy/com/netflix/conductor/test/integration/TaskDeclaredSecretsSpec.groovy
mp-orkes: `true`, it's redundant - will delete it

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
mp-orkes: I'll add a test

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
mp-orkes: not intentional.

core/src/test/java/com/netflix/conductor/service/ExecutionServiceTest.java
mp-orkes: true

core/src/test/java/com/netflix/conductor/core/env/EnvVariableEnvironmentDAOTest.java
mp-orkes: Noted but leaving this as it is for now.

core/src/main/java/com/netflix/conductor/core/env/EnvVarLookup.java
mp-orkes: Addressed by https://github.com/conductor-oss/conductor/pull/1251/changes/cfa73cb8c26e5987927c19bc9609b5369857bebb

test-harness/src/test/groovy/com/netflix/conductor/test/integration/TaskDeclaredSecretsSpec.groovy
mp-orkes: Addressed by https://github.com/conductor-oss/conductor/pull/1251/changes/f5bf46b6595a2e8623d3ac97894a44642d6d21b0

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
mp-orkes: Addressed by https://github.com/conductor-oss/conductor/pull/1251/changes/346f2d0ad3c03cd1510dfa3f0668cdd48c1dbc8b

core/src/test/java/com/netflix/conductor/service/ExecutionServiceTest.java
mp-orkes: Addressed by https://github.com/conductor-oss/conductor/pull/1251/changes/989fc16356577bf59efa2b8da9458763f9013911

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
mp-orkes: https://github.com/conductor-oss/conductor/pull/1251/changes/97aa9b11bc742337b8d63a4482f5dac06bba4219

core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java
mp-orkes: https://github.com/conductor-oss/conductor/pull/1251/changes/989fc16356577bf59efa2b8da9458763f9013911

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
mp-orkes: Added a try/catch - but let's come back to this later cc: @v1r3n

mp-orkes:
@nthmost-orkes does this look good now?

nthmost-orkes:
> @nthmost-orkes does this look good now?

That was all that stood out to me, thanks!

nthmost-orkes (approved):
thanks for the responsive updates!

PR #1255, opened 2026-07-06
runtimeMetadata — task-declared secrets & env variables for polled tasks

mp-orkes:
> **Base:** targets `feat/env-backed-secrets-and-environment` (this builds on that PR and reuses its `SecretsDAO`/`EnvironmentDAO`) — **not `main`.** Review the incremental diff.

## What

A `TaskDef` declares the **names** of values it needs. At **worker poll**, the server resolves each name (secrets store first, environment variables as fallback — env-backed by default) and injects a name→value map onto the returned `Task`. The workflow definition stays clean.

The field is named **`runtimeMetadata`** (per stakeholder request). It is a **source-neutral** concept covering **both secrets and environment variables** — a resolved value may come from either source. (Semantically these are runtime parameters injected at poll; the `runtimeMetadata` name is the agreed term.)

Complementary to the existing `${workflow.secrets.X}` / `${workflow.env.X}` reference resolution — a second way to hand a worker a managed value, declared once on the task definition.

Design doc: **`design/runtime-metadata.md`** (in this PR).

## Mechanism

```
TaskDef "llm_call":  runtimeMetadata: ["OPENAI_API_KEY", "REGION"]   # list of names

worker polls a "llm_call" task
  → resolve each name: SecretsDAO(CONDUCTOR_SECRET_<name>) then EnvironmentDAO(CONDUCTOR_ENV_<name>)
  → returned Task:  runtimeMetadata: { "OPENAI_API_KEY": "…", "REGION": "…" }   # name→value
```

## Model
- `TaskDef.runtimeMetadata` (`List<String>`) — the names the task needs.
- `Task.runtimeMetadata` (`Map<String,String>`) — resolved name→value, injected at poll.

## Key properties

- **Wire-only, never persisted.** Values are set only on the polled `Task` in `ExecutionService.poll`, never on the persisted `TaskModel`. `Task.runtimeMetadata` is excluded from `toString()` (no values in logs), `equals`/`hashCode`, and `copy()`; serialized only when non-empty (`@JsonInclude(NON_EMPTY)`).
- **Missing names are omitted** (logged), not injected as null.
- **Reuses the base PR's DAOs** via `RuntimeMetadataResolver` (secrets-first, env-fallback); no new config (inert when both providers are `noop`). The prefix boundary still applies — a TaskDef can only name operator-exposed `CONDUCTOR_SECRET_*` / `CONDUCTOR_ENV_*` vars.
- **REST/JSON only** (no `@ProtoField`); **poll-only** (worker tasks, not server-side system tasks).

## How to test

Register a `TaskDef` with `runtimeMetadata: ["API_KEY"]`, set `CONDUCTOR_SECRET_API_KEY` (or `CONDUCTOR_ENV_…`) on the server, poll the task → `Task.runtimeMetadata` is populated. Covered by unit tests (`RuntimeMetadataResolverTest`, `ExecutionServiceTest`, model tests) and an end-to-end `TaskDeclaredSecretsSpec` (secrets-store hit + env fallback + missing-omitted + not-persisted).

## Open questions for review (see design doc §10)

- Unified list + secrets→env fallback into one `runtimeMetadata` map (current) vs. splitting secrets and env into separate fields.
- Declared-but-missing name: silent omission (current) or surface to the operator beyond the server log?
- Is REST-only acceptable for v1, or is gRPC (`@ProtoField`) support required?


mp-orkes:
@v1r3n 
Screenshot 2026-07-07 at 13 44 59

v1r3n (approved):

v1r3n (commented):
Solid, well-scoped PR that adds task-declared runtime secrets/env values injected at poll time.

**Strengths**
- The critical security invariant is enforced correctly: resolved values are set only on the outgoing wire `Task` in `ExecutionService.poll` (never on `TaskModel`), so nothing lands in the datastore/UI/history. `Task.runtimeMetadata` is excluded from `copy()`/`deepCopy()`/`equals()`/`hashCode()`/`toString()` and serialized only when non-empty (`@JsonInclude(NON_EMPTY)`), so normal workflow/task GET responses (built via `TaskModel.toTask()`) never carry it.
- DAO method names (`getSecret`, `getEnvVariable`) match; secrets-first / env-fallback ordering matches the design doc.
- Good test coverage: unit tests for the resolver, the poll injection point, model round-trips (incl. equals/hashCode/copy/serialization), and a full Spring integration spec asserting non-persistence and env fallback.
- Design doc clearly documents the decisions, non-goals (gRPC, system tasks), and the known output-leakage caveat.

**Minor**
- Missing declared names log a WARN on every poll, which can be noisy on high-throughput task types (see inline).

No blocking correctness or security issues found.

core/src/main/java/com/netflix/conductor/core/secrets/RuntimeMetadataResolver.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        for (String name : names) {
+            String value = secretsDAO.getSecret(name);
+            if (value == null) {
+                value = environmentDAO.getEnvVariable(name);
+            }
+            if (value != null) {
+                out.put(name, value);
+            } else {
+                LOGGER.warn(
```
v1r3n, line 52: `resolve()` runs on every poll of a task that declares `runtimeMetadata`, and a declared-but-missing name logs a WARN each time. For a busy task type (polled many times/second) with an optional or not-yet-provisioned name, this floods the logs. Consider debug level, or de-duplicating/rate-limiting the warning (e.g. log once per name). The design doc's open question #2 also flags that missing-name handling may deserve reconsideration.

v1r3n (approved):

PR #1259, opened 2026-07-08
fix(sweeper): re-queue on lock miss to eliminate multi-minute workflow pauses

NicholasDCole:
## Problem

Reported on Conductor **3.30.2** (Spanner persistence, Redis queue + locks): a workflow with many JOIN boundaries stalls for an intermittent **~10 minutes** per run after upgrading from 3.21.4.

## Root cause

The active sweeper (`org.conductoross.conductor.core.execution.WorkflowSweeper`, default via `conductor.app.sweeper.enabled`) **bare-returned** when it couldn't acquire the workflow lock — no `decide()`, no re-queue:

```java
    log.error("Couldn't acquire lock to sweep workflow {}", workflowId);
    return;   // <-- workflow parks
}
```

A lock-contended workflow (e.g. the post-completion `decide()` at a FORK/JOIN boundary) then had no near-term wake-up. Its only remaining decider-queue entry is postponed by `ExecutionService.adjustDeciderQueuePostpone()` to `responseTimeoutSeconds * 1000` when a worker polls a task. With the affected task defs carrying `responseTimeoutSeconds=600`, that fallback is **600 s ≈ 10 minutes** — the observed pause. It's intermittent ("sometimes") because it only surfaces when the event-driven decide is missed.

## Fix

On a lock miss, re-queue the workflow to the decider queue with a bounded backoff (`lockLeaseTime/2`, capped by `maxPostponeDurationSeconds`) so a contended workflow is retried in **seconds** instead of parking:

```java
    long backoffSeconds = lockContentionBackoffSeconds();
    log.error("Couldn't acquire lock to sweep workflow {}", workflowId);
    queueDAO.push(DECIDER_QUEUE, workflowId, 0, backoffSeconds);
    return;
}
```

- The identical backoff math previously inlined in the `decide()==null` path is factored into a shared `lockContentionBackoffSeconds()` helper.
- The `"Couldn't acquire lock to sweep workflow"` **error log is kept verbatim** so existing log-based diagnostics/alerts still fire.
- This changes only recovery *latency*; it does not alter the (correct) `responseTimeout`-based postpone for genuinely long-running tasks.

## Tests

| Test | Purpose |
|---|---|
| `WorkflowSweeperTest.sweepReQueuesOnLockMissWithBoundedBackoffInsteadOfParking` | **Regression** — fails before fix (zero pushes), passes after (30 s re-queue) |
| `WorkflowSweeperTest.sweepReQueuesWith30sBackoffWhenDecideCannotAcquireLock` | Pins the `decide()==null` backoff at `lockLeaseTime/2` |
| `ExecutionServiceTest.testPollSetsDeciderQueuePostponeToResponseTimeout_reproducesTenMinutePause` | Documents the 600 s decider postpone that is the pause duration's source |
| `TestWorkflowSweeper` (legacy path) | Legacy-sweeper reproduction of the same `responseTimeout`-driven fallback, retained as a diagnostic |

All green; spotless clean.

## Note for affected deployments

As an immediate config mitigation before adopting a build with this fix, lowering `responseTimeoutSeconds` on the task defs (e.g. 600 → 60) shrinks the fallback pause proportionally.

## Test evidence (validated with / without the fix)

All tests were run both with the fix in place and with the lock-miss re-queue reverted (call site A in `WorkflowSweeper.sweep()` restored to the original bare `return`, i.e. delete `queueDAO.push(DECIDER_QUEUE, workflowId, 0, backoffSeconds)`), on the same commit otherwise. The two tests that assert the fix flip from pass → fail when the fix is removed, confirming they are genuine regression gates (not vacuous). Environment: `local_only` execution lock; the e2e spec runs against the real Redis-backed decider queue the harness provisions via Testcontainers (`redis:6.2-alpine`).

### ✅ With the fix (current branch)

```
$ ./gradlew :conductor-core:test \
    --tests "org.conductoross.conductor.core.execution.WorkflowSweeperTest" \
    --tests "com.netflix.conductor.service.ExecutionServiceTest" \
    --tests "com.netflix.conductor.core.reconciliation.TestWorkflowSweeper"
WorkflowSweeperTest      tests=6  failures=0  skipped=0
ExecutionServiceTest     tests=13 failures=0  skipped=0
TestWorkflowSweeper      tests=20 failures=0  skipped=0

$ ./gradlew :conductor-test-harness:test \
    --tests "com.netflix.conductor.test.integration.DynamicForkJoinLockContentionSpec"
DynamicForkJoinLockContentionSpec  tests=2  failures=0  skipped=0
  ✓ sweeper re-queues a lock-contended workflow instead of leaving it parked at the response timeout (17.3s)
  ✓ control: with no lock contention the same JOIN schedules the next task immediately (0.46s)
BUILD SUCCESSFUL
```

### ❌ With the fix reverted (bare `return` on lock miss)

**Unit regression** — `WorkflowSweeperTest.sweepReQueuesOnLockMissWithBoundedBackoffInsteadOfParking` (tests=6, failures=1):

```
Wanted but not invoked:
queueDAO.push(
    "_deciderQueue",
    "workflow-id",
    0,
    30L
);
-> at WorkflowSweeperTest.sweepReQueuesOnLockMissWithBoundedBackoffInsteadOfParking(WorkflowSweeperTest.java:221)
Actually, there were zero interactions with this mock.
```

The contended `sweep()` never re-queued the workflow — exactly the bug (the 30s = `lockLeaseTime/2` backoff push is absent).

**E2E** — `DynamicForkJoinLockContentionSpec` › "sweeper re-queues a lock-contended workflow…" (tests=2, failures=1; the control still passes):

```
Condition not satisfied:

queueDAO.containsMessage(Utils.DECIDER_QUEUE, workflowInstanceId)
|        |               |     |              |
|        false           |     _deciderQueue  bfd6fe7a-9aa6-4cfe-b714-06ea94c9c65a
|                        class com.netflix.conductor.core.utils.Utils
<io.orkes.conductor.mq.dao.RedisQueueDAO@... queueNamespace=integtest.test ...>
```

After the JOIN completes under a held workflow lock (post-JOIN `decide()` returns null, `integration_task_4` never scheduled) and the decider entry is cleared, a contended `sweep()` leaves nothing in `_deciderQueue` for the workflow — it is parked with no near-term wake-up (in production, until the `responseTimeoutSeconds`=600 postpone fires). The control case (no contention) schedules `integration_task_4` immediately in both variants, isolating the lock as the sole cause.

### Restore check

Restoring the fix returns `WorkflowSweeper.java` to this PR's committed content (`git diff` empty) and both suites go green again.


nthmost-orkes (commented):
The root cause identification looks correct and this fix seems necessary — without it, users running `workflow-offset-timeout=0s` appear to have no recovery path shorter than `responseTimeoutSeconds` (filed as #1273). One thing worth double-checking: `queueDAO.push()` adds a new decider queue entry rather than updating an existing one, whereas the rest of the codebase uses `setUnackTimeoutIfShorter()` for this pattern — is it possible that with `lockTimeToTry=1000ms`, multiple sweeper threads could push duplicate entries for the same workflow in rapid succession? Also curious whether the `lockLeaseTime/2` backoff should be sourced from the Redis execution lock specifically, or whether Spanner deployments (which have a separate `conductor.spanner.lockLeaseTimeMilliSeconds`) would end up with a longer retry window than intended.

v1r3n (approved):

manan164:
The 600s diagnosis is right, but the change is in the wrong file. The lost wake-up is `decide()` silently returning null on a lock miss — `core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java:1213-1215`:

```java
boolean lockAcquired = executionLockService.acquireLock(workflowId);
if (!lockAcquired) {
    return null;
}
```

Both completion-event callers ignore that null (`WorkflowExecutorOps.java:1072`, `AsyncSystemTaskExecutor.java:197`), so the workflow parks until its decider entry fires at +600s. Your sweeper change never executes during that window: `sweep()` only runs when the entry is popped, and it isn't due for 600s.

**Change to make** — in `decide(String)` at `WorkflowExecutorOps.java:1213`, replace the block above with:

```java
if (!lockAcquired) {
    long backoffMillis = Math.max(properties.getLockTimeToTry().toMillis() / 2, 100);
    queueDAO.push(DECIDER_QUEUE, workflowId, 0, Duration.ofMillis(backoffMillis));
    return null;
}
```

(`queueDAO`, `DECIDER_QUEUE`, `Duration`, and `getLockTimeToTry()` — default 500ms — are all already available in that file. The `push(String, String, int, Duration)` overload exists on `QueueDAO:47`.)

Backoff must be `lockTimeToTry`-scale (~250ms), not `lockLeaseTime/2` (30s): contention lasts milliseconds; 30s is for orphaned locks.

Also:
- `DynamicForkJoinLockContentionSpec`: drop the `queueDAO.remove()` before the sweep — on Redis, `pop` rescores the entry to +30s rather than removing it, so that state can't occur in production. Assert instead that `integration_task_4` is scheduled within seconds of releasing the lock (passes only with the `decide()` fix).
- Remove customer names from tests (`MIKHAIL_*`, `CloudAware`, `testRegressWorkflow*`).
- Drop the three new `TestWorkflowSweeper` tests — legacy sweeper, deprecated and off by default.

Your `WorkflowSweeper.java` change is fine to keep as a backstop.

NicholasDCole:
You're right — thanks for the precise catch, @manan164. Reworked and pushed.

**Concession on placement:** the lost wake-up is `decide()` swallowing the lock miss, and my sweeper change couldn't help the real case — `sweep()` only runs when the decider entry is popped, and that entry isn't due for `responseTimeoutSeconds`, so it never fired during the pause window. Moved the fix to where the miss happens.

**What changed:**
- `WorkflowExecutorOps.decide(String)`: on a lock miss, re-queue to `DECIDER_QUEUE` with a `lockTimeToTry/2` backoff (contention-scale, per your note — not `lockLeaseTime/2`) before returning `null`. Primary fix; benefits every caller (`updateTask`, `AsyncSystemTaskExecutor`, sweeper).
- `WorkflowSweeper.sweep()`: kept the top-level lock-miss re-queue as a backstop but at the same short backoff; dropped the redundant `decide()==null` re-queue (`decide()` owns it now); removed the lock-lease-based helper.
- New unit test `TestWorkflowExecutor.testDecideReQueuesWorkflowOnLockMiss` (fails against the old bare return).

**E2E — you were right that it was manufactured.** Dropped the `queueDAO.remove()` + manual-sweep-under-lock. It's now a true reproduction: drive the real dynamic fork/join to the JOIN boundary, hold the workflow lock from a foreign thread, run the JOIN (its post-completion `decide()` misses the lock), release the lock, then assert `integration_task_4` is scheduled within seconds **via the real background sweeper** — no manual sweep, no synthetic queue state. Verified it **times out without the `decide()` fix** and passes with it; the no-contention control passes either way.

Also scrubbed the customer identifiers from the tests and dropped the three legacy (deprecated, off-by-default) `TestWorkflowSweeper` cases.

Your #3787 (`rerun` → `423 LOCK`) is a separate, correct catch — a caller-facing mutation where a deferred re-queue wouldn't make sense.

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
```diff
@@ -158,9 +158,24 @@ public CompletableFuture<Void> sweepAsync(String workflowId) {
...
+    }
+
     public void sweep(String workflowId) {
         if (!executionLockService.acquireLock(workflowId)) {
+            // Backstop: another worker/node holds the lock. Re-queue for a prompt retry rather than
+            // bare-returning (which would leave the workflow parked on its far-future decider-queue
+            // entry). The primary re-queue on a contended decide lives in
+            // WorkflowExecutorOps.decide().
             log.error("Couldn't acquire lock to sweep workflow {}", workflowId);
+            queueDAO.push(DECIDER_QUEUE, workflowId, 0, lockContentionBackoff());
```
manan164: We can remove this.

manan164 (approved):

core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java
NicholasDCole: Removed sweeper changes @manan164

PR #1264, opened 2026-07-09
fix: POST /metadata/workflow honors overwrite query param

### LINKED ISSUE #1263: POST /api/metadata/workflow ignores overwrite=true and returns HTTP 500 if workflow exists
## Summary

`POST /api/metadata/workflow?overwrite=true` returns HTTP 500 when a workflow with the same name and version already exists, instead of overwriting it.

## Expected behavior

When `overwrite=true` is passed as a query param, an existing workflow definition should be replaced.

## Actual behavior

```
{"status":500,"message":"Workflow with <name>.<version> already exists!","retryable":false}
```

The `overwrite` query parameter is silently ignored. `POST` always behaves as a create-only operation.

## Workaround

`PUT /api/metadata/workflow` (body: array of `WorkflowDef`) performs an upsert and works correctly. SDK users who need overwrite behavior must call this endpoint directly.

## Impact

Every SDK user who calls `Register(overwrite=true)` on an already-registered workflow hits this error. The hello world examples for all SDKs fail on the second run because of this.

## Reproduce

1. Register any workflow via `POST /api/metadata/workflow`.
2. Call `POST /api/metadata/workflow?overwrite=true` with the same workflow def.
3. Observe HTTP 500.

Confirmed against `conductoross/conductor-standalone:3.15.0`.

## Fix

`POST /api/metadata/workflow` should respect the `overwrite` query parameter and update the existing definition when `overwrite=true`.

Related: conductor-oss/go-sdk#252

nthmost-orkes:
`POST /api/metadata/workflow?overwrite=true` was ignoring the `overwrite` param and returning HTTP 500 if the workflow already existed.

Fix: when `overwrite=true`, route to `metadataDAO.updateWorkflowDef()` instead of `createWorkflowDef()`. The `overwrite=false` path (and the no-param default) behave exactly as before.

Fixes #1263
Related: conductor-oss/go-sdk#252, conductor-oss/go-sdk#253


v1r3n (dismissed):
The change correctly threads an `overwrite` param through `MetadataResource.create`, backed by a new `findWorkflowDef` on the service/DAO. The core logic reads well (empty → register, overwrite → update, else conflict), and `updateWorkflowDef(WorkflowDef)`/`ConflictException(String, Object...)`/`WorkflowDef.key()` all exist as used.

However, there is one blocking problem: the `create` method signature changed from `create(WorkflowDef)` to `create(WorkflowDef, boolean)`, but `MetadataResourceTest.testCreateWorkflow` still calls `metadataResource.create(workflowDef)` with a single argument. This will not compile, breaking the `rest` module build.

Secondary points: no new test exercises the `overwrite=true` (update) or conflict (409) branches, and the check-then-act on `findWorkflowDef`/`register` is a non-atomic TOCTOU (acceptable, DB uniqueness will still guard, but worth noting).

---
### Inline findings
- `rest/src/test/java/com/netflix/conductor/rest/controllers/MetadataResourceTest.java:51` — This won't compile: `create` now has signature `create(WorkflowDef, boolean overwrite)`, but this call still passes a single argument. Update to `metadataResource.create(workflowDef, false)` (and ideally add cases covering the `overwrite=true` update path and the `ConflictException` path). As written, the `rest` module build breaks.
- `rest/src/main/java/com/netflix/conductor/rest/controllers/MetadataResource.java:63` — New behavior (returning 409 on existing def when overwrite=false, and updating when overwrite=true) has no test coverage. Consider adding tests for both the conflict and overwrite-update branches so the new contract is pinned down.

nthmost-orkes:
Thanks for the thorough review, Viren!

Fixed in b9743cd:

- **Compile break**: updated `testCreateWorkflow` to call `create(workflowDef, false)` and stubbed `findWorkflowDef` to return `Optional.empty()`
- **Missing coverage**: added `testCreateWorkflowOverwriteUpdatesExisting` (overwrite=true path → calls `updateWorkflowDef`) and `testCreateWorkflowConflictThrowsWhenNoOverwrite` (409 path → `ConflictException` thrown, neither register nor update called)

On the TOCTOU note — agreed it's acceptable given the DB uniqueness constraint as the real guard; left it as-is.

v1r3n (commented):
Threads an `overwrite` query param through `POST /api/metadata/workflow`, backed by a new `findWorkflowDef(name, version)` on the service and DAO. The control flow is correct and readable:

- blank/null name → delegate straight to `registerWorkflowDef` (preserves full `@Valid` validation errors, matching the earlier fix in `32d01bd`);
- def not found → register;
- found + `overwrite=true` → `updateWorkflowDef`;
- found + `overwrite=false` → `ConflictException`.

I verified that `WorkflowDef.version` defaults to `1` (so the `@NotNull` version constraint on `findWorkflowDef` is always satisfied), and that both `registerWorkflowDef` and `updateWorkflowDef(WorkflowDef)` are `@Valid`-annotated, so validation is not lost on the update branch. `updateWorkflowDef`, `ConflictException(String, Object...)`, and `WorkflowDef.key()` all exist as used.

The compile break flagged in the prior review (`testCreateWorkflow` calling the old single-arg signature) is fixed, and the new tests cover the overwrite-update and conflict branches. The previously noted TOCTOU is acceptable given the DB uniqueness constraint.

No blocking issues.

v1r3n (approved):

PR #1271, opened 2026-07-09
feat(ui): Agent Execution debugger, Agent Definition graph, and Create Agent CTA

bradyyie:
## Summary

Brings the AgentSpan product UI's execution-debugging experience into `ui-next` for agent-classified executions (workflows compiled by the AgentSpan SDK, tagged via `metadata.agentDef`/`agent_sdk` — see #1268), rebuilt natively against ui-next's own components/types rather than vendored, so it stays in sync with ui-next's actual infra instead of drifting like the pre-existing agentspan/ui fork.

Regular (non-agent) workflow executions are unaffected by every change in this PR — all new behavior is gated on `isAgentWorkflowExecution()` (`pages/execution/helpers.ts`).

## What's new

**Agent Execution debugger tab** (`pages/execution/AgentExecution/`) — turn-by-turn reaflow diagram of an agent run (LLM calls, tool calls, handoffs, guardrails, sub-agents), an event timeline view, a resizable detail panel (Summary/Input/Output/JSON, Monaco + markdown rendering), and a human-in-the-loop approval panel for paused executions. Defaults to this tab automatically for agent-classified executions; regular workflows keep Diagram as the default.

**Agent Definition graph** — the "Agent Definition" tab (agent executions' relabeled Summary tab, same as AgentSpan's own UI) now renders a graph of the agent's static structure — model/strategy, sub-agents (fanned out or chained with gate/decision nodes for sequential strategies), tools by category (`@TOOL`/HTTP/MCP/RAG), guardrails — instead of the plain Conductor execution summary.

**Curated tab set for agent executions** — matches AgentSpan's own tab bar exactly: Agent Execution, Debug View (Diagram, relabeled), Task List, Timeline, Agent Definition, JSON. Drops Workflow Input/Output, Variables, and Tasks to Domain, which AgentSpan's UI never had. Regular workflows keep the full original tab set, unrelabeled.

**Create Agent CTA** — agents aren't created directly in this UI yet (SDK/CLI-only). Adds a "Create Agent" button (top-right of the Agents list) opening a modal that points to the AgentSpan SDK/CLI on GitHub, plus the same guidance as a persistent link in both Agents' and Agent Executions' empty states.

**Dedicated `/agentExecutions/:id` route** — agent executions previously shared `/execution/:id` with plain workflows, so the sidebar highlighted "Workflow" instead of "Executions" (under Agents) when viewing one. Now routed under `/agentExecutions/:id/:taskId?` (same `<Execution/>` component — same precedent as `WorkflowDefinition`/`TaskDefinition`/etc. at multiple paths), with a matching `activeRoutes` entry so the sidebar highlights correctly, and a client-side redirect (`replace`, no extra history entry) for any existing `/execution/:id` link that turns out to be agent-classified.

**Fix: agent executions leaking into the plain Executions search** — `/executions` (`BasicSearch`/`AdvancedSearch`) was showing AgentSpan-compiled runs alongside regular workflows. The backend already supports a `classifier` param on `/workflow/search` (#1268); `SearchObj` now threads it through as its own REST param (not folded into the free-text query), and both search UIs pass `classifier: "workflow"` unconditionally.

## Notable fix along the way

`initDiagram`'s default-tab transition was landing the XState state node on `agentExecution` but never syncing `context.currentTab` — the field `LeftPanelTabs`/`Execution.tsx` actually read to decide what to render — so the debugger silently never appeared on initial load despite the machine "knowing" it should. Root-caused and fixed via `state/agentDefaultTab.test.ts`, which interprets the real machine end-to-end (mocked fetch only) rather than testing guards in isolation.

While chasing that down, found 16 test suites silently failing on `localStorage` (a newer-Node experimental global that's non-functional without `--localstorage-file`, breaking anything that reads `localStorage` at import time). Added a minimal polyfill to `setupTests.ts` — incidental fix, unrelated to agents, but it unblocks those suites (385 → 600 tests passing).

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm prettier:check` all clean.
- `pnpm test`: 47/47 suites, 600 passed + 1 skipped, zero regressions.
- `pnpm build` succeeds.
- Manually verified end-to-end against a local Postgres-backed server with AgentSpan embedded (`conductor.integrations.ai.enabled=true`): a real deployed agent (`classifier_greeter`), its execution, the debugger tab defaulting correctly, the Agent Definition graph rendering from real `agentDef` metadata, the `/agentExecutions/:id` route + sidebar highlighting, the plain Executions search excluding agent runs, and the empty-state/Create-Agent flow after deleting the agent definition.

## Notes for reviewers

- New components mirror `AgentExecutionDiagram`'s existing pan/zoom/reaflow architecture (`AgentDefinitionView`) rather than introducing a new pattern.
- `SubAgentTable.tsx` and a few unused exports in agentspan's original `agentExecutionUtils.ts` (`groupEventsIntoTurns`, `transformSubWorkflowToAgentRun`, etc.) were confirmed dead code upstream and intentionally not ported.


v1r3n (approved):

ui-next/src/pages/agent/CreateAgentSdkModal.tsx
ajlai-orkes, line 1: this modal looks a bit inconsistent with the other modals we have

Image

ui-next/src/pages/execution/AgentExecution/TurnBar.tsx
```diff
@@ -0,0 +1,189 @@
...
+  useEffect(() => {
+    const el = scrollRef.current?.querySelector(
+      `[data-turn="${selectedTurn}"]`,
+    ) as HTMLElement | null;
+    el?.scrollIntoView({
+      block: "nearest",
+      inline: "center",
+      behavior: "smooth",
+    });
+  }, [selectedTurn]);
```
ajlai-orkes: The useEffect + querySelector('[data-turn="..."]') approach works, but it's a bit indirect — we're querying the DOM for an element React already knows about.

Consider ref-on-selected instead:
```
useLayoutEffect(() => {
  selectedRef.current?.scrollIntoView({
    block: "nearest",
    inline: "center",
    behavior: "smooth",
  });
}, [selectedTurn]);
```

v1r3n:
Instead of the modal dialog, can we use the space that in the viewport to give links to agentspan, code samples etc?
basically something like this https://agentspan-ai.github.io/agentspan/#get-started-in-30-seconds

bradyyie:
> Instead of the modal dialog, can we use the space that in the viewport to give links to agentspan, code samples etc? basically something like this https://agentspan-ai.github.io/agentspan/#get-started-in-30-seconds

Modal was a product request on the create agent button till we support on both sides.

ui-next/src/pages/execution/AgentExecution/TurnBar.tsx
bradyyie: done !

PR #1279, opened 2026-07-13
fix(handoff): prevent infinite router loop in HANDOFF multi-agent strategy

shaileshpadave:
## What was the problem?

When running a HANDOFF multi-agent setup (like `13_hierarchical_agents.py`), the router LLM would keep re-calling agents that had already responded instead of saying DONE — causing the workflow to loop indefinitely until it hit the 100-turn hard cap.

Two things caused this:

1. **The router had no memory of who already spoke.** Nothing in the prompt told the LLM that an agent it already routed to had responded. So it would happily route to the same agent again and again.

2. **The default turn limit was too high.** The default max turns was 100, so even with a bad router decision the loop would grind through all 100 iterations before stopping. For a 3-agent team, the loop should stop after at most 4 turns — not 100.

## How we fixed it

- **Router prompt rewritten** — the router is now told each agent may be called at most once. Agents that have already responded leave a visible marker in the conversation history, and the router is instructed to say DONE once all needed agents have replied.

- **Smarter default turn limit** — the default max turns for HANDOFF is now `number of agents + 1` instead of 100. This means a 2-agent team caps at 3 turns, a 5-agent team at 6, and so on. Explicit `maxTurns` settings still override this.

## Verified

Ran `13_hierarchical_agents.py` (CEO → Engineering Lead → Backend Dev) against the fixed server — completed cleanly with no looping.

### DIFF
--- a/agentspan-server/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/AgentCompiler.java
+++ b/agentspan-server/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/AgentCompiler.java
@@ -290,7 +290,7 @@ public class AgentCompiler {
         // Guarded path: LLM + guardrails in DoWhile loop
         String contentRef = ref(llmRef + ".output.result");
         String loopRef = toRef(config.getName()) + "_loop";
-        int maxTurns = config.getMaxTurns() > 0 ? config.getMaxTurns() : 25;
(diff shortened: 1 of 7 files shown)


shaileshpadave:
The SDK default is 25, but users can pass any value they want:
  
  ceo = Agent(
      name="ceo",
      instructions="...",
      max_turns=5, 
  )
  
  so lets not have pr for this

PR #1294, opened 2026-07-15
fix(executor): notify TaskStatusListener on sync system task state change

### LINKED ISSUE #1176: [BUG] TaskStatusListener.onTaskInProgress not fired for sync system tasks (HUMAN) — engine-driven transitions to IN_PROGRESS bypass the listener
### Describe the bug

When a workflow contains a sync system task (`WorkflowSystemTask.isAsync() == false`) that transitions itself to `IN_PROGRESS` from `WorkflowExecutor.scheduleTask(...)` — `HUMAN` is the canonical example — the registered `TaskStatusListener` never receives an `onTaskInProgress` event. The terminal  `COMPLETED` event still fires because human completion arrives via `POST /api/tasks` → `WorkflowExecutor.updateTask(TaskResult)`, which is the only place that calls `notifyTaskStatusListener(task)` today.
  
Where the gap is, in `core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java` (sync branch of `scheduleTask`):
  
  ```java
  for (TaskModel task : systemTasks) {
      WorkflowSystemTask workflowSystemTask = systemTaskRegistry.get(task.getTaskType());
      ...
      if (!workflowSystemTask.isAsync()) {
          try {
              workflowSystemTask.start(workflow, task, this);   // Human.start() -> setStatus(IN_PROGRESS)
          } catch (Exception e) { ... }
          startedSystemTasks = true;
          executionDAOFacade.updateTask(task);                  // persisted, but listener never notified
      } else {
          tasksToBeQueued.add(task);                            // async path - SCHEDULED fires via addTaskToQueue
      }
  }
  ```
  
The same shape exists for async system tasks in `AsyncSystemTaskExecutor.executeSystemTask(...)` -`systemTask.start(...)` flips the task to `IN_PROGRESS`, the `finally` block calls `executionDAOFacade.updateTask(task)`, and the listener is again never invoked. We have not independently reproduced this on `WAIT` yet, but the code path is structurally identical so the same fix applies.
  
**Reproduction (verified on v3.30.2)**
  
Wire up `task-status-listener` with the HTTP publisher pointing at any receiver:
  
  ```properties
  conductor.task-status-listener.type=task_publisher
  conductor.status-notifier.notification.url=http://my-listener:8080
  conductor.status-notifier.notification.endpointTask=task_status
  conductor.status-notifier.notification.subscribedTaskStatuses=SCHEDULED,IN_PROGRESS,COMPLETED,FAILED,CANCELED,TIMED_OUT
  ```
  
Workflow definition with a single HUMAN task:
  
  ```json
  {
    "name": "human_demo",
    "version": 1,
    "tasks": [
      { "name": "review", "taskReferenceName": "review", "type": "HUMAN" }
    ]
  } 
  ```
  
Start the workflow, wait for the human task, then complete it via:
  
  ```bash
  curl -X POST http://localhost:8080/api/tasks \
    -H 'Content-Type: application/json' \
    -d '{"taskId":"<id>","status":"COMPLETED","workflowInstanceId":"<wf>"}'
  ```
  
Notifications observed at the receiver:
  
  ```
  POST /task_status  -> { "taskId": "...", "status": "COMPLETED" }
  # no IN_PROGRESS event
  ```
  
Expected: `IN_PROGRESS` is also emitted at the moment the engine flips the human task from `SCHEDULED` to `IN_PROGRESS` (i.e. when the workflow scheduler runs `Human.start(...)`).
  
**Suggested fix**
  
Notify the listener after the engine persists a sync system task transition:
  
  ```diff
  --- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
  +++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
  @@ scheduleTask(...)
                   if (!workflowSystemTask.isAsync()) {
                       try {
                           workflowSystemTask.start(workflow, task, this);
                       } catch (Exception e) { ... }
                       startedSystemTasks = true;
                       executionDAOFacade.updateTask(task);
  +                    try {
  +                        notifyTaskStatusListener(task);
  +                    } catch (Exception e) {
  +                        LOGGER.error("Error notifying TaskStatusListener for sync system task {}",
  +                                     task.getTaskId(), e);
  +                    }
                   } else {
                       tasksToBeQueued.add(task);
                   }
  ```
  
And the symmetric call in the async path:
  
  ```diff
  --- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
  +++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
  @@ executeSystemTask(...) finally
           } finally {
               executionDAOFacade.updateTask(task);
  +            try {
  +                // notify listener for engine-driven system-task transitions
  +                // (mirrors notifyTaskStatusListener in WorkflowExecutor.updateTask)
  +            } catch (Exception e) { ... }
               if (shouldRemoveTaskFromQueue) { ... }
  ```
  
(Or, cleaner: fold listener notification into a single helper that wraps every `executionDAOFacade.updateTask(task)` site so engine-driven and worker-driven transitions emit consistently.)
  
Happy to PR — would also like guidance on whether the preferred shape is per-call-site notification or a centralized notify-on-persist helper.

PreAgile:
Pull Request type
----
- [x] Bugfix

Changes in this PR
----
A synchronous system task (`HUMAN` is the canonical one) sets its own status to `IN_PROGRESS` inside `start()` while it is being scheduled. That branch of `scheduleTask()` persists the task with `executionDAOFacade.updateTask(task)` but never calls `notifyTaskStatusListener`, which today only runs on the `updateTask(TaskResult)` hand-off and the cancel path. So a registered `TaskStatusListener` never sees `onTaskInProgress` for a HUMAN task, even though it does get the terminal `COMPLETED` event later (that arrives via `POST /api/tasks`). The result is a listener that reports the end of a wait but not the start of it.

This notifies the listener right after the task is persisted in the synchronous branch of `scheduleTask()`. `notifyTaskStatusListener` already dispatches by the task's current status, so it also fires the correct terminal event for a sync task that reaches a terminal status directly inside `start()`, and it stays a no-op for `SCHEDULED`.

Added tests in `TestWorkflowExecutor`:
- a sync HUMAN task fires `onTaskInProgress` once
- a sync task that completes inside `start()` fires `onTaskCompleted`, not `onTaskInProgress`
- an async system task is not notified from `scheduleTask` (it is queued and notified on its own path), so there is no double notification

Fixes #1176

Alternatives considered
----
The same gap exists in `AsyncSystemTaskExecutor.executeSystemTask()`, which also transitions a task to `IN_PROGRESS` and persists it without notifying the listener. It has no `TaskStatusListener` dependency today, so wiring it up is a larger change. I left it out of this PR to keep the fix focused on the sync path, but I'm happy to extend it here or in a follow-up if you'd prefer.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
@@ -1845,6 +1845,20 @@ public class WorkflowExecutorOps implements WorkflowExecutor {
                     }
                     startedSystemTasks = true;
                     executionDAOFacade.updateTask(task);
+                    // Synchronous system tasks (e.g. HUMAN) transition their own status inside
+                    // start(...) here in the decide loop, and never reach the
+                    // updateTask(TaskResult)
+                    // hand-off path that notifies the listener. Notify explicitly so status changes
+                    // such as IN_PROGRESS are not silently dropped. See issue #1176.
+                    try {
+                        notifyTaskStatusListener(task);
+                    } catch (Exception e) {
+                        String errorMsg =
+                                String.format(
+                                        "Error while notifying TaskStatusListener: %s for workflow: %s",
+                                        task.getTaskId(), workflow.getWorkflowId());
+                        LOGGER.error(errorMsg, e);
+                    }
                 } else {
                     tasksToBeQueued.add(task);
                 }
@@ -1893,6 +1907,7 @@ public class WorkflowExecutorOps implements WorkflowExecutor {
             WorkflowSystemTask systemTask, WorkflowModel workflow, TaskModel task) {
         Map<String, Object> literalInput = task.getInputData();
         try {
+            task.setInputData(parametersUtils.substituteSecrets(literalInput));
             task.setInputData(parametersUtils.substituteSecrets(literalInput));
             systemTask.start(workflow, task, this);
         } finally {
--- a/core/src/test/java/com/netflix/conductor/core/execution/TestWorkflowExecutor.java
+++ b/core/src/test/java/com/netflix/conductor/core/execution/TestWorkflowExecutor.java
@@ -109,6 +109,11 @@ public class TestWorkflowExecutor {
             return new Wait();
         }
 
+        @Bean(TASK_TYPE_HUMAN)
+        public Human human() {
+            return new Human();
+        }
+
         @Bean("HTTP")
         public WorkflowSystemTask http() {
             return new WorkflowSystemTaskStub("HTTP") {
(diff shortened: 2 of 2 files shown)


nikhiln64 (commented):
This is a real gap and the fix is in the right place. A synchronous system task like HUMAN moves its own status inside start() during the decide loop and never goes through the updateTask(TaskResult) hand off, so its IN_PROGRESS transition was never reaching the listener. Wiring notifyTaskStatusListener into scheduleTask after the sync start closes that, and since notifyTaskStatusListener dispatches by current status, a task that completes immediately inside start() fires only the terminal event while an async task queued for later is left untouched. The three tests cover exactly those cases, including the async no notification path, which is the one that would have caught a double notify.

One thing to fix before this merges. In startSyncSystemTaskWithSecrets the change adds a second identical task.setInputData(parametersUtils.substituteSecrets(literalInput)) directly above the line that was already there, so the substitution now runs twice on the same literal input. It is harmless because the substitution is idempotent on the captured literalInput, but it reads like a rebase artifact and does the work twice for nothing. Dropping the added duplicate keeps the diff to just the notification change.

One small question on the SCHEDULED path. notifyTaskStatusListener treats SCHEDULED as a no op, so a sync task that returns from start() still SCHEDULED is correctly silent, but nothing pins that. If that state is reachable for any sync system task, one more case would stop a future change to the dispatch switch from leaking a spurious event.

nthmost-orkes (commented):
Thanks for tackling this -- the sync path fix reads well and the tests cover it nicely.

One thing though: the issue points at two spots that drop the notification, and this PR handles the first (the sync `scheduleTask` path) but not the second. The async side (`AsyncSystemTaskExecutor`, which is where WAIT and friends live) has the same gap and isn't touched here, so #1176 wouldn't be fully closed by merging this.

Would you be up for extending it to cover the async path too? If you'd rather keep this PR focused (totally reasonable), just swap `Fixes #1176` for `Part of #1176` so the issue stays open for a follow-up.

Small thing I also noticed: looks like the `substituteSecrets` line got duplicated in `startSyncSystemTaskWithSecrets` (called twice back to back) -- probably an accidental paste, worth dropping one.

nthmost-orkes:
Separate note so we don't lose track of it (not something you need to fix in this PR): while poking around I noticed the decide-loop `execute()` path (the one INLINE / SET_VARIABLE go through) also never notifies the listener. It's a different spot from the two in the issue, so I'll track it on its own -- just flagging it here for context.

PR #1288, opened 2026-07-14
Resolve #1286: Support conductor agents in the AGENT tasks

v1r3n:
Closes #1286

Automated resolution by the Conductor coding harness (code_parallel).

> Support conductor agents in the AGENT tasks

PR #1295, opened 2026-07-15
AGENT (conductor) task: fix error classification, resume validation, output.text, pull-wait resume, and input parsing

NicholasDCole:
Fixes five correctness bugs in #1288's `AGENT` (conductor) task branch, found during code review. Each fix is covered by unit tests that fail without it. Targets `harness/issue-1286` so the fixes land with the feature.

### 1. Permanent errors are retried instead of failing terminally
**Before:** `ConductorAgentDelegate` only treats the SDK's `NonRetryableException` as terminal — but nothing on this code path throws it. What actually gets thrown is `NotFoundException` (unknown `name`, purged execution) and `IllegalStateException` (nothing pending to resume), and both fall into the generic catch that marks the task retryable `FAILED`. A typo'd agent name burns every retry and, while polling, the poll-failure budget (30 polls × 5s) before giving up.
**After:** `start()`/`execute()` classify `NotFoundException`, `IllegalArgumentException`, and `IllegalStateException` as `FAILED_WITH_TERMINAL_ERROR`.

### 2. Resuming with a blank prompt crashes into a retry loop
**Before:** the resume path (`executionId` set) calls `Map.of("result", request.getPrompt())` without validating the prompt. A blank prompt → `Map.of` NPEs → generic catch → retryable `FAILED` → identical NPE on every retry.
**After:** blank prompt on resume fails terminally with `"AGENT (conductor) requires 'prompt'"`, same as the fresh-start validation. `respond()` is never reached.

### 3. `output.text` is always null
**Before:** `getStatus()` reads the `"text"` key from the completed workflow's output — but the agent compiler never emits `"text"`; it emits the canonical `"result"` key (every `setOutputParameters` site in `AgentCompiler`). So `${run_agent_ref.output.text}` in the examples/docs never resolves.
**After:** `extractText()` prefers an explicit `"text"` key, falls back to `"result"` when it's a String; schema-typed (Map) results stay in `output` only.

### 4. A `PULL_WORKFLOW_MESSAGES` wait can be seen but never answered
**Before:** `getStatus()` reports `waiting=true` for pending HUMAN **and** PULL_WORKFLOW_MESSAGES tasks, but `respond()` only looks for HUMAN and throws `IllegalStateException` otherwise. A parent workflow following the documented resume contract for a message-pull wait always fails (and, per bug 1, gets retried).
**After:** `respond()` — in both `ConductorAgentDelegate` and `AgentService` — answers each wait through the channel it actually consumes: HUMAN via `updateTask`, PULL_WORKFLOW_MESSAGES via a workflow-message-queue `push` + `decide` (mirroring `WorkflowMessageQueueResource`; the task pops its input from the queue, so completing it directly would be wrong). If the queue feature is disabled, the error names the correct API (`POST /api/workflow/{workflowId}/messages`). `AgentTask` injects `Optional<WorkflowMessageQueueDAO>` — a plain DAO with no `WorkflowExecutor` dependency, so this cannot recreate the startup bean cycle.

### 5. Malformed task input throws out of the system-task entry points
**Before:** all four `AgentTask` entry points call `parseRequest(task)` for branch dispatch before any error handling. Input like `agentVersion: "abc"` makes `convertValue` throw an `IllegalArgumentException` that escapes `start`/`execute`/`cancel`/`getEvaluationOffset` — verified red on the base branch (4 failing tests). The input was also Jackson-converted twice per call, including on every decider sweep via `getEvaluationOffset`.
**After:** each entry point parses once inside a guarded scope; malformed input → `FAILED_WITH_TERMINAL_ERROR` (`cancel()` logs and still cancels, never throws); `getEvaluationOffset()` dispatches off a raw `agentType` key read — no per-sweep Jackson conversion — and falls back to the default cadence on any error.

## Files

- `ai/.../ConductorAgentDelegate.java` — fixes 1–4 (guarded parse, resume validation, exception classification, `extractText`, dual-channel `respond`)
- `ai/.../AgentTask.java` — fix 5 + optional queue-DAO injection
- `agentspan/.../AgentService.java` — fix 4 on the REST path
- Tests: `ConductorAgentDelegateTest` (+8), `AgentTaskConductorBranchTest` (+4), `AgentServiceRespondTest` (new, 5) — hand-written fakes/JDK proxies, no mock frameworks per AGENTS.md

## Test evidence

- Fix 5's tests run **red against the unfixed base** (scratch worktree at `43a9a9769`): all 4 fail with the escaping `IllegalArgumentException`; green here.
- Fixes 1–4 assert against expressions that are byte-identical to code proven red in this PR's first iteration (pre-rearchitecture).
- Gate: `./gradlew :conductor-agentspan:test` (full module) and `:conductor-ai:test --tests "org.conductoross.conductor.ai.agent.*" --tests "org.conductoross.conductor.ai.a2a.*"` → BUILD SUCCESSFUL.

## History

This PR originally fixed 7 review findings against an earlier version of `harness/issue-1286`. The base branch was then rearchitected (runtime abstraction replaced by `WorkflowExecutor`-backed execution — which also resolved the CI-breaking startup circular reference). After merging the updated base, two findings were verified as superseded upstream (null identity overwrites; search-index-based idempotency — now a native core `idempotencyKey` + correlation lookup) and dropped; the five above were re-verified as still present and re-implemented for the new architecture.


NicholasDCole:
**CI fix (`9756d6fb2`):** the unit-test job was failing in `:conductor-core:compileTestJava` — not in this PR's code. The base branch (`c5f9b68ad`) added `ExecutorUtilsTest` cases for a four-argument, `SystemTaskRegistry`-aware `computePostpone(...)` (the mechanism that re-sweeps a workflow at a synchronous system task's `getEvaluationOffset` cadence — what the AGENT task's polling contract depends on), but the production overload itself was never committed.

This PR now implements that missing half per the tests' spec: a registry-aware `computePostpone` overload (sync system task's advertised offset taken as the minimum candidate; async system tasks ignored; 3-arg overload delegates unchanged) and wires `systemTaskRegistry` at the single production call site in `WorkflowExecutorOps`, so the AGENT task is actually re-polled at `pollIntervalSeconds` at runtime. Full `:conductor-core` suite green locally.

Issue #1308, opened 2026-07-16
Nested DO_WHILE: body task names collide when outer loop runs more than 1 iteration


nthmost-orkes:
## Summary

When an outer DO_WHILE runs **2 or more iterations** and its `loopOver` contains
an inner DO_WHILE, the inner loop's body tasks get the same names in outer
iteration 2 as they had in outer iteration 1. The workflow gets stuck in RUNNING
state and never completes.

## Details

Conductor appends a single iteration suffix (`__N`) to tasks inside a DO_WHILE.
With nesting, this produces:

| Outer iter | Inner iter | Body task name |
|-----------|-----------|---------------|
| 1 | 1 | `body__1` |
| 1 | 2 | `body__2` |
| 2 | 1 | `body__1` ← **collision** |
| 2 | 2 | `body__2` ← **collision** |

The naming scheme does not compose iteration suffixes across nesting levels. A
correct scheme would produce compound names (e.g. `body__1__1`, `body__1__2`,
`body__2__1`, `body__2__2`) or an equivalent that guarantees uniqueness.

## Reproduction

Register a workflow with outer DO_WHILE `loopCondition` that runs 2 or more
iterations, with an inner DO_WHILE in its `loopOver`. Start it. After the first
outer iteration completes, the workflow will stall on the second outer iteration
— the inner body tasks already exist in the task list and cannot be re-scheduled.

Reproduction workflow: `ks_combo_do_while_do_while` in `conductor-oss/conductor-test-harness`.
Workaround in the battery: limit outer loop to 1 iteration.

## Files

- `core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java`
- Anywhere `appendIteration` / iteration-suffix logic lives that handles nested loops

## Related

Companion to #1307 (the loopCondition TypeError for nested DO_WHILE). Both surface
together when using nested DO_WHILE. The TypeError in #1307 masks this bug because
the inner loop never gets far enough to collide.

nthmost-orkes:
Discovered against **Conductor 3.32.0-rc.9** running on JDK 21 (loki.local).

nthmost:
@manan164 @v1r3n I realize we don't support nested DO_WHILE loops yet -- do we want to handle this in some kinder way to the user until we do?

mp-orkes:
@nthmost  nested `DO_WHILE` is a documented limitation (use a Sub Workflow). Same shape as #1256/#1307. 
+1 on failing fast at registration.

PR #1312, opened 2026-07-16
fix: strip iteration suffix from DO_WHILE task ref in loopCondition context

### LINKED ISSUE #1307: Nested DO_WHILE: loopCondition TypeError when inner DO_WHILE runs inside outer DO_WHILE
## Summary

When a DO_WHILE task is nested inside another DO_WHILE's `loopOver`, the inner
loop fails immediately on its first iteration evaluation with a TypeError:

```
Unable to evaluate condition if ($.innerRef['iteration'] < $.maxIter) ...
TypeError: Cannot read property 'iteration' of undefined
```

## Root Cause

`DoWhile.java::evaluateCondition()` builds the condition input map by calling
`conditionInput.put(task.getReferenceTaskName(), task.getOutputData())` (line 518).

For a nested DO_WHILE task, `getReferenceTaskName()` returns the name *with* the
outer loop's iteration suffix appended (e.g. `dwdw_inner__1`). The `loopCondition`
expression references the bare name `$.dwdw_inner['iteration']` — without the
suffix — so the lookup returns `undefined` and the expression throws.

The loop **body** tasks on lines 534-536 already call
`TaskUtils.removeIterationFromTaskRefName()` to strip the suffix correctly.
The self-reference on line 518 is the only place where stripping was missed.

## Fix

```java
// DoWhile.java line 518 — before
conditionInput.put(task.getReferenceTaskName(), task.getOutputData());

// after
conditionInput.put(
    TaskUtils.removeIterationFromTaskRefName(task.getReferenceTaskName()),
    task.getOutputData());
```

## Reproduction

Register a workflow with an outer DO_WHILE (any `loopCondition`) whose `loopOver`
contains an inner DO_WHILE. Start the workflow. The inner loop will throw the
TypeError on the first call to `evaluateCondition()` for the inner task.

Minimal reproduction definition available in the kitchen-sink battery at
`conductor-oss/conductor-test-harness` — workflow `ks_combo_do_while_do_while`.

## Files

`core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java`

nthmost-orkes:
Fixes #1307.

## Problem

When a `DO_WHILE` task is nested inside another `DO_WHILE`, the inner loop's
`loopCondition` evaluation throws:

```
TypeError: Cannot read property 'iteration' of undefined
```

`DoWhile.java::evaluateCondition()` builds the condition-input map by calling:

```java
conditionInput.put(task.getReferenceTaskName(), task.getOutputData());
```

At runtime, a nested DO_WHILE task has its outer iteration suffix appended to its
reference name (e.g. `inner__1` instead of `inner`). The `loopCondition` expression
references the bare name (`$.inner['iteration']`), so the lookup returns `undefined`
and the expression throws.

## Fix

Apply `TaskUtils.removeIterationFromTaskRefName()` — the same helper already used
for loop body tasks in lines 534-536 of the same method — to the self-entry:

```java
// Before
conditionInput.put(task.getReferenceTaskName(), task.getOutputData());

// After
conditionInput.put(
    TaskUtils.removeIterationFromTaskRefName(task.getReferenceTaskName()),
    task.getOutputData());
```

One-line change, no new imports (TaskUtils is already imported).

## User impact

First-time and experienced users writing nested DO_WHILE workflows would hit an
opaque TypeError and have no path to fix it (the `loopCondition` syntax was
correct; the engine was at fault). After this fix, nested DO_WHILE loops with
correct `loopCondition` expressions evaluate cleanly.

## Test

Reproducible with the `ks_combo_do_while_do_while` workflow in
[conductor-oss/conductor-test-harness](https://github.com/nthmost-orkes/conductor-test-harness)
kitchen-sink battery. With this fix deployed, the nested DO_WHILE runs to
COMPLETED correctly (outer=1 iteration).

Note: A companion bug (#1308) exists where outer>1 iterations cause body-task
naming collisions. That is a separate, deeper issue.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/tasks/DoWhile.java
@@ -515,7 +515,12 @@ public class DoWhile extends WorkflowSystemTask {
                         workflow,
                         task.getTaskId(),
                         taskDefinition);
-        conditionInput.put(task.getReferenceTaskName(), task.getOutputData());
+        // Strip iteration suffix (e.g. "dwdw_inner__1" → "dwdw_inner") so that
+        // $.taskRef in the loopCondition resolves correctly when this DO_WHILE is
+        // nested inside another DO_WHILE whose iteration suffix is appended at runtime.
+        conditionInput.put(
+                TaskUtils.removeIterationFromTaskRefName(task.getReferenceTaskName()),
+                task.getOutputData());
         List<TaskModel> loopOver =
                 workflow.getTasks().stream()
                         .filter(


nthmost-orkes:
Discovered and tested against **Conductor 3.32.0-rc.9**, JDK 21.

manan164 (changes requested):
Hi @nthmost-orkes , we dont support do_while inside another do_while loop fully yet.

PR #1314, opened 2026-07-16
Add long-term (OCG-backed) agent memory compilation to agentspan-server

NicholasDCole:
## Overview

Adds OCG-backed long-term memory for AgentSpan agents, plus the supporting MCP and execution UI work.

## OCG memory

- Configure an agent with an OCG URL, server-side credential name, agent identity, optional user, and visibility.
- Recall relevant memory before model work begins. Recalled content is bounded and treated as supporting evidence, not instructions.
- Capture completed or terminated root runs in an observable `ocg_memory_capture` workflow. OCG handles folding, summaries, ranking, and retention.
- Show a completed run’s memory summary in the execution page and let users mark it helpful or not helpful with a reason.
- Keep credentials on the server. Child runs do not create their own capture or feedback records.

## MCP and execution UX

- Honor configured MCP tool allowlists and compile known OCG graph tools without an unnecessary discovery call.
- Preserve tool identity through dynamic execution so results return to the right model turn.
- Show clearer MCP labels and aggregate token counts across agent sub-workflows.
- Load nested execution details in one server request instead of recursively from the browser.

The screenshots below show the end-to-end experience.


agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/OcgLongTermMemoryCompiler.java
```diff
@@ -0,0 +1,294 @@
...
+    private static WorkflowTask optionalTask(String type, String referenceName) {
+        WorkflowTask task = new WorkflowTask();
+        task.setName(type);
+        task.setType(type);
+        task.setTaskReferenceName(referenceName);
+        task.setOptional(true);
+        return task;
+    }
+
+    private static String recallArgumentsScript() {
```
kowser-orkes: Can we have static variables constants like these?

PR #1335, opened 2026-07-16
Fix duplicate async system task execution (#1321), late terminal-task overwrite (#1322), and output guardrail misfire on tool-call turns (#1323)

### LINKED ISSUE #1321: Async system tasks running longer than the queue unack window are executed twice (duplicate LLM_CHAT_COMPLETE provider calls)
## Summary

## Mechanism

## Evidence

```
```

|---|---|---|---|

## Repro

## Suggested fix

```java
```

### LINKED ISSUE #1322: Late completion of a duplicate system task attempt overwrites outputData/endTime of an already-terminal task
## Summary

This corrupts two things:

## Evidence

```
```

```
user:      (context)
```

## Suggested fix

### LINKED ISSUE #1323: AgentSpan: output guardrails misfire on tool-call turns (output.result is [] mid-tool-use) and inject retry messages into the conversation
## Summary

AgentSpan output-position guardrails are evaluated on **every** loop iteration, including legitimate tool-call turns. On a tool-call turn `${<agent>_llm.output.result}` is the empty list `[]` (only non-tool turns produce a result Map), so any non-empty-content guardrail (e.g. a regex `allow` pattern like `\w`) **fails on every tool round**, and the compiled `retry` branch injects its correction message into the conversation while the model is correctly mid-tool-use.

## Effect

With a simple "reply must be non-empty" output guardrail (`guardrailType: regex, mode: allow, patterns: ["\\w"], onFail: retry`) on an agent with one worker tool:

- Every tool-call turn fails the guardrail (`passed=false`, retry) even though the turn is valid; only the final prose/JSON turn passes.
- The retry message ("your reply came back empty, write your final answer now...") is appended as a `user` message alongside the tool results, so the model receives contradictory pressure each round: *"answer NOW"* vs. its instructions to use the tool first. Observed result: the model starts emitting junk hedge tool calls (`name_query: "dummy"`, `name_query: "placeholder"`) to satisfy both, adding wasted rounds, tokens and latency.
- Guardrail `maxRetries` budget is silently consumed by healthy tool turns.

Observed on an agent execution where 3 of 4 turns were tool rounds: the guardrail retry branch ran 3× and each of those turns carried an injected "[Output validation failed: ...]" user message (visible in the next turn's `messages` input).

```
regex_guardrail__1..3:  passed=false, on_fail=retry    (content was [] — tool-call turns)
regex_guardrail__4:     passed=true                    (content was the final reply Map)
```

## Where

- `GuardrailCompiler` binds the guardrail INLINE's `content` input to the LLM task's output result ref, with no awareness of tool turns:
  https://github.com/conductor-oss/conductor/blob/c2439b7e6b10f618e1ba73e78369b96a08373263/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/GuardrailCompiler.java#L177
- The generated script (`JavaScriptBuilder.regexGuardrailScript`) evaluates the pattern against that content unconditionally.

## Suggested fix

An output guardrail should only judge **final replies** — a tool-call turn is legally empty:

1. In `GuardrailCompiler.compileRegexGuardrail` (and the other output-position variants), additionally bind `toolCalls: ${<agent>_llm.output.toolCalls}` into the guardrail task's inputs.
2. In `JavaScriptBuilder.regexGuardrailScript`, short-circuit before pattern evaluation:

```javascript
if ($.toolCalls && $.toolCalls.length > 0) {
  return { passed: true, on_fail: 'pass', should_continue: false, guardrail_name: ..., message: '' };
}
```

Guardrail behavior on genuine final turns (empty/thinking-only replies, which such guardrails exist to catch) stays unchanged.

Related: #1321 / #1322 amplified this in the observed run (a zombie task overwrite duplicated the injected retry messages in the rebuilt history), but the misfire itself reproduces independently with any output guardrail + any tool-using agent.

NicholasDCole:
Fixes #1321
Fixes #1322
Fixes #1323

## Summary

Three related defects observed in a single agent-loop execution (all four LLM turns billed twice, the losing duplicate attempt corrupting the stored task records, and output guardrails injecting retry messages on every healthy tool round). #1321 is fixed with a small, general **execution-lease mechanism**: a system task that blocks declares how long its invocation may run, and `AsyncSystemTaskExecutor` extends the queue message's visibility by that amount before invoking it. Unit tests fail without the fixes, and the integration repro requires **no real LLM calls**.

Design docs in this PR: [chat-turn execution sequence (vanilla vs fixed, mermaid)](docs/design/2026-07-17-agentspan-chat-turn-sequence.md) and [the fix options considered](docs/design/2026-07-17-issue-1321-fix-options.md).

### #1321 — async system tasks running longer than the queue redelivery window are executed twice
Annotated system-task workers (`LLM_CHAT_COMPLETE` et al.) run synchronously inside `AnnotatedWorkflowSystemTask.start()` — for an LLM turn this blocks 60–150s on the provider HTTP call — while the task row stays `SCHEDULED` and nothing renews the queue message's lease. Every QueueDAO redelivers popped-but-unacked messages (redis: 30s; sqlite/postgres/mysql: ~60s, hardcoded), so a second system-task-worker pops the same message mid-execution and invokes the same method again — a duplicate paid provider call.

**Fix — an execution-lease hook (Option A of the options doc):**
- `WorkflowSystemTask.getExecutionLease(TaskModel)` — new overridable method, default `Optional.empty()`. A task that blocks declares the duration its invocation may legally run.
- `AsyncSystemTaskExecutor.extendLease(...)` — before `start()`/`execute()`, extends the message's visibility by the declared lease via `queueDAO.setUnackTimeout` (the same per-message visibility idiom core already uses for the decider queue in `WorkflowExecutorOps`/`WorkflowSweeper`). Lease failures log a warning and never block execution.
- `AnnotatedWorkflowSystemTask` declares `responseTimeoutSeconds` from its task def (3600s for LLM task defs). Tasks without a task def or with `responseTimeoutSeconds == 0` declare nothing.
- `HttpTask` declares its configured `connectionTimeOut + readTimeOut` when that exceeds 20s (safely under the tightest redelivery window, redis's 30s) — the same bug class applies to long-timeout HTTP calls; default-configured HTTP tasks (6s) declare nothing.

**Performance: zero cost for tasks that don't opt in.** The default hook returns the `Optional.empty()` singleton — no allocation, no I/O, no queue write, behavior byte-identical to before. Only blocking invocations that declare a lease pay one `setUnackTimeout` write, amortized against a call that runs seconds to minutes.

### #1322 — late completion of a duplicate attempt overwrites an already-terminal task
The losing attempt's late completion overwrote `outputData`/`endTime`/`finishReason` of a task the workflow had already consumed — recorded `endTime`s landed after the workflow's own `endTime`, and agent-loop conversations rebuilt from the corrupted records contained phantom assistant replies.

**Resolution: fixed at the root, no fallback guard.** The overwrite is only reachable through a duplicate attempt, which #1321's fix eliminates. Task updates deliberately stay last-write-wins: legitimate engine flows update terminal tasks terminal→terminal (the decider converting an optional `FAILED` task to `COMPLETED_WITH_ERRORS`, sub-workflow retry/restart/resume syncing the parent `SUB_WORKFLOW` task `FAILED`→`COMPLETED`) — a guard in the update path broke exactly those flows in the full test-harness suite during development, and regression pin tests in `ExecutionDAOFacadeTest` now document that last-write-wins behavior.

### #1323 — AgentSpan output guardrails misfire on tool-call turns
Output-position guardrails evaluate every loop iteration, but on a tool-call turn the LLM task's `output.result` is `[]`, so any non-empty-content guardrail (e.g. regex allow `\w`, onFail retry) failed every healthy tool round and injected its retry message into the conversation mid-tool-use — contradictory pressure that produced junk hedge tool calls and silently burned the `maxRetries` budget.

**Fix:** `GuardrailCompiler` binds a `toolCalls` input (from the agent LLM task's `output.toolCalls`) into all four output-position guardrail variants (regex/llm/custom/external), and the generated scripts in `JavaScriptBuilder` short-circuit to a pass before content evaluation when `toolCalls` is a non-empty array. Tool-level guardrails don't bind `toolCalls` (they must judge tool calls), and genuine final turns — including genuinely empty replies these guardrails exist to catch — behave exactly as before.

## Tests

**Unit (regression) tests** — each fails on the unfixed code:
- `AsyncSystemTaskExecutorTest`: the declared lease is applied via `setUnackTimeout` BEFORE the blocking `start()` (ordering asserted); no queue write when no lease is declared; execution proceeds when the lease write throws.
- `TestAnnotatedWorkflowSystemTask`: `getExecutionLease` returns `responseTimeoutSeconds` from the task def; empty without a task def or with a zero timeout.
- `HttpTaskUnitTest`: no lease for default timeouts (zero-write hot path), lease equals configured `connectionTimeOut + readTimeOut` for long calls, no lease for malformed input.
- `ExecutionDAOFacadeTest`: pins that task updates stay last-write-wins for non-terminal-stored updates and rerun-style terminal→`SCHEDULED` resets.
- `GuardrailCompilerTest` / `Issue1323GuardrailToolCallBindingTest`: output guardrails bind `toolCalls`.
- `GuardrailToolCallShortCircuitScriptTest` / `Issue1323GuardrailToolCallMisfireTest`: the real generated scripts, evaluated in GraalJS — tool-call turn passes; empty final reply still fails.

**Integration repro (test-harness, deterministic, no LLM/network):**
- `Issue1321DuplicateAsyncSystemTaskSpec`: a latch-controlled `@WorkerTask` bean (`ControllableWorker`) registered through the **real** `WorkerTaskAnnotationScanner`, executed via the **real** `AsyncSystemTaskExecutor` against the **real** queue; the redelivery window is compressed to ~1s and a second worker polls after it elapses. Asserts the operation runs exactly once. Passes unchanged against the relocated fix.

## Evidence

Pre-fix failures (proving the bugs):

```
#1321  Issue1321DuplicateAsyncSystemTaskSpec:
       controllableWorker.invocations.get() == 1   ->  actual: 2  (worker method ran twice)
#1321  AsyncSystemTaskExecutorTest: no setUnackTimeout call before the blocking invocation
#1323  guardrail on tool-call turn: passed == false, should_continue == true  ->  expected passed: true
#1323  compiled INLINE task inputs missing toolCalls binding
```

Post-fix verification (this branch, re-run after the lease refactor):
- `:conductor-core:test` — annotated-task, `AsyncSystemTaskExecutorTest`, and `ExecutionDAOFacadeTest` suites: 0 failures
- `:conductor-http-task:test` — full module: 0 failures
- `:conductor-test-harness:test --tests Issue1321DuplicateAsyncSystemTaskSpec` — passes
- `:conductor-agentspan:test` — 544 tests, 0 failures (guardrail fix unchanged by the refactor)

## Core-surface note for reviewers

Unlike an earlier revision of this branch, this PR **does** touch two core orchestration files, deliberately and minimally (see the [options doc](docs/design/2026-07-17-issue-1321-fix-options.md) for the alternatives considered):
- `WorkflowSystemTask` — one new overridable method with an inert default.
- `AsyncSystemTaskExecutor` — one guarded `extendLease` call before task invocation.

Placing the queue write in the executor (which already owns the message's `postpone`/`remove` lifecycle) instead of inside one task class means any blocking system task — annotated workers today, `HttpTask` with raised timeouts, future ones — gets duplicate-execution protection by overriding one method, and no task pays anything unless it opts in.

Known follow-up (out of scope here): multi-agent-compiled `LLM_CHAT_COMPLETE` tasks carry no embedded task def, resolve to an ad-hoc def with `responseTimeoutSeconds=0`, and therefore declare no lease; the compilers should set `responseTimeoutSeconds` explicitly (or the scanner should register task defs for `@WorkerTask` types).


NicholasDCole:
Reworked per team feedback (core orchestration must stay unchanged — `WorkflowExecutor` / `AsyncSystemTaskExecutor` / `DeciderService`):

- **Reverted** both prior core edits; `core/src/main/java/com/netflix/conductor/**` is now byte-identical to `main`.
- **#1321** fix relocated to `AnnotatedWorkflowSystemTask` — the org.conductoross adapter every annotated system-task worker (LLM_CHAT_COMPLETE et al.) runs through, i.e. the component that actually blocks without managing its queue lease. Its `start()` now extends the message's visibility to the task's `responseTimeoutSeconds` before invoking the worker method; `WorkerTaskAnnotationScanner` supplies the `QueueDAO`.
- **#1322** fallback guard dropped entirely: the terminal-record overwrite is only reachable via a duplicate attempt, which the #1321 fix eliminates at the root. (During development a guard in the shared update path broke legitimate terminal→terminal flows — optional-task conversion and sub-workflow retry sync — which confirmed last-write-wins is intentional there; pin tests now document it.)
- Fix-sensitivity re-verified at the new location: with only the adapter change reverted, `Issue1321DuplicateAsyncSystemTaskSpec` fails with `invocations = 2` (duplicate execution); with it, exactly once. Full suites green (core 853, agentspan 544, test-harness 279 with only the two `AIReasoningEndToEndTest` failures that already fail on `main`).

PR #1340, opened 2026-07-17
fix(core): persist start marker before async system task execution to stop sweeper repair race

michal-treter-text:
Queue message is acked before AsyncSystemTaskExecutor runs the side effect (HTTP call etc.), and task row only updates after it returns. During that window the task looks orphaned (SCHEDULED, no queue message), so the sweeper repushes it and a second worker re-executes it. Persist startTime before start() runs, and have the sweeper skip repair while startTime is within a grace window (SweeperProperties#startedTaskRepairGraceMillis, default 60s, 0 disables).

Pull Request type
----
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Queue message is acked before `AsyncSystemTaskExecutor` runs the side effect (HTTP call etc.), and task row only updates after it returns. During that window the task looks orphaned (SCHEDULED, no queue message), so the sweeper repushes it and a second worker re-executes it.

This PR persists `startTime` before `start()` runs, and has the sweeper skip repair while `startTime` is within a grace window (`SweeperProperties#startedTaskRepairGraceMillis`, default 60s, 0 disables).

Issue #

Alternatives considered
----

Want it pushed to PR description via gh, or just keep as draft file?

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -152,6 +152,16 @@ public class AsyncSystemTaskExecutor {
                 task.incrementPollCount();
             }
 
+            if (task.getStatus() == TaskModel.Status.SCHEDULED) {
+                task.setStartTime(System.currentTimeMillis());
+                Monitors.recordQueueWaitTime(task.getTaskType(), task.getQueueWaitTime());
+                // Persist the start marker before running the task. The queue message is already
+                // acked, so until this write the task is indistinguishable from an orphaned one;
+                // the sweeper would repush it mid-execution and the task would run twice. Must
+                // happen before secrets substitution so resolved secrets are never persisted.
+                persistStartMarkerQuietly(task);
+            }
+
             if (task.getStatus() == TaskModel.Status.SCHEDULED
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -222,6 +230,19 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
         }
     }
 
+    private void persistStartMarkerQuietly(TaskModel task) {
+        try {
+            executionDAOFacade.updateTask(task);
+        } catch (Exception e) {
```
v1r3n: don't catch exception if it fails it will be handled by upstream

mp-orkes:
The startTime + grace-window approach is a mitigation rather than a full fix: the grace is a fixed global window (60s), but the tasks that trigger #1321 in practice are multi-minute LLM/HTTP runs. Those still look orphaned once they outlast the grace, so the sweeper repushes and the duplicate comes back. 

We're most likely going with #1369, which fixes it structurally: the message is never acked at poll, so it stays present (`containsMessage == true`) for the whole run and repair never treats an in-flight task as orphaned. 

@michal-treter-text really appreciate you digging into this. Let's keep collaborating on the fix!

PR #1341, opened 2026-07-17
feat(a2a): backward compat with A2A v0.2 clients (tasks/send, /rpc alias)

shaileshpadave:
The `a2a-cli` on PyPI targets A2A v0.2, but Conductor implements v0.3. Two things don't match:

- The CLI posts to `/a2a/{workflow}/rpc` — Conductor only had the root path
- The CLI sends `tasks/send` / `tasks/sendSubscribe` — Conductor expects `message/send` / `message/stream`

This PR adds the missing aliases on the Conductor side so v0.2 clients work out of the box without any config.

**On the a2a-cli side**, a few bugs also need to be fixed upstream (separate PR to the a2a-cli repo):
- `display_task_info` called with positional `color` arg instead of keyword
- `TaskState` enum not JSON serializable in `--json` mode
- Artifacts deserialized as plain dicts but display code expects Pydantic objects
- SSE events need `taskId` → `id` mapping and JSON-RPC result unwrapping to match v0.2 event shape


ai/src/main/java/org/conductoross/conductor/ai/a2a/server/A2AServerResource.java
```diff
@@ -115,16 +118,18 @@ public Object jsonRpc(
...
             JsonNode params = request.get("params");
-            if ("message/stream".equals(method)) {
+            // message/stream and its v0.2 alias tasks/sendSubscribe both return SSE
+            if ("message/stream".equals(method) || "tasks/sendSubscribe".equals(method)) {
                 return streamResponse(workflow, params, id);
             }
             Object result;
             switch (method) {
                 case "message/send":
+                case "tasks/send": // A2A v0.2 backward compat alias
```
v1r3n, line 128: No test covers the new v0.2 aliases (`tasks/send`, `tasks/sendSubscribe`) or the `/rpc` path suffix. `A2AServerResourceTest` already exercises `message/send`/`message/stream` and would be the natural place to assert the aliases route to the same handlers and that the metric is recorded as `message/send`. Adding those cases would lock in this backward-compat behavior.

v1r3n (commented):
This PR adds A2A v0.2 backward-compat method aliases (`tasks/send` → `message/send`, `tasks/sendSubscribe` → `message/stream`) and an alternate `/rpc` suffix on the JSON-RPC endpoint.

The change is correct and minimal:
- `tasks/send` reuses `parseMessage`, which expects `params.message` — present in both v0.3 `message/send` and v0.2 `TaskSendParams`, so parsing works.
- The metric tag is correctly hardcoded to `"message/send"` for the aliased case, preserving the bounded-tag invariant noted in the surrounding comment.

Main gap: no test exercises the new aliases or the `/rpc` path variant, despite `A2AServerResourceTest` being the natural home for them and AGENTS.md asking for tests on new behavior.

v1r3n:
Addressed the review feedback on this PR.

- Engine: `code_parallel`
- Branch: `feature/a2a-v2-compat`

See the latest commits for the changes.

v1r3n (approved):

PR #1346, opened 2026-07-17
fix(ui-next): Don't run playwright in Build UI CI

### LINKED ISSUE #1344: fix(ui-next): stale Playwright snapshot blocks CI on all PRs (execution-search-results-completed-desktop.png)
## Summary

The `build-ui` CI check is failing on all open PRs due to a stale Playwright screenshot snapshot. The snapshot for the workflow execution search results page no longer matches the current UI.

## Failing test

**File:** `ui-next/e2e/workflow-execution-search-filters.spec.ts`
**Test:** `Should match search results after clicking search`
**Snapshot:** `execution-search-results-completed-desktop.png`

## Root cause

Commit `da6e07680` (PR #1216, merged 2026-06-29) changed the visual layout of the Executions search page but did not regenerate the Playwright snapshots. The stored baseline no longer matches what the page renders.

## Impact

Every PR targeting `main` fails `build-ui` CI, including unrelated PRs like #907 and #1205.

## Fix

Regenerate the snapshot by running in `ui-next/`:

```bash
npx playwright test workflow-execution-search-filters --update-snapshots
```

Then commit the updated PNG files.

### LINKED ISSUE #1345: bug(ui-next): maxDiffPixels: 1500 in snapshot config causes spurious CI failures on native runner
## Summary

The `build-ui` CI job fails on PRs that don't touch the UI at all (e.g. #907, #1205) because the Playwright snapshot config in `workflow-execution-search-filters.spec.ts` uses a hard `maxDiffPixels: 1500` cap that is too tight for the rendering difference between the Docker snapshot environment and GitHub Actions' native Ubuntu runner.

## Root cause

Commit `da6e07680` (PR #1216) introduced `workflow-execution-search-filters.spec.ts` with this config:

```ts
const SCREENSHOT_CONFIG = {
  maxDiffPixelRatio: 0.03,
  maxDiffPixels: 1500,
};
```

The baseline snapshots were generated inside Docker (`mcr.microsoft.com/playwright:v1.60.0-noble`). The CI `build-ui` job runs Playwright **natively** on `ubuntu-latest` via `pnpm test:e2e`, which uses a different Chromium build and renders the page with ~2140 pixels of difference vs. the Docker baseline.

Playwright applies **both** thresholds and fails if either is exceeded. Because `maxDiffPixels: 1500 < 2140`, the test always fails on the native runner — even though the `maxDiffPixelRatio: 0.03` (3%) threshold would pass comfortably (ratio ≈ 0.01).

## Why two environments diverge

- **Snapshot generation:** `pnpm test:e2e:snapshots:update` → Docker container (`mcr.microsoft.com/playwright:v1.60.0-noble`, pinned Chromium)
- **CI validation:** `pnpm test:e2e` → native `ubuntu-latest` runner, Playwright installed from `pnpm-lock.yaml` (`^1.59.1`)

Minor font-rendering and sub-pixel differences between the two Chromium builds accumulate to just over 1500 pixels on the desktop-sized screenshot.

## Impact

Every PR targeting `main` fails `build-ui`, regardless of whether it touches the UI. Confirmed on #907 and #1205 (both fail with exactly 2140 pixels differing, all 3 retries).

## Fix options

**Option A (quick):** Remove `maxDiffPixels` from `SCREENSHOT_CONFIG` and rely solely on `maxDiffPixelRatio: 0.03`. A ratio-only threshold is environment-agnostic and is the correct tool for cross-environment tolerance.

```ts
const SCREENSHOT_CONFIG = {
  maxDiffPixelRatio: 0.03,
};
```

**Option B (correct long-term):** Change the `build-ui` CI job to run snapshot tests via Docker (`pnpm test:e2e:snapshots`) instead of natively, matching the environment the baselines were generated in. This guarantees pixel-identical comparisons but adds Docker setup overhead to the main CI job.

Option A is the minimal fix to unblock PRs. Option B would make the setup more coherent.

## Files

- `ui-next/e2e/workflow-execution-search-filters.spec.ts` — `SCREENSHOT_CONFIG` constant (line 18–21)
- `.github/workflows/ci.yml` — `build-ui` job runs `pnpm test:e2e` natively
- `ui-next/docker-compose.snapshots.yml` — Docker environment used for snapshot generation

nthmost-orkes:
## Summary

Removes Playwright E2E tests from the `build-ui` job in `ci.yml`.

Fixes #1344, fixes #1345.

## Why

`ui-next-ci.yml` (`E2E Mocked` job) and `ui-next-integration-ci.yml` already run Playwright tests — in Docker — whenever `ui-next/**` files change on a PR targeting `main` or a push to `main`. Running them again natively in `build-ui` was redundant and caused spurious failures due to rendering differences between the Docker snapshot environment and the native GitHub Actions runner.

Note: if UI tests should run on _any_ push to `main` (not just when `ui-next/**` changes), the `paths` filters in `ui-next-ci.yml` and `ui-next-integration-ci.yml` would need to be removed. That's a separate decision.

## Change

Deleted from `build-ui` in `ci.yml`:
- Cache Playwright browsers
- Install Playwright browsers
- Install Playwright browser deps (cached)
- Run Playwright E2E Tests
- Upload Playwright report

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -405,28 +405,16 @@ jobs:
       - name: Build UI
         run: NODE_OPTIONS=--max-old-space-size=4096 pnpm build
 
-      - name: Cache Playwright browsers
-        uses: actions/cache@v5
-        id: playwright-cache
-        with:
-          path: ~/.cache/ms-playwright
-          key: playwright-chromium-${{ hashFiles('ui-next/pnpm-lock.yaml') }}
-
-      - name: Install Playwright browsers
-        if: steps.playwright-cache.outputs.cache-hit != 'true'
-        run: pnpm exec playwright install --with-deps chromium
-
-      - name: Install Playwright browser deps (cached)
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: pnpm exec playwright install-deps chromium
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
 
       - name: Run Playwright E2E Tests
-        run: pnpm test:e2e
+        run: docker compose -f docker-compose.snapshots.yml run --rm playwright
 
       - name: Upload Playwright report
         uses: actions/upload-artifact@v7
         if: failure()
         with:
           name: playwright-report
-          path: ui-next/playwright-report
+          path: ui-next/playwright-snapshots-report/
           retention-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -404,17 +404,3 @@ jobs:
 
       - name: Build UI
         run: NODE_OPTIONS=--max-old-space-size=4096 pnpm build
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
-
-      - name: Run Playwright E2E Tests
-        run: docker compose -f docker-compose.snapshots.yml run --rm playwright
-
-      - name: Upload Playwright report
-        uses: actions/upload-artifact@v7
-        if: failure()
-        with:
-          name: playwright-report
-          path: ui-next/playwright-snapshots-report/
-          retention-days: 7
--- a/ui-next/e2e/workflow-execution-search-filters.spec.ts
+++ b/ui-next/e2e/workflow-execution-search-filters.spec.ts
@@ -17,6 +17,7 @@ import { mockCommonApis } from "./helpers/mockApi";
 
 const SCREENSHOT_CONFIG = {
   maxDiffPixelRatio: 0.03,
+  maxDiffPixels: 1500,
 };
 
 const VIEWPORTS = [
--- a/ui-next/e2e/workflow-execution-search-filters.spec.ts
+++ b/ui-next/e2e/workflow-execution-search-filters.spec.ts
@@ -17,7 +17,6 @@ import { mockCommonApis } from "./helpers/mockApi";
 
 const SCREENSHOT_CONFIG = {
   maxDiffPixelRatio: 0.03,
-  maxDiffPixels: 1500,
 };
 
 const VIEWPORTS = [


ajlai-orkes:
We should not be running playwright natively to prevent this specific issue. The playwright tests should always run in docker to guarantee consistency

ajlai-orkes (dismissed):
build-ui should only be building the UI and not running any playwright tests

ui-next/e2e/workflow-execution-search-filters.spec.ts
```diff
@@ -17,7 +17,6 @@ import { mockCommonApis } from "./helpers/mockApi";
 
 const SCREENSHOT_CONFIG = {
   maxDiffPixelRatio: 0.03,
-  maxDiffPixels: 1500,
```
ajlai-orkes, line 20: revert this change?

.github/workflows/ci.yml
```diff
@@ -405,28 +405,16 @@ jobs:
...
-        if: steps.playwright-cache.outputs.cache-hit != 'true'
-        run: pnpm exec playwright install --with-deps chromium
-
-      - name: Install Playwright browser deps (cached)
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: pnpm exec playwright install-deps chromium
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
 
       - name: Run Playwright E2E Tests
```
ajlai-orkes: we can remove playwright from the `build ui` CI completely as `ui-next-ci` and `ui-next-integration-ci` run when there are ui-next/** changes on:
- a PR targeting main, or
- a push to main

if we want to run on ANY push to main and not just changes to ui-next/** then we need to update each ci.yml file

nthmost-orkes: Ahhh i see.

ajlai-orkes (approved):

v1r3n (approved):

PR #1349, opened 2026-07-17
Fix #1321/#1322: annotated system tasks persist IN_PROGRESS before blocking invocations; redelivered executions do nothing

### LINKED ISSUE #1321: Async system tasks running longer than the queue unack window are executed twice (duplicate LLM_CHAT_COMPLETE provider calls)

### LINKED ISSUE #1322: Late completion of a duplicate system task attempt overwrites outputData/endTime of an already-terminal task

NicholasDCole:
Fixes #1321
Fixes #1322

## Summary

Minimal fix for duplicate execution of blocking annotated system tasks, expressed as the plain **system-task status contract** (per review feedback — no custom markers):

- **The system task never touches persistence** (per review). `AnnotatedWorkflowSystemTask` only *declares* `isBlockingStart()` and follows the status contract: `execute()` on an `IN_PROGRESS` task does nothing — except a worker-requested re-execution (`IN_PROGRESS + callbackAfterSeconds > 0`), preserving the long-running LLM/A2A worker flow (`LLMWorkers.java:79`, `A2AWorkers.java:481`).
- **The caller performs the hand-off, through `WorkflowExecutor`**: before invoking a blocking `start()`, `AsyncSystemTaskExecutor` calls `workflowExecutor.updateTask(TaskResult(IN_PROGRESS, callbackAfterSeconds=lease))` — the exact contract a remote worker uses before long work. That one sanctioned call persists the `IN_PROGRESS` transition (a redelivered queue message then sees the in-flight status instead of `SCHEDULED`) *and* postpones the queue message past the blocking call (`responseTimeoutSeconds`, or 3600s default), so redelivery normally never fires at all. The `execute()` guard remains as defense-in-depth.

Production change: `WorkflowSystemTask.isBlockingStart()` (new, default false), one guarded block in `AsyncSystemTaskExecutor`, and the adapter/status-contract changes in `AnnotatedWorkflowSystemTask`. Tasks that don't declare `isBlockingStart()` are byte-identical to today.

## Why a pre-invocation hand-off must exist somewhere

For async system tasks, nothing persists between pop and completion: `AsyncSystemTaskExecutor` writes the task only in its `finally`, *after* the blocking call returns, so on redelivery the row still reads `SCHEDULED` and any status check sees nothing (demonstrated below — the engine performs this hand-off for worker tasks at poll time, `ExecutionService.java:179-189`, but had no equivalent for async system tasks). Routing it through `WorkflowExecutor.updateTask` keeps task updates owned by the engine, exactly like a worker's update.

## Behavior on redelivery

The queue message IS still redelivered while the call runs; the redelivered execution sees `IN_PROGRESS`, does nothing, and `getEvaluationOffset` postpones the message past the blocking call (`responseTimeoutSeconds`, or 3600s default) — which also serves as the crash-recovery retry delay.

## #1322 is fixed at the root

The terminal-task overwrite in #1322 was only reachable through the duplicate attempt from #1321: the losing attempt's late completion rewrote `outputData`/`endTime` of a task the workflow had already consumed, corrupting the execution record and the agent-loop conversation history rebuilt from it. With the in-flight claim, the duplicate invocation never runs, so the corrupting write never exists. Task updates deliberately stay last-write-wins (legitimate engine flows update terminal tasks terminal→terminal: optional `FAILED` → `COMPLETED_WITH_ERRORS`, sub-workflow retry syncing the parent task).

**Not addressed here:** #1323 (AgentSpan output guardrails misfiring on tool-call turns) is an unrelated agentspan compiler bug and needs its own fix (`GuardrailCompiler`/`JavaScriptBuilder`, currently on the #1335 branch).

## Verification (`Issue1321DuplicateAsyncSystemTaskSpec` — real scanner, executor, redis queue)

Latch-controlled `@WorkerTask` bean blocks mid-"provider call"; the unack window is compressed to ~1s; a second worker polls and executes after it elapses.

- **Without this fix** (adapter/scanner stashed to main state, same test): `controllableWorker.invocations.get() == 1` → **actual: 2** — FAILED.
- **With this fix**: exactly once — PASSED.

Unit tests pin: claim persisted (IN_PROGRESS + marker) *before* the method is invoked and released after; redelivered execution with the marker returns false without invoking; legitimate `IN_PROGRESS`-callback re-execution is not blocked; the no-facade constructor keeps historical behavior. Full `:conductor-core:test` green.

## Known limitations (for reviewers)

- **Crash recovery**: if the JVM dies mid-invocation, the task stays `IN_PROGRESS`; the postponed message redelivers after `responseTimeoutSeconds` (or the 3600s default) and re-invokes — a bounded retry rather than a stall. Registering task defs for annotated types (so `responseTimeoutSeconds` is explicit) remains the recommended companion follow-up.
- **Narrow overwrite race**: a redelivered executor pass persists the (stale, IN_PROGRESS) task in its `finally`; if the real completion lands in that millisecond window, last-write-wins can briefly regress the record. Same class of race that exists today for any concurrent task update; noted for completeness.

## Alternatives considered

#1335 (queue-lease before blocking calls, executor hook) and #1347 (poll-worker execution mode) solve the same bug at different layers; this PR is the smallest change confined to the adapter that owns the blocking invocation. See `docs/design/2026-07-17-issue-1321-fix-options.md` on the #1335 branch for the full comparison.


core/src/main/java/org/conductoross/conductor/core/execution/tasks/annotated/AnnotatedWorkflowSystemTask.java
```diff
@@ -119,6 +178,31 @@ public boolean execute(
...
+     */
+    private void markInProgress(TaskModel task) {
+        if (executionDAOFacade == null) {
+            return;
+        }
+        try {
+            task.setStatus(TaskModel.Status.IN_PROGRESS);
+            task.setCallbackAfterSeconds(0);
+            task.setWorkerId(Utils.getServerId());
+            executionDAOFacade.updateTask(task);
```
v1r3n: update is done by the caller, it shouldn't do update here.

NicholasDCole: @v1r3n , I polled a workflow on my local -- LLM tasks are staying SCHEDULED and then flipping to COMPLETED . There is no IN_PROGRESS updating happening on main branch for these long running async system tasks


```
1784329597160 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329598198 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329599236 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329600280 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329601325 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329602374 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329603423 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329604468 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=08980cc5
1784329605510 coach_chat_agent_llm__1 COMPLETED pollCount=1 taskId=08980cc5
1784329606551 coach_chat_agent_llm__1 COMPLETED pollCount=1 taskId=08980cc5
```

NicholasDCole: Here is evidence of fix: 

```
1784330046858 coach_chat_agent_llm__1 SCHEDULED pollCount=0 taskId=46fa0c60
.........
1784330109346 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330110391 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330111433 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330112477 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330113524 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330114562 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330115605 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330116646 coach_chat_agent_llm__1 IN_PROGRESS pollCount=0 taskId=46fa0c60
1784330117684 coach_chat_agent_llm__1 COMPLETED pollCount=1 taskId=46fa0c60
```

v1r3n (commented):
Automated review.

manan164:
Hi @NicholasDCole , Executing LLM tasks as system tasks is what exposes us to this bug; executing them as poll-based workers puts them on the same battle-tested lifecycle our custom workers already run at scale — so let's converge on the worker model rather than patching the system-task path. I have created draft [PR](https://github.com/conductor-oss/conductor/pull/1359) for this.

mp-orkes:
> Hi @NicholasDCole , Executing LLM tasks as system tasks is what exposes us to this bug; executing them as poll-based workers puts them on the same battle-tested lifecycle our custom workers already run at scale — so let's converge on the worker model rather than patching the system-task path. I have created draft [PR](https://github.com/conductor-oss/conductor/pull/1359) for this.

Agree with @manan164 -  #1359 fixes this at the right layer. 

I would also add:

1. `isBlockingStart` creates a contradiction. A sync task (`isAsync=false`) already blocks. The point of `isAsync=true` is to not block the decider, yet these LLM tasks still block. So "async blocking" just moves where it blocks.

2. It inflates `responseTimeout`. The hand-off sets callbackAfterSeconds = responseTimeoutSeconds (or 3600), and checkForTaskResponseTimeout uses responseTimeout + callbackTime. So responseTimeoutSeconds=60 now times out at 120s; with no task def, a hung call isn't flagged for ~an hour.

3. If the call outlasts it, the duplicate still happens. The original #1321 failure mode, just with a wider window.

mp-orkes:
Closing in favor of #1369.

The pre-invocation `IN_PROGRESS` hand-off works, but it's a bigger change to the system-task status contract than the race needs: a new `isBlockingStart()`, an extra engine-owned write before every blocking `start()`, and it still relies on the redelivery firing and being caught (defense-in-depth) rather than not firing at all. Persisting `IN_PROGRESS` before `start()` also collides with tasks whose `start()` branches on `SCHEDULED` and shifts the response-timeout budget.

#1369 fixes the same race structurally and with a smaller surface: the message is never acked at poll, so it stays present (`containsMessage == true`) for the whole run and repair never treats an in-flight task as orphaned — no status change, no new contract. Good call folding #1322 in though; that reasoning carries over.

Thanks for the deep dive here, Nicholas — the redelivery write-up was genuinely useful.

PR #1350, opened 2026-07-17
ci: run python SDK agent e2e against a SQLite server built from source

ling-senpeng13:
## What

Adds a gating **`python-sdk-e2e`** job to CI that boots the conductor server **built from this commit** in **SQLite mode** (the default persistence — no external DB) and runs the released **python SDK agent e2e** suite against it. This catches server changes that would break the SDK end-to-end **before a release**.

## Why

Today nothing in this repo's CI exercises the agent SDK against a from-source server. The python SDK ships an e2e suite (`conductor-oss/python-sdk`, released as the `conductor-ai-e2e-python-*` bundle); wiring it here means a server-side regression in the agent surface (LLM chat/complete, tools, MCP, guardrails, scheduling, …) fails CI instead of surfacing after release.

## How

- **From-source server, SQLite:** builds `:conductor-server:bootJar` and boots it on `:8080`. SQLite is the default (`conductor.db.type=sqlite` / `queue.type=sqlite` / `indexing.type=sqlite`), so no Postgres/Redis/ES services are needed.
- **openai auto-config:** the server reads `conductor.ai.openai.api-key=${OPENAI_API_KEY:}`, so setting the env var is enough — no manual integration registration.
- **Bundle:** fetches the pinned `conductor-ai-e2e-python-2.0.0-rc2` bundle from `conductor-oss/python-sdk` at runtime, sha256-verified; `run.sh` installs `conductor-python[agents]==<version>` and runs the full suite.
- **Skip support (green today):** known failures are xfail-ed via an external pytest plugin (`.github/agent-e2e/known_failures_plugin.py`, loaded with `-p known_failures_plugin`) reading a per-repo list (`.github/agent-e2e/known-failures-python.json`). The suite is green against this server, so the list is **empty** — the mechanism stays so the lane can gate while any future gap is fixed (a fix XPASSes; a stale/mismatched entry is a harmless no-op that can't hide a regression).
- **Cost control:** runs on push/dispatch, and on PRs only when agent-relevant paths change (`ai/`, `conductor-agentspan/`, `server/`, `core/`, `sqlite-persistence/`, the workflow) via a `detect-changes` `agent` filter — so unrelated PRs don't spend LLM budget.

## ⚠️ Requires a repo secret

The job needs **`OPENAI_API_KEY`** (and optionally `ANTHROPIC_API_KEY`) configured as a repository/organization secret. GitHub does **not** pass secrets to fork PRs, so this job can only run with real LLM calls on branches in this repo (or once the secret is available) — a maintainer should validate it on a branch. The bundle version pin (`2.0.0-rc2`) should be bumped deliberately as new python-sdk releases cut their e2e bundles.

### DIFF
--- a/.github/agent-e2e/known-failures-python.json
+++ b/.github/agent-e2e/known-failures-python.json
@@ -1,0 +1,4 @@
+{
+  "_README": "conductor-oss known failures for the agent python e2e suite (bundle: conductor-ai-e2e-python-*, from conductor-oss/python-sdk), consumed by known_failures_plugin.py via E2E_KNOWN_FAILURES. Each key is a pytest node-id (or a '<file>::<Class>::<test>' suffix); the value is the reason + a tracking link. The suite is expected GREEN against the SQLite server built from this repo, so this list is EMPTY. Add an entry only for a genuine, tracked, deterministic failure while its fix is in flight; a key that matches nothing is a harmless no-op (the test still runs and the gate still catches a real break), and a fixed bug turns its test XPASS — remove the entry then. Do NOT list flaky tests (those are handled by rerun).",
+  "_HOWTO": "Reconcile from the run's results/junit-e2e.xml. Example entry: \"test_suite4_mcp_tools.py::TestSuite4McpTools::test_mcp_lifecycle\": \"reason — conductor-oss/conductor#NNNN\""
+}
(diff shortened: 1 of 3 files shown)


ling-senpeng13:
Superseded by #1351 — same change pushed to an internal branch so the org OPENAI_API_KEY/ANTHROPIC_API_KEY secrets are available to CI (GitHub withholds secrets from fork PRs, so this draft could never exercise the job).

PR #1352, opened 2026-07-17
Fix #1323: AgentSpan output guardrails pass through on tool-call turns

### LINKED ISSUE #1323: AgentSpan: output guardrails misfire on tool-call turns (output.result is [] mid-tool-use) and inject retry messages into the conversation

NicholasDCole:
Fixes #1323

## Summary

AgentSpan **output-position guardrails misfire on every legitimate tool-call turn**: they bind `content` to `${<agent>_llm.output.result}`, which is the empty list `[]` when the LLM turn is a tool call, so any non-empty-content guardrail (e.g. `require_nonempty_reply`, regex allow `\w`, `onFail: retry`) fails every healthy tool round and injects its retry message ("Your reply came back empty. Write your final answer now…") into the conversation **while the model is correctly mid-tool-use**. The contradictory pressure produces junk hedge tool calls (observed: `name_query: "placeholder"`, `"dummy"`) and silently burns the guardrail's `maxRetries` budget before the real final turn.

**Fix** (agentspan only, two production files):
- `GuardrailCompiler` binds a `toolCalls` input (from the agent LLM task's `output.toolCalls`) into all four output-position guardrail variants — regex / LLM-judge / custom / external.
- The generated scripts in `JavaScriptBuilder` short-circuit to `passed: true` **before content evaluation** when `toolCalls` is a non-empty array.

Deliberately unchanged: tool-level guardrails don't bind `toolCalls` (their job is to judge tool calls), and genuine final turns — including the genuinely empty/thinking-only replies these guardrails exist to catch — evaluate exactly as before.

## Test evidence

**Without the fix** (production files stashed to main state, same tests):

```
GuardrailToolCallShortCircuitScriptTest > shortCircuitsToPassedWhenToolCallsPresentAndContentEmpty  FAILED
Issue1323GuardrailToolCallBindingTest   > compiledOutputGuardrailBindsToolCallsFromLlmOutput        FAILED
Issue1323GuardrailToolCallMisfireTest   > toolCallTurn_guardrailMustPass_andMustNotInjectRetry      FAILED
6 tests completed, 3 failed
```

**With the fix**: full `:conductor-agentspan:test` suite — BUILD SUCCESSFUL, 0 failures.

The tests cover both layers and the end-to-end behavior:
- `Issue1323GuardrailToolCallBindingTest` — the compiled guardrail INLINE task's inputs contain the `toolCalls` binding for all output-position variants.
- `GuardrailToolCallShortCircuitScriptTest` / `Issue1323GuardrailToolCallMisfireTest` — the **real generated scripts evaluated in GraalJS**: a tool-call turn (content `[]`, non-empty `toolCalls`) passes and injects no retry; a genuinely empty final reply (no tool calls) still fails, preserving the incident behavior the guardrail was built for.
- `GuardrailCompilerTest` additions pin the binding across guardrail types.

## Scope

Six files, all under `agentspan/` — no core, no queue, no executor changes. Extracted from the #1335 branch so it can be reviewed and merged independently of the #1321/#1322 fix (#1349).


v1r3n (commented):
Automated review.

v1r3n (approved):

Issue #1331, opened 2026-07-16
Agent execution prune with large olderThanDays deletes RECENT executions (negative epoch cutoff + mishandled negative comparison) — data loss


NicholasDCole:
## Describe the bug

`POST /api/agent/executions/prune?olderThanDays=99999` deleted an execution that had completed the same day, returning `{"deleted":1}`. Any `olderThanDays` ≳ 20,000 (epoch-start / 86,400,000 ms) silently turns "delete ancient records" into "delete everything terminal" — unrecoverable data loss from a plausible input on a destructive, unscoped endpoint.

## Root cause (two layers)

**1. No parameter validation + negative epoch cutoff** — [`AgentService.pruneExecutions`](https://github.com/conductor-oss/conductor/blob/main/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java#L579):

```java
long cutoffEpochMs = Instant.now().minus(olderThanDays, ChronoUnit.DAYS).toEpochMilli();
```

With `olderThanDays=99999` the cutoff is a **negative** epoch (~year 1752). There is no validation on the parameter — note that a *negative* `olderThanDays` produces a **future** cutoff and legitimately matches **all** terminal executions.

**2. Search backend mishandles the negative comparison.** The generated query `workflowType IN (...) AND status = '<S>' AND endTime < <negativeLong>` should match nothing, but the search backend returns executions with recent positive endTimes. Verified read-only: a freshly completed execution matched the negative-cutoff query (`totalHits=1`, expected 0). Everything that query returns is then hard-deleted in batches (`workflowService.deleteWorkflow`).

## To Reproduce

Non-destructive (observe the wrong search match):

```bash
curl -G 'http://localhost:8080/api/workflow/search' \
  --data-urlencode 'query=workflowType IN (<an-agent-name>) AND status = "COMPLETED" AND endTime < -1000000000000' \
  --data-urlencode 'freeText=*'
# -> totalHits > 0 despite the impossible predicate
```

Destructive confirmation: `POST /api/agent/executions/prune?olderThanDays=99999` → recent terminal executions deleted.

## Expected behavior

- Reject `olderThanDays` outside a sane range (must be ≥ 1, and the computed cutoff must be ≥ 0) with a 400.
- The search layer should evaluate numeric comparisons with negative literals correctly (that half also affects any user-supplied search query, not just prune).

## Details

- Conductor version: 3.32.0-rc (main @ c2439b7e6)
[...]

PR #1360, opened 2026-07-20
Guard pruneExecutions against overflow and non-positive age deleting recent executions

nikhiln64:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

`pruneExecutions` computed its cutoff as `Instant.now().minus(olderThanDays, DAYS).toEpochMilli()` with no bounds check. Two ways that turned the prune endpoint into a data-loss operation:

- very large values (the issue's `olderThanDays=99999`) push the computed epoch **negative**, and the search backend matched recent executions against the negative bound, which were then hard-deleted in batches
- non-positive values put the cutoff **in the future**, matching every terminal execution

The fix extracts the cutoff computation into `computePruneCutoffEpochMs(int, Instant)`: rejects `olderThanDays < 1` with a clear message, and clamps the cutoff to epoch start, where `endTime < 0` correctly matches nothing ("older than anything that exists").

Extracted as a static method with an injectable clock so the boundary behavior is unit-testable deterministically: `AgentServicePruneCutoffTest` covers the normal case, the 99999-day clamp, the largest non-clamped age, and rejection of zero/negative. 4/4 green; `spotlessApply` run.

Issue #1331

Alternatives considered
----

Also considered chasing why the search backend matches positive `endTime` values against a negative long bound (the second layer of the bug) - worth a follow-up in the search query serialization, but the guard here closes the data-loss hole regardless of backend behavior.

Developed with AI assistance (Claude Code), reviewed and driven by a human.

### DIFF
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
@@ -30,6 +30,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.mvc.method.annotation.SseEmitter;
 
+import com.netflix.conductor.annotations.VisibleForTesting;
 import com.netflix.conductor.common.config.ObjectMapperProvider;
 import com.netflix.conductor.common.metadata.tasks.Task;
 import com.netflix.conductor.common.metadata.tasks.TaskDef;
@@ -567,6 +568,27 @@ public class AgentService {
     }
 
     /**
+     * Computes the prune cutoff, guarding the two ways an unchecked {@code olderThanDays} turned
+     * the prune into a data-loss operation (issue #1331): non-positive values put the cutoff in the
+     * future (matching every terminal execution), and very large values push the computed epoch
+     * negative, which the search backend matched against recent executions. A cutoff clamped to
+     * epoch start matches nothing, which is the correct meaning of "older than anything that
+     * exists".
+     *
+     * @param olderThanDays minimum age in days, must be >= 1
+     * @param now the current instant
+     * @return cutoff in epoch milliseconds, never negative
+     */
+    @VisibleForTesting
+    static long computePruneCutoffEpochMs(int olderThanDays, Instant now) {
+        if (olderThanDays < 1) {
+            throw new IllegalArgumentException(
+                    "pruneExecutions: olderThanDays must be >= 1, got " + olderThanDays);
+        }
+        return Math.max(0L, now.minus(olderThanDays, ChronoUnit.DAYS).toEpochMilli());
+    }
+
+    /**
      * Bulk-delete completed execution records older than {@code olderThanDays} days.
      *
      * <p>Searches for COMPLETED, FAILED, TERMINATED, and TIMED_OUT executions whose end time is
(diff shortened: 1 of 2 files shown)


v1r3n (commented):
Looks good. The cutoff helper rejects unsafe non-positive ages, clamps pre-epoch cutoffs, is wired into the actual prune path, and has deterministic coverage for the boundary cases. The existing controller advice maps the new `IllegalArgumentException` to a `400 Bad Request`, so invalid API input is handled correctly.

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
```diff
@@ -576,8 +576,28 @@ public void deleteExecutionRecord(String executionId, boolean archiveTasks) {
      * @param archiveTasks if true, archive task records instead of deleting
      * @return number of executions deleted
      */
```
kowser-orkes: This documentation is for `pruneExecutions` is now dangling. Can you please fix it?

nikhiln64:
The branch was just updated to the latest main and the workflow run is waiting for maintainer approval. Whenever someone has a moment to approve it, the PR is otherwise ready for review.

kowser-orkes (changes requested):

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
kowser-orkes: Thanks for the PR. It looks good to me except that documentation issue got created

nikhiln64: Good catch, thank you. The helper insertion left the pruneExecutions javadoc stranded above it. Moved the doc back onto the method and pushed.

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
```diff
@@ -566,6 +566,26 @@ public void deleteExecutionRecord(String executionId, boolean archiveTasks) {
...
+     * future (matching every terminal execution), and very large values push the computed epoch
+     * negative, which the search backend matched against recent executions. A cutoff clamped to
+     * epoch start matches nothing, which is the correct meaning of "older than anything that
+     * exists".
+     *
+     * @param olderThanDays minimum age in days, must be >= 1
+     * @param now the current instant
+     * @return cutoff in epoch milliseconds, never negative
+     */
+    static long computePruneCutoffEpochMs(int olderThanDays, Instant now) {
```
kowser-orkes, line 583: It should be `private`

kowser-orkes (changes requested):
Can you please rebase and squash all changes. Three commits seems redundant :-)

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
nikhiln64: I kept it package private on purpose so AgentServicePruneCutoffTest can exercise the boundary cases deterministically with a fixed clock, since making it fully private would force the tests to construct the whole service. I have annotated it with @VisibleForTesting following the pattern in WorkflowMonitor and WorkflowRepairService so the intent is explicit. If you would still rather have it private I am happy to switch and rework the tests through the public method instead, wdyt.

nikhiln64:
Done, rebased onto the latest main and squashed everything into a single commit. On your earlier note about visibility, I annotated the helper with @VisibleForTesting following the WorkflowMonitor pattern since the boundary tests call it directly with a fixed clock, but if you would rather have it fully private I will rework the tests through the public method. The fresh CI run will need an approval when someone gets a chance.

kowser-orkes (approved):

Issue #1311, opened 2026-07-16
SWITCH task: javascript evaluator rejects runtime-bound expressions at registration time


nthmost-orkes:
## Summary

A SWITCH task with `"evaluatorType": "javascript"` fails **registration** with a
`ReferenceError` if the expression references any variable, even one that will be
available at runtime via `inputParameters`. This makes the javascript evaluator
essentially unusable for dynamic dispatch — its primary purpose.

## Error (at `PUT /api/metadata/workflow`)

```
Expression is not well formatted: ReferenceError: value is not defined
```

## Example Definition

```json
{
  "type": "SWITCH",
  "evaluatorType": "javascript",
  "expression": "$.inputValue > 10 ? 'big' : 'small'",
  "inputParameters": {
    "inputValue": "${workflow.input.myValue}"
  },
  "decisionCases": {
    "big":   [...],
    "small": [...]
  }
}
```

This definition is rejected at registration even though `inputValue` (and
therefore `$.inputValue`) will be fully defined at runtime.

## Root Cause

The workflow registration validator executes the JS expression with no variable
bindings in order to "validate" it. Any expression that uses anything other than
literal constants will throw at validation time.

## Impact

Only constant-expression SWITCH conditions work (e.g. `"'big'"`, `"1+1"`). Any
expression that dispatches based on workflow or task data — which is the whole
point of a javascript evaluator in a conditional task — is impossible to register.

## Fix

The validator should not execute the expression with empty bindings. Options:
- Parse/compile the expression without evaluating it (Graal `Source.create` / AST parse only)
- Skip expression execution during registration; validate only that it is syntactically valid JS
- Document clearly that only the `value-param` evaluator type works for dynamic dispatch (not a fix, but a workaround note)

## Files

`core/src/main/java/com/netflix/conductor/validations/WorkflowTaskTypeConstraint.java`

nthmost-orkes:
Discovered against **Conductor 3.32.0-rc.9** running on JDK 21 (loki.local).

nikhiln64:
[...]

PR #1361, opened 2026-07-20
Validate SWITCH javascript expressions by syntax only at registration

nikhiln64:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

`validateScriptExpression` executed the SWITCH expression via `ScriptEvaluator.eval(expression, inputParameters)` at registration time - but the bindings map still holds unresolved `${...}` placeholders at that point, so any expression operating on runtime-bound values throws and registration wrongly rejects a valid definition. Example that fails today but is perfectly valid at runtime: `$.items.filter(i => i > 10).length > 0 ? 'big' : 'small'` with `items: "${workflow.input.items}"` - at registration `$.items` is a placeholder *string*, so `.filter` throws.

The fix adds `ScriptEvaluator.validateScriptSyntax(String)`: parses the source via GraalJS `Context.parse` **without executing**, throwing only on genuine syntax errors (`PolyglotException.isSyntaxError()`). Registration becomes a syntax check, which is all it can correctly be before placeholder resolution. Bonus: the parse warms the same `Source` cache the runtime eval uses.

Two tests added to `WorkflowTaskTypeConstraintTest`: the runtime-bound expression above must produce no "Expression is not well formatted" violation (verified this test FAILS against the previous implementation - it rejects the expression), and a genuinely malformed expression (`$.inputValue > 10 ? 'big' :`) must still be rejected. 35/35 green; `spotlessApply` run.

Issue #1311

Alternatives considered
----

Treating `ReferenceError` as a pass while still evaluating was considered and rejected: evaluation with placeholder bindings can fail in many shapes (TypeError on string methods, NPE-ish member access), and whitelisting error types is a losing game. Parse-only is the correct semantic for registration.

Developed with AI assistance (Claude Code), reviewed and driven by a human.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/events/ScriptEvaluator.java
+++ b/core/src/main/java/com/netflix/conductor/core/events/ScriptEvaluator.java
@@ -238,6 +238,28 @@ public class ScriptEvaluator {
     }
 
     /**
+     * Validates that the script is syntactically well formed WITHOUT executing it. Registration
+     * time validation must not evaluate expressions: the input bindings still hold unresolved
+     * {@code ${...}} placeholders, so any expression referencing a runtime-bound value would throw
+     * a ReferenceError and wrongly reject a valid definition (issue #1311).
+     *
+     * @param script Script whose syntax should be checked.
+     * @throws IllegalArgumentException if the script has a syntax error.
+     */
+    public static void validateScriptSyntax(String script) {
+        ensureInitialized();
+        try (Context context = createNewContext()) {
+            context.parse(getSource(script));
+        } catch (PolyglotException e) {
+            if (e.isSyntaxError()) {
+                throw new IllegalArgumentException(e.getMessage());
+            }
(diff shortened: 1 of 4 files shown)


v1r3n (approved):

v1r3n:
@nikhiln64 can you address the test failures?

nikhiln64:
Fixed and pushed. The failing test was testUpdateWorkflowDefWithJavscriptEvaluator, which used 1>abcd as its invalid expression. That is actually valid javascript syntax and it only fails when evaluated, which is the exact thing this PR stops doing at registration. I switched the test to a genuine syntax error so it still protects against malformed expressions, and the whole core suite passes for me locally now.

nikhiln64:
@v1r3n after the test fix and the branch update the new CI run has not been triggered yet. Could you approve the workflow run so the fixed suite shows green here? The full core tests pass for me locally.

v1r3n (approved):

PR #1364, opened 2026-07-20
feat(core): poll-time claim for blocking async system tasks

### LINKED ISSUE #1321: Async system tasks running longer than the queue unack window are executed twice (duplicate LLM_CHAT_COMPLETE provider calls)

NicholasDCole:
Fixes #1321 and #1322. Related: #1347.

## Problem

Async system tasks were popped and ACKed before their state changed. A blocking annotated task such as `LLM_CHAT_COMPLETE` could remain `SCHEDULED` for longer than the poll timeout, allowing recovery/redelivery to dispatch a duplicate provider request.

## Change

`SystemTaskWorker` now claims work through `ExecutionService.pollQueue()` instead of popping the queue directly.

- Persists `SCHEDULED → IN_PROGRESS`, worker ID, start time, and poll count before dispatching the system task.
- Reuses the established remote-worker claim path, including decider postponement to the response timeout.
- Polls the exact queue name, preserving execution namespace and isolation-group queues.
- Removes the annotated-task `IN_PROGRESS` guard: claimed initial work is already `IN_PROGRESS`, and the guard prevented its first invocation.

The design note documents the implemented flow, recovery boundary, and includes a Mermaid sequence diagram:
[`docs/design/2026-07-20-system-task-worker-poll-claim.md`](docs/design/2026-07-20-system-task-worker-poll-claim.md)

## Verification

- `./gradlew :conductor-core:spotlessApply :conductor-test-harness:spotlessApply`
- `./gradlew :conductor-core:test --tests com.netflix.conductor.core.execution.tasks.TestSystemTaskWorker`
- `./gradlew :conductor-test-harness:test --tests com.netflix.conductor.test.integration.SystemTaskPollClaimSpec`

The integration test drives the real `SystemTaskWorker`, blocks an annotated provider worker, observes `IN_PROGRESS` while it runs, then verifies exactly one invocation and completion.


core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorker.java
```diff
@@ -120,8 +122,18 @@ void pollAndExecute(WorkflowSystemTask systemTask, String queueName) {
...
+                            .pollQueue(
+                                    queueName,
+                                    "system-task-worker",
+                                    messagesToAcquire,
+                                    queuePopTimeout)
+                            .stream()
+                            .map(Task::getTaskId)
+                            .toList();
 
             Monitors.recordTaskPoll(queueName);
```
v1r3n, line 138: can you check if the executionService already does publish the counter?

core/src/main/java/com/netflix/conductor/service/ExecutionService.java
```diff
@@ -110,11 +110,45 @@ public List<Task> poll(String taskType, String workerId, int count, int timeoutI
 
     public List<Task> poll(
             String taskType, String workerId, String domain, int count, int timeoutInMilliSecond) {
+        return pollFromQueue(
```
v1r3n, line 113: do not change this one, there shouldn't be any changes here

Issue #202, opened 2024-07-08
Duplicate async system tasks


arorashivam:
**Issue Description**
We are seeing duplicate sub_workflows were getting triggered for a workflow

On further investigation, We identified a race condition b/w sweeper thread and system task worker thread leading to a duplicate SUB_WORKFLOW system task. The race condition is not just limited to sub_workflow system task but can be replicated with any async System task. 

Explanation:
1. The system task worker thread-1 polls and acknowledges the task from the task's corresponding queue & triggers the AsyncSystemTaskExecutor::execute. The execute method isn't yet completed, therefore the status is still `SCHEDULED`
2. Meanwhile, sweeper tries to sweep this workflow and finds it's task is in repairable state. It then adds the task to the queue as it finds the task is not in the queue. 
3. The thread-1 now marks the status of the sub_workflow task to IN_PROGRESS
4. The system task worker thread-2 polls the same task (added by sweeper) from the queue and executes it as new sub_workflow. 

**Details**
Conductor version: 3.20.0
Persistence implementation: Postgres
Queue implementation: Dynoqueues
Lock: Redis
Workflow definition: This definition is just for reference purpose, to explain that we are using a sub_workflow task which is getting duplicated.
`{
    "name": "parent_workflow",
    "description": "Parent Workflow",
    "version": 1,
    "schemaVersion": 2,
    "inputParameters": [],
    "tasks": [
        {
            "name": "loopTask",
            "taskReferenceName": "loopTask",
            "type": "DO_WHILE",
            "inputParameters": {
                "batch": "${workflow.input.batch}",
                "batchSize": "${workflow.input.batch.length()}"
            },
            "loopCondition": "$.loopTask['iteration'] < $.batchSize",
            "loopOver": [
                {
                    "name": "loopTask_prepare",
                    "taskReferenceName": "loopTask_prepare",
                    "type": "INLINE",
                    "inputParameters": {
                        "evaluatorType": "javascript",
                        "expression": "function sFun(){ if($.batch === null) {return null} else { return $.batch.get($.iteration-1) } } sFun();",
                        "batch": "${loopTask.input.batch}",
                        "iteration": "${loopTask.output.iteration}"
                    },
                    "asyncComplete": false
                },
                {
                    "name": "sub_workflow_task",
                    "taskReferenceName": "sub_workflow_task",
                    "type": "SUB_WORKFLOW",
                    "inputParameters": {
                        "item": "${loopTask_prepare.output.result}",
[...]

PR #1369, opened 2026-07-22
Fix #1321: reserve an async system task's queue message while it runs

### LINKED ISSUE #1321: Async system tasks running longer than the queue unack window are executed twice (duplicate LLM_CHAT_COMPLETE provider calls)

mp-orkes:
Fixes #1321. Alternative to #1367 — fixes the duplicate at the queue/scheduler layer instead of in the annotated adapter.

## Problem

The `system-task-worker` **removes** a message at poll (`ackTaskReceived`) so while a task runs it has **no** queue message — and the `WorkflowSweeper` repair or `WorkflowRepairService` re-queue any running task whose message is "missing". This is causing duplicates.

Affects every async system task; long-running LLM/A2A tasks hit it reliably (their run outlasts the ~30s sweep).

We didn't find a new bug, we found an old one that keeps coming back:
- https://github.com/Netflix/conductor-community/issues/70
- https://github.com/conductor-oss/conductor/issues/202
- https://github.com/conductor-oss/conductor/issues/630

## Fix
1. Keep a running task's message present so repair leaves it alone. 
2. Time out a run that overruns instead of re-executing it — on the shared path, no per-task gating:

- **`SystemTaskWorker`** no longer removes the message at poll. The popped message stays in the queue (invisible while unacked), so a running task still *has* a message.
- **`AsyncSystemTaskExecutor`** reserves the message before invoking: `setUnackTimeout(queue, taskId, responseTimeoutSeconds)` Uses the existing per-task `responseTimeout` (no new property), falling back to `TaskDef.ONE_HOUR` when unset.
- **Overruns time out, not duplicate.** The executor persists `startTime` before the first invocation (status left unchanged so `SUB_WORKFLOW` etc. whose `start()` branches on `SCHEDULED` still work). On any redelivery, if the task has started and hasn't responded within `responseTimeout` (`now - updateTime ≥ responseTimeout`), it's marked `TIMED_OUT` (retriable) — the retry/timeout policy makes a new attempt or fails; it is never re-run in parallel under the same taskId. Using `updateTime` keeps the long-running `IN_PROGRESS + callbackAfterSeconds` (LLM/A2A) flow from timing out while it checks in.

`AsyncSystemTaskExecutor`'s `finally` still owns the normal outcome (remove on terminal, postpone on non-terminal).

## How to test
`./gradlew :conductor-core:test --tests "*AsyncSystemTaskExecutorTest*" --tests "*TestSystemTaskWorker*"` — reserve with `responseTimeout` (and `ONE_HOUR` fallback) before `start()`; a redelivered overrun is `TIMED_OUT`, not re-invoked; the poll no longer acks/removes the message.

## Notes
- A crash *mid-execute* leaves the message reserved and recovered after `responseTimeout` (bounded), slower than repair's ~30s.
- Set a realistic `responseTimeout` for long tasks; a task whose `responseTimeout` is shorter than its work will time out and (per policy) retry/fail rather than run twice.
- **Zombie late-write (#1322, not fixed here):** a timed-out task's original invocation keeps running and, on completion, can overwrite the `TIMED_OUT`/retry state under the original taskId — rejecting that late write is a separate follow-on.
- Design write-up: `docs/design/2026-07-21-issue-1321-poll-level-reserve.md`.

## Update (review follow-up)

The overrun timeout is now scoped to **`SCHEDULED`** tasks only. A blocking `start()` never leaves `SCHEDULED`, so that's the only overrun `DeciderService.isResponseTimedOut` (which budgets `responseTimeout + callbackAfterSeconds`) can't already see. The earlier unscoped check fired sooner than the decider and could wrongly time out an `IN_PROGRESS` task waiting for its next callback when `callbackAfterSeconds ≥ responseTimeout`; `IN_PROGRESS` response-timeouts stay with the decider. Added tests: an `IN_PROGRESS` callback-loop task is not timed out, and `WorkflowSweeperTest` proves repair does not re-queue an in-flight async system task whose message is present (#202/#630 guard).

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -155,12 +155,15 @@ public class AsyncSystemTaskExecutor {
 
             boolean scheduled = task.getStatus() == TaskModel.Status.SCHEDULED;
             if (scheduled || task.getStatus() == TaskModel.Status.IN_PROGRESS) {
-                if (hasExceededResponseTimeout(task)) {
-                    // The message was redelivered while a previous invocation is still in flight
-                    // past responseTimeout. Do NOT run it again in parallel (issue #1321): the task
-                    // has exceeded its allowed time, so time it out and let the retry/timeout
-                    // policy
-                    // decide (retry as a new attempt, or fail). The finally block below removes the
+                if (scheduled && hasExceededResponseTimeout(task)) {
+                    // A blocking task's start() runs synchronously and never moves the task to
+                    // IN_PROGRESS, so it stays SCHEDULED for its whole run. If the message is
+                    // redelivered (reservation lapsed) while that run is still in flight past
+                    // responseTimeout, do NOT run it again in parallel (issue #1321): time it out
+                    // and let the retry/timeout policy decide. Only SCHEDULED is handled here;
+                    // IN_PROGRESS response-timeouts are owned by DeciderService.isResponseTimedOut
+                    // (invoked from decide()), which uses responseTimeout + callbackAfterSeconds and
+                    // so must not be duplicated/pre-empted here. The finally block removes the
                     // message and re-evaluates the workflow.
                     task.setStatus(TaskModel.Status.TIMED_OUT);
                     task.setReasonForIncompletion(
@@ -276,10 +279,10 @@ public class AsyncSystemTaskExecutor {
     /**
      * True if the task has already started (startTime set) and has not been updated within its
      * responseTimeout — i.e. a redelivered message belongs to a run that overran its allowed time
-     * and should be timed out rather than re-executed. Uses updateTime (time since the last
-     * response), so a worker that keeps checking in within responseTimeout (long-running LLM/A2A
-     * callbacks) is not timed out. Requires updateTime > 0: a just-scheduled task whose mapper set
-     * startTime (e.g. JOIN) has updateTime == 0 and must not be treated as overrun.
+     * and should be timed out rather than re-executed. Only meaningful for a SCHEDULED task (a
+     * blocking start() that never returned); the caller gates on that. Requires updateTime > 0: a
+     * just-scheduled task whose mapper set startTime (e.g. JOIN) has updateTime == 0 and must not be
+     * treated as overrun.
      */
     private boolean hasExceededResponseTimeout(TaskModel task) {
         return task.getStartTime() > 0
--- a/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
@@ -274,14 +274,16 @@ public class AsyncSystemTaskExecutor {
     }
 
     /**
-     * True if the task has already started (startTime set, persisted on the first execution) and
-     * has not been updated within its responseTimeout — i.e. a redelivered message belongs to a run
-     * that overran its allowed time and should be timed out rather than re-executed. Uses
-     * updateTime (time since the last response), so a worker that keeps checking in within
-     * responseTimeout (long-running LLM/A2A callbacks) is not timed out.
+     * True if the task has already started (startTime set) and has not been updated within its
+     * responseTimeout — i.e. a redelivered message belongs to a run that overran its allowed time
+     * and should be timed out rather than re-executed. Uses updateTime (time since the last
+     * response), so a worker that keeps checking in within responseTimeout (long-running LLM/A2A
+     * callbacks) is not timed out. Requires updateTime > 0: a just-scheduled task whose mapper set
+     * startTime (e.g. JOIN) has updateTime == 0 and must not be treated as overrun.
      */
     private boolean hasExceededResponseTimeout(TaskModel task) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/netflix/conductor/core/execution/tasks/SystemTaskWorker.java
```diff
@@ -141,8 +137,9 @@ void pollAndExecute(WorkflowSystemTask systemTask, String queueName) {
                                 queueName);
                         Monitors.recordTaskPollCount(queueName, 1);
 
-                        executionService.ackTaskReceived(taskId);
```
mp-orkes, line 144: Don't do this!
Sweeper's repair logic will cause issues. It will send a duplicate message for the task if it's not in the queue.
```
1. POP
2. ACK
## → Sweeper race: after acking and before finishing 
execution the sweeper may push a message for this 
task to repair execution.
3. EXECUTE
```

Image

core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -152,28 +153,52 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
...
-                    if (task.getStatus() == TaskModel.Status.SCHEDULED) {
-                        task.setStartTime(System.currentTimeMillis());
-                        Monitors.recordQueueWaitTime(task.getTaskType(), task.getQueueWaitTime());
-                        systemTask.start(workflow, task, workflowExecutor);
-                    } else {
-                        systemTask.execute(workflow, task, workflowExecutor);
+                } else {
+                    // Keep the message present but invisible for the run so repair can't re-queue
+                    // this running task (#1321).
+                    reserveInflightMessage(queueName, task);
```
mp-orkes: `reserveInflightMessage` extends the message's unack lease for `responseTimeoutSeconds`

mp-orkes:
1. **[MINOR] `AsyncSystemTaskExecutor.java:155`** — the SCHEDULED overrun check assumes "a blocking `start()` never leaves SCHEDULED." If any `start()` returns quickly leaving status SCHEDULED and gets postponed, then once `callbackAfterSeconds >= responseTimeout` it's falsely `TIMED_OUT` on redelivery — the same failure class just fixed for IN_PROGRESS. Likely theoretical (built-in `start()`s transition status), but it's an unverified invariant load-bearing for correctness. Confirm no system task postpones while SCHEDULED.

2. **[MINOR] Residual IN_PROGRESS gap** — `reserveInflightMessage` falls back to `ONE_HOUR` when `responseTimeoutSeconds == 0`, but `DeciderService.isResponseTimedOut` returns false when the TaskDef's `responseTimeoutSeconds == 0` (`DeciderService.java:937`). So an IN_PROGRESS task with no configured timeout whose `execute()` blocks past the 1h reserve can be redelivered and re-run in parallel with nothing to time it out. Rare (blocking work is in `start()`), but it's the one corner the fix doesn't close — worth a doc line.

3. **[MINOR] `WorkflowSweeperTest`** is close to tautological — it stubs `containsMessage → true` and asserts `push` isn't called, restating the repair predicate. It guards the sweeper's decision but never shows the reserve actually keeps the message present. A real-queue integration test (pop → reserve → sweep → assert not re-pushed) would be the real regression guard for #202/#630.

4. **[MINOR] `AsyncSystemTaskExecutor.java` (SCHEDULED branch)** — the pre-`start()` `updateTask` doubles per-task persistence + indexing writes (`ExecutionDAOFacade.updateTask:571` indexes on each call when async indexing is off). Disclosed in test comments (`2 * updateTask`) but the added write/index volume on the scheduling hot path isn't called out. Confirm it's acceptable at throughput.

core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -221,6 +246,45 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
...
+    /**
+     * Extend the task's queue message visibility to cover the invocation (issue #1321), using its
+     * {@code responseTimeoutSeconds} or the default {@link TaskDef#ONE_HOUR} when unset.
+     */
+    private void reserveInflightMessage(String queueName, TaskModel task) {
+        try {
+            queueDAO.setUnackTimeout(
+                    queueName, task.getTaskId(), effectiveResponseTimeoutSeconds(task) * 1000L);
+        } catch (Exception e) {
+            LOGGER.error(
```
manan164, line 271: One edge case I can think of is if pop is successful but setUnackTimeout throws exception then again this issue will happen that after T+60 seconds the message will be eligible to pop again. Ideally this method should have boolean return type and if reserveInflightMessage returns false  we should not start the task.

mp-orkes: You're right. Good catch.

mp-orkes:
Note: To reproduce the issue locally, for testing purposes, something like [SlowFakeWorker.java](https://gist.github.com/mp-orkes/b0762f1d686f9e2ccc7a7009f4ab0f73) can be used.

v1r3n (approved):

manan164 (approved):

PR #1370, opened 2026-07-22
fix: resolve 4 open issues (#1304, #1310, #1316, #1319)

### LINKED ISSUE #1304: [UI] Stale "AgentSpan" branding in the empty Agents definitions tab
**Found by:** Deepti
**Area:** Agents UI / branding

**Description**
The **Agents** definitions tab empty state (shown when no agent is deployed) still references "AgentSpan".

**Steps to reproduce**
1. Open the Conductor UI
2. Open the **Agents** definitions tab
3. With no agent deployed, view the empty state

**Actual** — empty state reads:
- "Agents are AI agent definitions compiled and run as native Conductor workflows by the embedded **AgentSpan** runtime."
- "No agents deployed yet? Build one with the **AgentSpan** SDK."

**Expected**
No `AgentSpan` references — use the current naming.

**Environment**
OSS getting-started UI. Screenshot available (attach in-thread).

_Same root cause as the Create Agent dialog branding issue._

### LINKED ISSUE #1310: WAIT task: misleading error 'date or duration is null' when ISO-8601 duration format is used
## Summary

A WAIT task configured with an ISO-8601 duration (e.g. `"PT1S"`) fails with an
error that says the value is null, when the actual problem is a format mismatch.

## Error

```
Either date or duration is passed as null
```

## Cause

`DateTimeUtils.parseDuration()` accepts only a custom format like `"1s"`, `"2m"`,
`"1h 30m"`. When the regex does not match (e.g. for `"PT1S"`), it throws an
`IllegalArgumentException` that gets caught and re-thrown as the misleading
"passed as null" message.

ISO-8601 duration format is widely used (Java's `Duration.parse()`, XML Schema,
many external APIs) and a natural thing to try. The error gives no hint that the
input was present but in the wrong format.

## Fix Options

1. Accept ISO-8601 duration as an alternative in `parseDuration()` (e.g. via
   `java.time.Duration.parse()` as a fallback).
2. Emit a clearer error: `"WAIT duration must use format '1s', '2m', '1h 30m'. ISO-8601 format (e.g. 'PT1S') is not supported."`.
3. At minimum, fix the error message so it does not imply the value was absent.

## Reproduction

```json
{
  "type": "WAIT",
  "name": "wait1",
  "taskReferenceName": "wait_ref",
  "inputParameters": {
    "duration": "PT1S"
  }
}
```

Task fails immediately with "Either date or duration is passed as null" despite
`duration` being present.

## Files

`core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java`

### LINKED ISSUE #1316: [3.32.0-rc.9] API Docs button in sidebar doesn’t open in a new tab and when you go back it gets stuck on loading
### Describe the bug

Steps to reproduce:
1. Click on API Docs button in the side bar
2. Click the back button in the browser

It shows this spinner forever until you navigate to a different page:
Image

### LINKED ISSUE #1319: Scheduler Pause/Resume silently fails (wrong HTTP method GET instead of PUT)
**Describe the bug**
Scheduler Pause/Resume silently fails (wrong HTTP method GET instead of PUT)

**Details**
Conductor version:
Open Schedules list then click Pause on any schedule there is no effect 

**To Reproduce**
1. Go to 'Schedules'
2. Click on 'Pause' on any schedule
3. See error

Image

**Code to reproduce (paste relevant snippet):**
```
[paste the minimal code that triggers the bug]
```

**Expected behavior**

**Screenshots**

**Additional context**

kunalrawat425:
## Summary

Four focused bug fixes — one commit covers all; each fix is independent.

| Issue | Area | Fix |
|---|---|---|
| #1310 | Java / WAIT task | `DateTimeUtils.parseDuration()` now accepts ISO-8601 durations (`PT1S`, `PT1H30M`) as a fallback via `Duration.parse()`. Error messages distinguish absent value from unsupported format. |
| #1319 | UI / Scheduler | `Schedules.tsx` `handlePauseSchedule` and `handleResumeSchedule` both used `method: "get"` — changed to `"put"` to match the backend `PUT /scheduler/schedules/{name}/pause|resume` contract. |
| #1304 | UI / Agent branding | Removed stale "AgentSpan" copy from `AgentDefinitions.tsx` (empty state), `AgentExecutions.tsx` (intro), and `Skills.tsx` (intro). Replaced with current Conductor-native naming. |
| #1316 | UI / API Docs sidebar | `ApiReferencePage` used `window.location.href` (same-tab redirect). Navigating back landed on the spinner indefinitely. Now uses `window.open(..., "_blank")` + `navigate(-1)` so Swagger opens in a new tab and the user stays in the app. |

## Verification

**Java (`conductor-core`):**
- `./gradlew spotlessApply` — no changes (style already clean)
- `./gradlew :conductor-core:test --tests "com.netflix.conductor.core.utils.DateTimeUtilsTest"` — BUILD SUCCESSFUL
- Added parametrized tests: valid ISO-8601 (`PT1S`, `PT1M`, `PT1H`, `P1D`, `P1DT2H3M4S`, case-insensitive) + invalid (`Pnotvalid`, `PT`, `P1X`)

**UI (`ui-next`):**
- `pnpm typecheck` — 0 errors
- `pnpm test` — 44 test files, 590 passed, 1 skipped

Fixes #1310
Fixes #1319
Fixes #1304
Fixes #1316

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
@@ -35,8 +35,24 @@ public class DateTimeUtils {
                     Pattern.CASE_INSENSITIVE);
 
     public static Duration parseDuration(String text) {
+        // Accept ISO-8601 duration (e.g. "PT1S", "PT1H30M") as a convenience fallback.
+        if (text != null && text.toUpperCase().startsWith("P")) {
+            try {
+                return Duration.parse(text);
+            } catch (Exception e) {
+                throw new IllegalArgumentException(
+                        "Not valid ISO-8601 duration: "
+                                + text
+                                + ". Expected format e.g. 'PT1S', 'PT1H30M'.");
+            }
+        }
+
         Matcher m = DURATION_PATTERN.matcher(text);
-        if (!m.matches()) throw new IllegalArgumentException("Not valid duration: " + text);
+        if (!m.matches())
+            throw new IllegalArgumentException(
+                    "Not valid duration: "
+                            + text
+                            + ". Use format like '1s', '2m', '1h 30m', or ISO-8601 e.g. 'PT1S'.");
 
         int days = (m.start(1) == -1 ? 0 : Integer.parseInt(m.group(1)));
         int hours = (m.start(2) == -1 ? 0 : Integer.parseInt(m.group(2)));
--- a/core/src/test/java/com/netflix/conductor/core/utils/DateTimeUtilsTest.java
+++ b/core/src/test/java/com/netflix/conductor/core/utils/DateTimeUtilsTest.java
@@ -27,6 +27,32 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class DateTimeUtilsTest {
 
+    private static Stream<Arguments> iso8601Durations() {
+        return Stream.of(
+                Arguments.of("PT1S", Duration.ofSeconds(1)),
+                Arguments.of("PT1M", Duration.ofMinutes(1)),
+                Arguments.of("PT1H", Duration.ofHours(1)),
+                Arguments.of("PT1H30M", Duration.ofSeconds(90 * 60)),
+                Arguments.of("P1D", Duration.ofDays(1)),
+                Arguments.of("P1DT2H3M4S", Duration.ofSeconds(86400 + 2 * 3600 + 3 * 60 + 4)),
+                Arguments.of("pt1s", Duration.ofSeconds(1)) // case-insensitive
+                );
+    }
+
+    @ParameterizedTest(name = "[{0}] is valid ISO-8601 duration")
+    @MethodSource("iso8601Durations")
+    public void shouldParseIso8601Duration(String input, Duration expectedDuration) {
+        assertThat(parseDuration(input)).isEqualTo(expectedDuration);
+    }
+
+    @ParameterizedTest(name = "[{0}] is invalid ISO-8601 duration")
+    @ValueSource(strings = {"Pnotvalid", "PT", "P1X"})
+    public void shouldRejectInvalidIso8601Duration(String input) {
+        assertThatThrownBy(() -> parseDuration(input))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Not valid ISO-8601 duration");
+    }
+
     private static Stream<Arguments> validDurations() {
         return Stream.of(
                 Arguments.of("", Duration.ofSeconds(0)),
@@ -102,6 +128,6 @@ public class DateTimeUtilsTest {
     public void shouldValidateDuration(String input) {
         assertThatThrownBy(() -> parseDuration(input))
                 .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage("Not valid duration: " + input);
+                .hasMessageContaining("Not valid duration: " + input);
     }
 }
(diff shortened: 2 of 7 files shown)


kunalrawat425:
Hi @v1r3n — this is a first-time contribution from a fork so CI hasn't triggered yet (0/0 checks). Could you approve the workflow run when you get a chance? Happy to address any review feedback. Thanks!

core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
```diff
@@ -35,8 +35,24 @@ public class DateTimeUtils {
...
+        // Accept ISO-8601 duration (e.g. "PT1S", "PT1H30M") as a convenience fallback.
+        if (text != null && text.toUpperCase().startsWith("P")) {
+            try {
+                return Duration.parse(text);
+            } catch (Exception e) {
+                throw new IllegalArgumentException(
+                        "Not valid ISO-8601 duration: "
+                                + text
+                                + ". Expected format e.g. 'PT1S', 'PT1H30M'.");
+            }
```
NicholasDCole, line 47: These `IllegalArgumentException` messages will get lost at higher level. 
Another option to do would be to make a new Exception class that is subclass of `IllegalArgumentException` -- this way we do not modify the contract of this method, which may cause regression elsewhere. 

When we have the new 
```
public class DurationParseException extends IllegalArgumentException
```
We may take this and directly put it here, as well as in the `isWaitTaskValid` method in the `WorkflowTaskValidator` class

NicholasDCole (changes requested):
Please resolve conflicts and address 1 comment. 
Changes look great

core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
NicholasDCole: Specifically, it needs to be added here as additional catch block -- before the `IllegalArgumentException` catch
image
On line 331

mp-orkes:
Thanks for taking these on — four real issues, and the scheduler verb fix in particular is exactly right (`SchedulerResource.java:121,130` are both `@PutMapping`, so the old `"get"` was definitely broken).

**The main ask: please split this into four separate PRs.** Right now all four fixes are in a single commit, which means we can't undo one without undoing all of them. That's a problem because the risk levels are very different: the `DateTimeUtils` change affects every WAIT task in every workflow, while the AgentSpan change edits three strings. If the duration change causes trouble in production, rolling it back would also roll back the scheduler fix that people actually need. Separate PRs also each get reviewed and merged faster than one bundled one.

Two things worth fixing as you split:

1. **`ApiReferencePage.tsx`** — the new version can leave users on a blank page. It opens Swagger in a new tab and then calls `navigate(-1)` to go back, but if there's nothing to go back to (someone bookmarked `/api-reference`, refreshed it, or opened it directly), `navigate(-1)` does nothing and the component renders `null`. Same outcome if a popup blocker stops the `window.open`, which is likely since it runs in a `useEffect` rather than directly from the click. Simpler fix: make the sidebar "API Docs" item a plain `target="_blank"` link to the Swagger URL and delete this intermediate page entirely.

2. **`DateTimeUtils.parseDuration`** — the new ISO branch accepts negative durations. `Duration.parse("PT-1S")` succeeds, and `WaitTaskMapper:102-106` then sets a wait timeout in the past and a negative `callbackAfterSeconds`. The old regex only matched digits, so this wasn't possible before. Worth rejecting zero/negative durations and adding a test for it.

Minor: #1304 isn't fully closed — "AgentSpan" still appears in `utils/flags.ts:95`, `utils/query.ts:491`, `utils/constants/route.ts:40`, and `pages/execution/AgentExecution/mockData.ts`. Fine to keep that PR scoped to the user-visible strings, just mention what's left.

The scheduler fix is the easiest to land first — it just needs a small test asserting the PUT, since nothing covers that today.

Closing this one in favor of the split PRs — please reference this thread from them so the context carries over. Looking forward to the four follow-ups.

PR #1372, opened 2026-07-22
fix: address issue #1334 - Agent SSE stream endpoint accepts nonexistent execution IDs and heartbeats forever — no existence check in openStream

### LINKED ISSUE #1334: Agent SSE stream endpoint accepts nonexistent execution IDs and heartbeats forever — no existence check in openStream
## Describe the bug

`GET /api/agent/stream/{executionId}` opens a successful SSE stream for any ID, including garbage. The connection stays open emitting `:heartbeat` indefinitely — no 404, no terminal error event, no server-side close.

A client with a stale or mistyped execution ID hangs forever waiting for events that can never arrive, and each such connection leaks an emitter registration and connection resources on the server.

## To Reproduce

```bash
curl -N http://localhost:8080/api/agent/stream/deadbeef-0000-0000
# -> 200, ":heartbeat" every interval, never ends
```

## Expected behavior

404 before upgrading to SSE when the execution doesn't exist. (For executions that are already terminal, emitting the final state and closing would also be more useful than an open-ended stream.)

## Root cause

[`AgentService.openStream`](https://github.com/conductor-oss/conductor/blob/main/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java#L1180) (L1180-1183) passes the ID straight to `streamRegistry.register(...)` without verifying the execution exists:

```java
public SseEmitter openStream(String executionId, Long lastEventId) {
    log.info("Opening SSE stream for execution {} (lastEventId={})", executionId, lastEventId);
    return streamRegistry.register(executionId, lastEventId);
}
```

Contrast with the other read endpoints (`/{executionId}/status`, `/executions/{executionId}/full`), which correctly 404 on unknown IDs.

## Details

- Conductor version: 3.32.0-rc (main @ c2439b7e6)
- Found during systematic E2E functional testing of the agent APIs

kunalrawat425:

Closes #1334


mp-orkes:

A couple of notes if you'd like to take another run at #1334:

PR #1374, opened 2026-07-22
fix: address issue #1322 - Late completion of a duplicate system task attempt overwrites outputData/endTime of an already-terminal task

### LINKED ISSUE #1322: Late completion of a duplicate system task attempt overwrites outputData/endTime of an already-terminal task

kunalrawat425:

Closes #1322


mp-orkes:

I'm closing this PR because it doesn't contain any changes — it's empty (0 files changed), so there's nothing to review. Per our [contribution guidelines](https://github.com/conductor-oss/conductor/blob/main/CONTRIBUTING.md), please open a PR only once you have code ready for review.

A couple of notes to help if you'd like to take another run at #1322:
- The guidelines suggest discussing the approach in our Slack channel first, since there are a few valid ways to implement terminal-state protection on the task update path.
- Please target `main`, and include a test that reproduces the late/duplicate completion overwriting an already-terminal task.

Feel free to reopen a new PR when you have the fix ready — happy to review it then.

PR #1375, opened 2026-07-22
fix: address issue #1321 - Async system tasks running longer than the queue unack window are executed twice (duplicate LLM_CHAT_COMPLETE provider calls)

kunalrawat425:
I have started working on this issue.

Closes #1321

PR #1379, opened 2026-07-22
fix: address issue #1313 - SUB_WORKFLOW auto-recovery serializes retries on one execution-lock key, causing contention and an INFO log flood on transient errors (v3.30.2)

### LINKED ISSUE #1313: SUB_WORKFLOW auto-recovery serializes retries on one execution-lock key, causing contention and an INFO log flood on transient errors (v3.30.2)
**Describe the bug**

Since [PR #973](https://github.com/conductor-oss/conductor/pull/973) (present in v3.30.x, absent in v3.23.0), a `SUB_WORKFLOW` task that hits a transient backend error while launching its child is left `SCHEDULED` and becomes eligible for sweeper "repair". The sweeper re-pushes it, a system-task worker dequeues it, and `execute()` retries the launch through the locked idempotent path `startWorkflowIdempotent(childId)` on a **deterministic** child workflow id. Under any lock contention, `tryLock` times out and throws `TransientException("Error acquiring lock when creating workflow")`, the task stays `SCHEDULED`, gets repaired again, and the loop repeats. Each iteration logs `A transient backend error happened` at `INFO`.

The auto-recovery itself is desirable (it removes the previous manual `SCHEDULED -> COMPLETED` intervention). The problem is that **every retry of a given stuck task, on every sweeper/worker thread across every replica, targets the same deterministic `childId`, hence the same lock key**. The retries do not spread across keys: they serialize on one lock and compete with each other (and with the child workflow's own `decide`/`sweep`, which also lock that id). Combined with the sweep cadence, this turns a brief transient backend blip into a self-amplifying contention storm and an `INFO` log flood, on an otherwise healthy backend.

The `waitTime` of the acquisition makes this worse. Acquisition uses a blocking `tryLock(waitTime=lockTimeToTry, leaseTime=lockLeaseTime)`, so each contending sweeper/worker thread **parks for up to `lockTimeToTry` before giving up and throwing**. With many threads across replicas racing the same key, a large share of the worker pool can sit blocked on that single lock at any moment, so the contention does not only flood logs, it also throttles overall system-task throughput. Note this is only a trade-off dial, not a fix: a shorter `lockTimeToTry` fails faster and logs more, a longer one blocks threads longer; neither removes the underlying serialization on one key. (Observed with `lockTimeToTry=2s`; the default is `500ms`.)

The single change that matters is that `SubWorkflow.isAsync()` flips from `false` to `true`, which makes the sweeper treat a `SCHEDULED` `SUB_WORKFLOW` as repairable (`isTaskRepairable` is gated on `isAsync()`), and the launch now goes through `startWorkflowIdempotent` (locked, deterministic child id) instead of the old fire-once `startWorkflow`.

The transient error path (leave `SCHEDULED`) and the sweeper-driven retry (`execute()`) in v3.30.2:

```java
// SubWorkflow.start() (v3.30.2): a transient launch error leaves the task SCHEDULED
} catch (TransientException te) {
    task.setStatus(TaskModel.Status.SCHEDULED);
    task.addOutput(SUB_WORKFLOW_LAUNCH_ERROR, te.getMessage());
    LOGGER.info("A transient backend error happened ...", task.getTaskId(), workflow.toShortString(), name, te);
}

// SubWorkflow.execute() (v3.30.2): each revisit retries the locked launch on the same childId
if (StringUtils.isEmpty(workflowId)) {
    if (task.getStatus() == TaskModel.Status.SCHEDULED) {
        start(workflow, task, workflowExecutor); // -> startWorkflowIdempotent(childId) -> acquireLock(childId)
    }
    return false;
}
```

The contention loop:

```mermaid
sequenceDiagram
    participant SW as Sweeper (8 threads x N replicas)
    participant Q as Task queue
    participant W as System-task worker (8 threads x N replicas)
    participant L as Redisson RLock (key = childId)
    participant B as Backend (transiently unavailable)

    Note over SW,B: SUB_WORKFLOW task stuck SCHEDULED, child id still empty
    loop every sweep cycle, until backend recovers
        SW->>Q: repair -> push(taskId)
        Q->>W: dequeue taskId
        W->>W: execute() -> start() -> startWorkflowIdempotent(childId)
        W->>L: tryLock(childId, waitTime=lockTimeToTry=2s)
        alt lock free but backend transient
            L-->>W: acquired
            W->>B: create child workflow
            B-->>W: TransientException
            Note over W: task stays SCHEDULED (INFO logged)
        else lock held by another thread on same childId
            L-->>W: false after 2s (contention)
            Note over W: throw TransientException "Error acquiring lock" (INFO logged)
        end
        Note over SW,W: task still SCHEDULED -> repaired again
    end
```

Relevant source (tags `v3.23.0` and `v3.30.2`):

- Throw site: [`WorkflowExecutorOps.startWorkflowIdempotent` (v3.30.2)](https://github.com/conductor-oss/conductor/blob/v3.30.2/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java#L2089-L2141) acquires the lock on the child id and throws `TransientException("Error acquiring lock when creating workflow")` when acquisition fails.
- Lock timing: [`ExecutionLockService.acquireLock` (v3.30.2)](https://github.com/conductor-oss/conductor/blob/v3.30.2/core/src/main/java/com/netflix/conductor/service/ExecutionLockService.java#L51-L74) calls `lock.acquireLock(id, lockTimeToTry, lockLeaseTime)`; [`RedisLock.acquireLock` (v3.30.2)](https://github.com/conductor-oss/conductor/blob/v3.30.2/redis-lock/src/main/java/com/netflix/conductor/redislock/lock/RedisLock.java#L64-L72) maps that to Redisson `tryLock(waitTime=lockTimeToTry, leaseTime=lockLeaseTime)`, which returns `false` (no exception) on pure contention.
- Async flip: `SubWorkflow` does not override `isAsync()` in [v3.23.0](https://github.com/conductor-oss/conductor/blob/v3.23.0/core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java#L76-L79) (inherits `false`), so the sweeper never repairs it and the launch is attempted once then left blocked. In v3.30.2 it returns `true`, enabling sweeper repair.
- Sweeper repair: [`WorkflowSweeper` (v3.30.2)](https://github.com/conductor-oss/conductor/blob/v3.30.2/core/src/main/java/org/conductoross/conductor/core/execution/WorkflowSweeper.java) gates `isTaskRepairable` on `isAsync()` and re-pushes the `SCHEDULED` task to its queue on each pass.

**Details**

- Conductor version: `v3.30.2` (regression introduced between `v3.23.0` and `v3.30.2`, via [PR #973](https://github.com/conductor-oss/conductor/pull/973))
- Persistence implementation: Postgres
- Queue implementation: `redis_cluster`
- Lock: Redis (Redisson `RLock`), `conductor.app.workflow-execution-lock-enabled=true`, `lockTimeToTry=2s`, `lockLeaseTime=60s`
- Topology: 2 replicas running system tasks, `sweeperThreadCount=8`, `systemTaskWorkerThreadCount=8` per replica

**To Reproduce**

1. Run Conductor v3.30.2 with the Redis workflow-execution lock enabled, several sweeper/system-task threads, and at least one replica.
2. Run workflows that launch `SUB_WORKFLOW` tasks.
3. Induce a transient backend error during sub-workflow launch (for example brief lock unavailability or contention on the execution lock).
4. Observe the same `SUB_WORKFLOW` task id logging `A transient backend error happened` repeatedly within seconds, each accompanied by `TransientException: Error acquiring lock when creating workflow`, until the transient condition clears.

**Code to reproduce (paste relevant snippet):**

```
(No standalone snippet; see the source references above. The condition is triggered by a
transient failure on the execution lock during SUB_WORKFLOW launch, not by workflow content.)
```

**Expected behavior**

Recovery of a `SCHEDULED` `SUB_WORKFLOW` after a transient launch error should not serialize all retries, from all threads and replicas, on a single deterministic lock key, nor retry at sweep cadence without coordination. A transient launch error should recover without producing a lock-contention storm and an `INFO` log flood on an otherwise healthy backend.

**Screenshots**

N/A

**Additional context**

Observed in production: after upgrading `v3.23.0 -> v3.30.2` on an unchanged workload, occurrences of `A transient backend error happened` increased roughly 15x. The Redis cluster was healthy throughout (cluster state OK, low CPU and memory, no evictions), and the exception is always `Error acquiring lock when creating workflow`, i.e. the lock path itself rather than a keyspace or persistence failure.

Possible directions (keeping the recovery behavior):

- **Avoid re-locking when a launch is already in flight.** Since the child id is deterministic, repair could check whether the child already exists or is being created before taking the lock, instead of every thread racing the same `tryLock`.
- **De-duplicate repair across replicas.** With N replicas x several sweeper threads all repairing the same `SCHEDULED` task, the `queueDAO.containsMessage` guard is racy; a stronger single-flight guard would cut redundant relaunches.

kunalrawat425:

Closes #1313


mp-orkes:

A couple of notes if you'd like to take another run at #1313:

PR #1380, opened 2026-07-22
fix: address issue #1309 - FORK_JOIN + EXCLUSIVE_JOIN passes registration but always fails at runtime

### LINKED ISSUE #1309: FORK_JOIN + EXCLUSIVE_JOIN passes registration but always fails at runtime
## Summary

A workflow definition that places an EXCLUSIVE_JOIN immediately after a FORK_JOIN
registers successfully via `PUT /api/metadata/workflow` but **always** fails at
runtime with a confusing error. The validator should catch this at registration time.

## Runtime Error

```
Fork task definition is not followed by a join task.  Check the blueprint
```

## Root Cause

`ForkJoinTaskMapper.java` validates the task following FORK_JOIN with:

```java
if (joinWorkflowTask == null ||
    !joinWorkflowTask.getType().equals(TaskType.JOIN.name())) {
    throw new TerminateWorkflowException(
        "Fork task definition is not followed by a join task...");
}
```

`TaskType.JOIN.name()` is the literal string `"JOIN"`. EXCLUSIVE_JOIN is a
different `TaskType` and never matches, so this check always fails when
EXCLUSIVE_JOIN follows FORK_JOIN.

The workflow **registration** validator does not perform this check, so the
invalid definition is accepted and stored. It only fails — with a non-obvious
`TerminateWorkflowException` — when the workflow actually runs.

## Design Note

EXCLUSIVE_JOIN is semantically designed for SWITCH branches, where exactly one
branch executes and EXCLUSIVE_JOIN completes as soon as that branch finishes.
FORK_JOIN runs **all** branches in parallel and requires JOIN to wait for all of
them. Using EXCLUSIVE_JOIN after FORK_JOIN is almost certainly a mistake.

Whether EXCLUSIVE_JOIN-after-FORK should be rejected or (optionally) supported
as "complete when any branch finishes" is a design decision. Either way, the
current behavior — silent acceptance at registration, cryptic failure at runtime
— is incorrect.

## Fix

Add this check to the workflow definition validator so it returns a 4xx error at
registration time with a human-readable message:

> `EXCLUSIVE_JOIN` cannot follow `FORK_JOIN`. Use `JOIN` to wait for all fork
> branches, or use `SWITCH` + `EXCLUSIVE_JOIN` for conditional branching.

## Files

- `core/src/main/java/com/netflix/conductor/core/execution/mapper/ForkJoinTaskMapper.java`
- Workflow registration validator (`WorkflowTaskTypeConstraint.java` or equivalent)

kunalrawat425:

Closes #1309


mp-orkes:

A couple of notes if you'd like to take another run at #1309:

PR #1387, opened 2026-07-22
Fix mysql boot crashes

kowser-orkes:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

- Core Flyway and the scheduler's Flyway share the `conductor` schema (separate history tables).
  Whichever migrates second sees a non-empty schema, no history table → mysql throws
  `FlywayException` on boot.
- Fix: `baselineOnMigrate(true)` + `baselineVersion("0")` on core Flyway for mysql (new
  `FlywayConfigurationCustomizer`), postgres (had `baselineOnMigrate` only), sqlite (had neither).
- Also fixes `MySQLSchedulerConfiguration`'s unqualified `RetryTemplate` injection
  (`NoUniqueBeanDefinitionException` with 3 candidate beans) via `@Qualifier("mysqlRetryTemplate")`,
  same pattern postgres already used.

Issue #

Limitations
----

The dual-Flyway-chain design (core migrations + scheduler migrations, same schema, two independent
`FlywayConfigurationCustomizer`/bean setups) is not removed by this PR — `baselineVersion("0")` just
makes the existing race harmless regardless of which chain runs first. The two chains still exist,
still migrate independently, and still depend on their migration numbering never overlapping.

Alternatives considered
----

- `@DependsOn` to force scheduler-Flyway after core-Flyway instead of baseline-0 — fixes ordering
  but leaves an implicit ordering contract instead of an order-independent fix.

### DIFF
--- a/mysql-persistence/src/main/java/com/netflix/conductor/mysql/config/MySQLConfiguration.java
+++ b/mysql-persistence/src/main/java/com/netflix/conductor/mysql/config/MySQLConfiguration.java
@@ -22,6 +22,7 @@ import org.conductoross.conductor.mysql.dao.MySQLSkillPackageDAO;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration;
+import org.springframework.boot.autoconfigure.flyway.FlywayConfigurationCustomizer;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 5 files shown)


nthmost-orkes:
In `CassandraRateLimitingDAO.exceedsRateLimitPerFrequency`, the SELECT COUNT then INSERT are two separate Cassandra operations — can you check that we're not introducing a new race condition here? Under concurrent dispatch, two threads could read the same count, both find it below the limit, and both insert, exceeding the rate limit. The Redis impl avoids this with atomic operations. Curious whether a **Cassandra lightweight transaction (LWT)** is needed, or if the over-admission behavior here is acceptable and worth documenting.

nthmost-orkes (approved):
Seems to fix the build, if this is @v1r3n 's advice then let's merge it -- please read my non-blocking comment tho

kowser-orkes:
@nthmost-orkes only kept the mysql changes on this PR. cassandra/file-storage changes are addressed on upcoming PRs

manan164 (approved):
Approving to unblock — the baseline-0 fix is correct and order-independent, and the @Qualifier fix matches the postgres pattern (verified 3 unqualified RetryTemplate candidates in a mysql server context).

That said, the ideal fix is a single Flyway bean per database type — scheduler migrations join the main chain (renumbered to a reserved range, idempotent DDL so dual-chain-era DBs upgrade cleanly), making the race structurally impossible instead of harmless. Worth a follow-up issue; this PR shouldn't wait on it.

PR #1401, opened 2026-07-27
fix #1322 — first-terminal-write-wins guard on updateTask

### LINKED ISSUE #1322: Late completion of a duplicate system task attempt overwrites outputData/endTime of an already-terminal task

mp-orkes:
Defense-in-depth for #1322. 

**#1369 removed the main cause** - duplicate execution of async system tasks. The worker/REST path already rejects terminal updates (`WorkflowExecutorOps.java:944`). 

What's left is a narrow race: `AsyncSystemTaskExecutor` checks `isTerminal()` when it loads the task (`:91`) but its `finally { updateTask(task) }` (`:245`) is unguarded — so when #1369's overrun branch marks a task `TIMED_OUT` while the original blocking `start()` is still running, that run's late return overwrites the terminal record.

## What changed

`ExecutionDAO.updateTask` becomes first-terminal-write-wins: once the stored task for a `taskId` is terminal, later writes are rejected atomically. Engine paths that legitimately re-write a terminal task — rerun/retry, the `decide()` re-persist, sub-workflow sync, reconciliation — use a new `forceUpdateTask` bypass. Redis enforces it with a Lua script plus a hash-tagged marker key; the bypass clears the marker on a non-terminal write so rerun/retry can reopen a task.

**Redis only.** Postgres/MySQL/SQLite/Cassandra inherit an unguarded default and keep last-write-wins.

## How to test

`./gradlew :conductor-core:test :conductor-redis-persistence:test` — `RedisExecutionDAOTest` covers the guard against a real `redis:7-alpine` container (the in-memory backend no-ops scripting, so the guard is inert there).

## Notes

- `decide()` uses the bypass deliberately: it re-persists terminal tasks on nearly every workflow, so guarding it would drop those writes.
- `updateTask` stays `void` — rejections are logged, not signalled to callers.
- Design doc: `docs/design/2026-07-27-issue-1322-first-terminal-write-wins.md`.

**Open question:** with #1369 already covering the main cause, is this worth landing at all? If yes, does the Postgres guard land here or as a follow-up?

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
+++ b/core/src/main/java/com/netflix/conductor/core/dal/ExecutionDAOFacade.java
@@ -550,6 +550,20 @@ public class ExecutionDAOFacade {
      *     payload fails.
      */
     public void updateTask(TaskModel taskModel) {
+        doUpdateTask(taskModel, false);
+    }
+
+    /**
+     * Same as {@link #updateTask(TaskModel)} but bypasses the first-terminal-write-wins guard. Only
+     * for trusted engine state-machine paths that legitimately re-write a terminal task.
+     *
+     * @param taskModel the task to be updated in the data store
+     */
+    public void forceUpdateTask(TaskModel taskModel) {
+        doUpdateTask(taskModel, true);
+    }
+
+    private void doUpdateTask(TaskModel taskModel, boolean force) {
         if (taskModel.getStatus() != null) {
             if (!taskModel.getStatus().isTerminal()
                     || (taskModel.getStatus().isTerminal() && taskModel.getUpdateTime() == 0)) {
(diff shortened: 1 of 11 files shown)


mp-orkes:
@NicholasDCole FYI, this is a work in progress until we merge
- https://github.com/conductor-oss/conductor/pull/1369
- https://github.com/conductor-oss/conductor/pull/1402

v1r3n (approved):

mp-orkes:
https://github.com/conductor-oss/conductor/pull/1369 removed the main cause. What's left is a small narrow race in `AsyncSystemTaskExecutor`. TOCTOU.

PR #1412, opened 2026-07-28
feat(file-storage): replace LOCAL storage type with CONDUCTOR

v1r3n:
## Problem

The `LOCAL` backend handed clients a `file:` URI naming a path on the **server's** filesystem:

```java
// LocalFileStorage.java:95
return baseDirectory.resolve(storagePath).toAbsolutePath().toUri().toString();
```

Only a co-located client can honour that, and `LocalFileTransferAdapter` rejects any other scheme outright. So the zero-infrastructure default was unusable in the normal deployment — a containerized server with workers elsewhere. The worker writes bytes to its own `/tmp/...`, the server stats the container's `/tmp/...`, finds nothing, and `confirmUpload` throws.

Observed directly: three `FileStorageE2ETest` cases fail against the `conductor-server` container while passing against a host JVM, purely because of which filesystem each side sees.

## Change

Replace `LOCAL` with `CONDUCTOR`, making the server the transfer endpoint and returning an ordinary HTTPS URL:

```
PUT /api/files/content/{workflowId}/{fileId}
GET /api/files/content/{workflowId}/{fileId}
```

Bytes stream both ways — `HttpServletRequest.getInputStream()` in, `StreamingResponseBody` out — with no buffering of whole files.

**Existing SDKs need no change.** `FileClient.selectAdapter` misses `CONDUCTOR` in its adapter map and falls through to `GenericHttpFileTransferAdapter`, whose javadoc states its purpose exactly: *"conservative fallback for a server storage type newer than this SDK."* A dedicated adapter (for multipart) is an optional follow-up, not a prerequisite.

### Optional URL signing, default off

HMAC-SHA256 over a version-prefixed canonical string. This is **not** about keeping the URL secret — OSS has no auth, so anyone who can reach `download-url` can already reach the bytes. It is how a transfer client that deliberately strips Conductor credentials gets authorized on a deployment that *does* front `/api` with auth.

- Origin (scheme/host/port) is excluded from the canonical value, so proxy host rewriting does not invalidate a URL.
- `op` must match the HTTP method, so a download URL is not also an upload URL.
- Keys are a list: sign with the first, verify against all — that is the rotation story.
- Constant-time comparison via `MessageDigest.isEqual`.

### Production requires shared storage

`CONDUCTOR` stores bytes on the server filesystem. An HTTPS URL fixes *transfer*, not *storage locality* — behind a load balancer, node A writes an object node B cannot serve. Multi-node deployments must point `conductor.file-storage.conductor.directory` at a shared filesystem (NFS, EFS, `ReadWriteMany` PVC). Anyone unwilling to run one should use S3, GCS, or Azure Blob. Documented as a warning admonition rather than papered over.

## Also fixed

- **MySQL and Cassandra file-metadata DAOs had no Spring bean**, so `FileStorageServiceImpl` was unsatisfiable and `/api/files/**` unavailable on those databases. Both are now wired, conditional on `conductor.file-storage.enabled`.
- **`downloadContent` required the exact owner** while `getDownloadUrl` is family-accessible. Since for this backend that URL *is* the content endpoint, a sub-workflow received a download URL and was then 403'd on it — breaking parent/child file exchange. A test asserted this behaviour; it is inverted with a comment explaining why.
- **`FileResourceTest` mixed Mockito matchers with raw values** (`uploadContent(WORKFLOW_ID, FILE_ID, any())`), producing `InvalidUseOfMatchersException` and unfinished-stubbing leakage into adjacent tests — 3 defects, 5 failing tests.
- **The e2e test compared the SDK's `String` `storageType` against a server-side enum** — it would have compiled and always failed. Now asserts the wire value, keeping the e2e module black-box.

## Verification

```
:conductor-core:test :conductor-rest:test :conductor-file-storage:test
:conductor-server:test :conductor-common:test          BUILD SUCCESSFUL
compileJava compileTestJava spotlessCheck              BUILD SUCCESSFUL
```

**Not yet verified:** the container e2e run — the acceptance criterion in the design doc — because the servers available to me were still running the previous build. `FileStorageE2ETest` should now pass against a containerized server with no bind mount, using the current SDK. That needs confirming before merge.

Cassandra is wired but has never been executed against a live Cassandra.

## Scope

Deliberately excluded from this branch, though present in the local working tree: a Gradle 8.14.3 → 9.1.0 upgrade with its shadow-plugin and toolchain fallout, the `GROK_API_KEY` fallback removal in `application.properties`, commented-out test invocations in `run_tests-mysql.sh` / `run_tests-cassandra-es7.sh`, and an unrelated `ConcurrentExecLimitTests` timeout change. None belong in a file-storage PR.

Design: `docs/design/file-storage-proxy-transfer.md`


kowser-orkes (approved):

core/src/main/java/org/conductoross/conductor/core/storage/FileStorageServiceImpl.java
NicholasDCole, line 63: Do we need to ensure that contentType is not blank here adding a 
@NotBlank annotation in our FileUploadRequest to ensure that cannot come as null? 

In downloadFileContent() method, it appears we might run into an exception if this is null

v1r3n: will do a follow up PR

PR #1419, opened 2026-07-29
fix: release concurrency-limited successors from the facade, not each DAO

v1r3n:
## Problem

A task bound by `concurrentExecLimit` frees a slot when it reaches a terminal state. [#1105](https://github.com/conductor-oss/conductor/pull/1105) added logic to reset the next queued message so a postponed sibling stops waiting out `taskExecutionPostponeDuration` — but it placed a copy in **all five** persistence `ExecutionDAO`s, each taking its own `QueueDAO`.

A persistence DAO cannot know which queue is live: the queue and the execution store are configured independently. With `conductor.db.type=mysql` and `conductor.queue.type=redis_standalone` — exactly what `docker-compose-mysql.yaml` + `config-mysql.properties` ship — `MySQLExecutionDAO` peeked the empty `queue_message` table while the message sat in Redis.

`QueueDAO.peekFirstIds` defaults to `List.of()`, so this failed **silently**. The sibling waited the full 60s.

Measured on a containerized MySQL server, completing the in-progress task and reading the sibling's Redis score:

```
after postpone            +58403 ms
after completing A        +45972 ms   ← unchanged; queue_message had 0 rows
poll for sibling          []
```

## Change

Move the release into `ExecutionDAOFacade`, which already holds the active `QueueDAO`, and delete all five copies along with their `queueDAO` fields and constructor parameters. No persistence module reaches into the queue layer any more, so any `db.type` × `queue.type` pair works by construction rather than by each backend having been wired correctly.

Details that matter:

- It runs **after** `executionDAO.updateTask` — that write is what records the slot as free, and a poll in the gap would re-postpone.
- A failed wakeup is logged, not propagated. The task update already committed, and losing a wakeup only costs the successor its postpone window.
- Backend-specific bookkeeping is untouched: Redis's `TASKS_IN_PROGRESS_STATUS`/`TASK_LIMIT_BUCKET` maintenance and Cassandra's `removeTaskFromLimit` stay where they belong.

## Verification

Against the rebuilt containerized MySQL + Redis server, same probe:

| | sibling's score vs. now |
|---|---|
| after postpone | `+58403 ms` |
| after completing A — before | `+45972 ms` (unchanged) |
| after completing A — after | **`−2042 ms`**, polls immediately |

`ConcurrentExecLimitTests`, `-DSERVER_ROOT_URI=http://localhost:8127/api`:

```
tests="2" skipped="0" failures="0" errors="0"
Polled for B for 42 ms          (5s allowance, 60s postpone window)
```

Its first case dropped from **130.5s → 7.4s** — incidental proof the wakeup had been inert on MySQL all along, waiting out full postpone cycles.

Unit suites for every touched module plus `spotlessCheck`: `BUILD SUCCESSFUL`.

## Test changes

`ConcurrentExecLimitTests` gains a precondition. It now skips, with an explanation, when `taskExecutionPostponeDuration` is not at least 4× its poll window — otherwise the sibling becomes pollable by *natural expiry* and the test passes whether or not the release works. That was live: with the 5s postpone configured on one of our servers and a 5s assertion window, it could not distinguish the fix from the bug.

The two DAO-level tests move from `RedisExecutionDAOTest` to `ExecutionDAOFacadeTest`, which also gains coverage for the no-concurrency-limit path and for a queue failure not failing the task update.

## Follow-up worth considering separately

`QueueDAO.peekFirstIds` still defaults to `List.of()`. It is much less dangerous now — one caller, always the active queue — but that default is what turned a wiring mistake into a silent 60s stall instead of an error. Making it throw, as `ExecutionDAO.exceedsInProgressLimit` does, would surface the next such mistake immediately. Not done here because it changes behaviour for any custom `QueueDAO`.


mp-orkes (approved):
Approving — the production change is a strict improvement and holds up independently of the tests: the facade injects the same unqualified `QueueDAO` that `ExecutionService` postpones with, and every terminal task update in `core`/`rest` reaches `executionDAOFacade.updateTask`. Nothing in the main diff needs to change.

Two test-side findings I'd like fixed in this PR (both small), and two notes.

1. **[BLOCKING-ish, please fix here]** `ConcurrentExecLimitTests.java:355` — the new skip-guard can never fire. `/admin/config` returns `System.getProperties()` (`ConductorProperties.java:599`), not Spring config. Docker passes config via a properties file (`docker/server/bin/startup.sh` → `-DCONDUCTOR_CONFIG_FILE=...`), so `conductor.app.taskExecutionPostponeDuration` is never a system property → the helper always reads 60 → the `assumeTrue` at :222 always passes. Including against `config-postgres.properties:26`, which sets `10s`. Drop the guard and pin the duration in the e2e server config, or read the value from something that actually reports it. Worth fixing here rather than later because the description cites this guard as the thing that fixed a live false-green.

2. **[MAJOR]** `ExecutionDAOFacade.java:563-566` — the comment calls the ordering load-bearing, but nothing tests it. Swap the two lines and all four new tests still pass. An `InOrder` verify closes it.

3. **[MAJOR, not blocking this PR]** Nothing in PR CI exercises the moved release. The e2e matrix is empty except on `workflow_dispatch`/`schedule` (`ci.yml:328`) and the `e2e` check here is SKIPPED; the new unit tests mock `QueueDAO`, which is by construction the correct queue, so they can't fail for the reason the original bug existed. Repo-wide CI gap more than this PR's problem, but worth knowing the fix ships unexercised.

4. **[MINOR]** `WorkflowTestService.java:135` still calls `executionDAO.updateTask` directly with a terminal status, bypassing the facade, so that path loses the release. Negligible on a simulation harness, but the move isn't quite lossless as described.

The follow-up named in the description — making `QueueDAO.peekFirstIds` throw instead of defaulting to `List.of()` — is the right call, and finding 1 is the same failure mode repeating: a silent default standing in for a value that was never there.

PR #1420, opened 2026-07-29
fix(ai): default agentType(); fix agent-client registration and OkHttpClient wiring

manan164:
Three fixes for API and wiring problems introduced by #1358, plus one dead property left by #1413. **6 files, +143/−8** — five fixes and a test.

## 1. `ConductorAgentClient.agentType()` is a breaking interface change

#1358 added `agentType()` as an **abstract** method on the public `ConductorAgentClient` interface. That's source- and binary-breaking in a minor release: every external implementation — SDK-backed worker adapters, and any host supplying its own agent control plane — stops compiling until it adds the method, even though `"conductor"` is the only value such an implementation could reasonably return.

Now defaulted to `"conductor"`; clients backing another runtime override. The literal is inlined rather than referencing `A2AService.AGENT_TYPE_CONDUCTOR` (same value, asserted in the javadoc) to keep this boundary interface import-free, as it is today.

## 2. `BedrockAgentClient` / `AzureFoundryAgentClient` can register without their dependency

Both were added as plain `@Component` with no condition, unlike `ServiceConductorAgentClient` which is gated on `conductor.integrations.ai.enabled`. All three require `CredentialResolutionService`, which carries that flag — so leaving two unconditional means a client can exist while the service it needs does not.

On a stock server it's masked: `AgentSpanAutoConfiguration`'s scan shares the flag, so the clients are only discovered when the service also exists. It stops being masked for a host that component-scans the runtime package itself:

```
azureFoundryAgentClient
  -> NoSuchBeanDefinitionException: ...credentials.CredentialResolutionService
```

Both now carry the same condition as `ServiceConductorAgentClient`. Stock behaviour is unchanged — they activate exactly when they do today — and no new property is introduced, so every combination stays reachable: **bedrock only, azure-foundry only, both, or neither**, decided by a workflow's `agentType` rather than by configuration.

That works because neither client resolves credentials at construction. Bedrock reads `credentialRef` inside `buildRuntimeClient` (falling back to the default AWS credential chain), Azure Foundry inside its token exchange — both per request. A missing key cannot break startup; it surfaces only if a workflow actually routes to that runtime. Both javadocs now say so, since "why is this bean registered when I never configured Bedrock" is the obvious question.

`BedrockAgentClient`'s javadoc claimed activation by `conductor.ai.bedrock-agent.enabled`, a property that appears nowhere else in the repo and was never wired; it now names the flag actually used. Real per-runtime opt-in flags would be a reasonable follow-up, but they'd need config metadata and docs, and they aren't required to make the combinations work.

## 3. Unqualified `OkHttpClient` in `AzureFoundryAgentClient`

The one thing that could still break the both-registered case. Only one such bean exists on a stock server (`conductorAiHttpClient`), so it resolves today — but any host defining a second `OkHttpClient` bean fails that client with `NoUniqueBeanDefinitionException`, taking the whole context down rather than just that runtime. Now bound explicitly to `conductorAiHttpClient`, the client the AI module publishes for this purpose.

## 4. Dead `agentspan.embedded` property in a test

#1413 folded `agentspan.embedded` into `conductor.integrations.ai.enabled`. `ConductorAgentEndToEndTest` still set the old flag alongside the new one — the last reference in the repo to a property nothing reads. Removed; `git grep agentspan.embedded` is now empty.

## Verification

- `./gradlew :conductor-agentspan:test :conductor-ai:test` — **0 failures**
- `:conductor-test-harness:compileTestJava` clean
- Spotless clean
- End-to-end downstream check: published this branch to a local Maven repo and consumed it from a host that component-scans the runtime package. Before, that host needed a scan exclusion for both clients *and* its own `agentType()` override; with this PR it compiles and boots clean with **neither**, under AI enabled and disabled.

## 5. `agentType` collisions were silent

`A2AWorkers` indexed clients with `agentClients.put(agentType, client)`, so two clients claiming one type last-one-wins on discovery order and one vanishes with no signal. Before #1358 the single-client injection raised `NoUniqueBeanDefinitionException` — so that traded a loud failure for an invisible one, and **fix 1 above makes it easier to hit**, since an implementation that forgets to override `agentType()` now matches the built-in client. Keeps the first registration and logs which client was ignored, on both the Spring and SDK paths.

## Resolved, not changed

The `"AGENT (conductor) requires 'name'"` validation #1358 dropped from `ConductorAgentDelegate` is **deliberate and stays dropped**: the delegate now also serves bedrock and azure-foundry, whose identity comes from `rawConfig` / `assistantId` rather than a name — which is why the remaining messages were genericised to `"AGENT requires 'prompt'"`. Flagging it here so it isn't mistaken for an oversight.

## Verification

**New test.** `ExternalAgentClientGatingTest` covers the two external clients, which had none: absent when the AI integration is off; registered together with `CredentialResolutionService` when on; reporting their own agent types with **no credentials configured anywhere**; and Azure Foundry still resolving its `OkHttpClient` when a second `OkHttpClient` bean exists. That last case was confirmed to fail without the qualifier (`NoUniqueBeanDefinitionException ... found 2`) and pass with it — so it guards the fix rather than just accompanying it.

**Suites.** `:conductor-agentspan:test` 560 tests, `:conductor-ai:test` 750 tests, **0 failures**. `:conductor-test-harness:compileTestJava` clean. Spotless clean. CI runs `unit-test`, `build` and `test-harness` (~18 min, boots real servers and exercises agent registration / deployment / MCP discovery).

**Downstream, against this exact revision.** Published to a local Maven repo and consumed from a host that component-scans the runtime package. Before this PR that host needed a component-scan exclusion for both clients *and* its own `agentType()` override; with this PR it compiles across all modules and boots a full Spring context with **neither**, under AI integrations both enabled and disabled.

**Not covered:** no test exercises Bedrock or Azure Foundry against the real services — that needs credentials. This PR only changes when those beans register and how their HTTP client is bound, not their request paths.

### DIFF
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
@@ -30,6 +30,7 @@ import org.conductoross.conductor.ai.agent.credentials.OAuthTokenProvider;
 import org.conductoross.conductor.ai.agentspan.runtime.credentials.CredentialResolutionService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -57,8 +58,11 @@ import okhttp3.Response;
  *   <li>{@code endpoint} - the agentsEndpointUri for the AI Foundry project (optional if
  *       AZURE_FOUNDRY_ENDPOINT secret is set)
  * </ul>
+ *
+ * <p>Activated by {@code conductor.integrations.ai.enabled=true}, like the other agent clients.
  */
 @Component
+@ConditionalOnProperty(name = "conductor.integrations.ai.enabled", havingValue = "true")
 public class AzureFoundryAgentClient implements ConductorAgentClient {
 
     private static final Logger log = LoggerFactory.getLogger(AzureFoundryAgentClient.class);
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
@@ -30,6 +30,7 @@ import org.conductoross.conductor.ai.agent.credentials.OAuthTokenProvider;
 import org.conductoross.conductor.ai.agentspan.runtime.credentials.CredentialResolutionService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 
@@ -60,6 +61,9 @@ import okhttp3.Response;
  * </ul>
  *
  * <p>Activated by {@code conductor.integrations.ai.enabled=true}, like the other agent clients.
+ * Credentials are resolved per request from {@code credentialRef}, so the client registers whether
+ * or not Azure Foundry is configured; an unconfigured runtime fails only if a workflow routes to
+ * it.
  */
 @Component
 @ConditionalOnProperty(name = "conductor.integrations.ai.enabled", havingValue = "true")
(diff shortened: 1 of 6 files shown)


manan164:
Closing — the consolidation onto `conductor.integrations.ai.enabled` is the direction we want, so re-splitting the flag is the wrong fix. The embedding host follows the single flag instead, which needs no change here.

Two findings from this branch worth keeping, both from #1358 rather than #1413:

- `BedrockAgentClient`'s javadoc states *"Activated by `conductor.ai.bedrock-agent.enabled=true`"*, but no `@ConditionalOnProperty` was ever wired, so the bean registers unconditionally. `AzureFoundryAgentClient` documents no flag at all and injects an unqualified `OkHttpClient`.
- `ConductorAgentEndToEndTest:85` still sets `agentspan.embedded=true`, now a dead property.

Happy to send either as a small standalone PR.

shaileshpadave (approved):

PR #1424, opened 2026-07-29
fix: send status-notifier webhook body as UTF-8 so Content-Length matches the payload

### LINKED ISSUE #1399: status-notifier webhook sends wrong Content-Length for non-ASCII output (StringEntity defaults to ISO-8859-1)
## Describe the bug

With the status-notifier enabled (`workflow_publisher` / `task_publisher`), Conductor POSTs the workflow/task summary to the configured webhook. If any string in that summary contains a non-Latin-1 character (for example an en dash `–` U+2013 or a non-breaking space U+00A0), the receiver rejects the POST because the declared `Content-Length` header does not match the actual UTF-8 byte length of the body. The notification never lands, and the retry path just fails the same way. Pure-ASCII payloads always deliver fine.

The cause is that `RestClientManager` builds the request entity with the single-argument `new StringEntity(data)`. On the pinned httpclient 4.5.x that constructor defaults to ISO-8859-1, so the body bytes and the `Content-Length` are computed from a non-UTF-8 view of the string while the payload is actually UTF-8. Any multibyte character desyncs the two.

## To Reproduce

1. Point a status listener at a webhook receiver that validates the request body length against `Content-Length` (strict HTTP servers do this):
   ```
   conductor.workflow-status-listener.type=workflow_publisher
   conductor.status-notifier.notification.url=http://<receiver>
   conductor.status-notifier.notification.endpointWorkflow=events/workflow
   conductor.status-notifier.notification.subscribedWorkflowStatuses=COMPLETED
   ```
2. Run any workflow that terminates with an output field containing a multibyte UTF-8 character, e.g. `{"note": "in–network"}` (en dash) or a value with a non-breaking space (U+00A0).
3. The webhook POST is emitted but rejected by the receiver with a Content-Length mismatch. Change the same field to ASCII only and it delivers fine.

## Expected behavior

The webhook body is sent as UTF-8 with a `Content-Length` equal to its UTF-8 byte length, so notifications carrying non-ASCII output are delivered like any other.

## Root cause

[`RestClientManager.createPostRequest`](https://github.com/conductor-oss/conductor/blob/main/task-status-listener/src/main/java/com/netflix/conductor/contribs/listener/RestClientManager.java#L235):

```java
StringEntity entity = new StringEntity(data);
```

The single-arg `StringEntity(String)` constructor uses the default charset (ISO-8859-1) on httpclient 4.5.x, not UTF-8. `Content-Length` is derived from that encoding, so it disagrees with the UTF-8 body whenever `data` contains a character outside Latin-1. Both publishers go through this method, so `task_publisher` and `workflow_publisher` are both affected.

## Proposed fix

Pass an explicit UTF-8 content type:

```java
StringEntity entity = new StringEntity(data, ContentType.APPLICATION_JSON);
```

`ContentType.APPLICATION_JSON` is UTF-8, so the body bytes and `Content-Length` are both computed as UTF-8. This fixes the length mismatch and also preserves the characters (the ISO-8859-1 path replaces anything it cannot represent). Passing an explicit charset is the documented way to construct `StringEntity`. Happy to send a PR.

## Details

- Conductor: `main` (defect present in the current source)
- Module: `task-status-listener`, `org.apache.httpcomponents:httpclient:4.5.14`
- Affects: `workflow_publisher` and `task_publisher` status listeners

nikhiln64:
## Summary

Fixes #1399.

RestClientManager built the webhook entity with the single argument StringEntity constructor, which on the pinned httpclient 4.5.x encodes the body as ISO-8859-1. As soon as the workflow or task summary contains a multibyte character such as an en dash or a non breaking space, the declared Content-Length no longer matches the UTF-8 payload and strict receivers reject the POST. The retry path fails the same way, so the notification is lost.

The entity is now built with ContentType.APPLICATION_JSON, which encodes the body as UTF-8 and declares the charset in the Content-Type header. The manual Content-type header is removed since the entity now carries it.

## Testing

Added RestClientManagerTest covering a payload with an en dash: the entity's Content-Length must equal the UTF-8 byte length, the body bytes must round trip as UTF-8, and the declared content type must carry the UTF-8 charset. createPostRequest was widened to package visibility with @VisibleForTesting for this. Module suite passes with ./gradlew :conductor-task-status-listener:test.


nthmost-orkes:
Superseded by #1473, which fixes the same issue (#1399) with the same change to `RestClientManager` and is already approved. Consolidating on that one.

If I've closed this in error please respond — this is a quick cleanup sweep and I'm happy to reopen. Thanks for the contribution!

PR #1434, opened 2026-07-30
fix(mysql): register skill storage from its own configuration

manan164:
`MySQLSkillMetadataDAO` / `MySQLSkillPackageDAO` were declared inside `MySQLConfiguration`. A host that brings its own MySQL persistence layer excludes that class wholesale from its component scan to avoid duplicate DAO beans (orkes-conductor does, by an exact-name filter) and lost skill storage as collateral — leaving no MySQL implementation of `SkillMetadataDAO`/`SkillPackageDAO`, so substituting a Postgres one fails its `BYTEA`/`TIMESTAMPTZ` DDL against a MySQL `DataSource` and the context never refreshes.

- Moves both beans to a new `MySQLSkillDAOConfiguration` (`org.conductoross.conductor.mysql.config`, inside `Conductor.java`'s existing scan), gated on `conductor.db.type=mysql` + `@ConditionalOnMissingBean` so a host can still override and the Postgres pair stays mutually exclusive.
- Retries use core's `onTransientErrorRetryTemplate` instead of `mysqlRetryTemplate` — the latter is declared by `MySQLConfiguration`, so it doesn't exist in a host that excludes it. Slight behaviour delta for these two DAOs in the stock server; flagging it deliberately.
- Pure relocation otherwise — same DAOs, same `@DependsOn` on Flyway, same `V10__agentspan_skills.sql` tables. No test added; happy to add an `ApplicationContextRunner` case asserting both beans register under `db.type=mysql` if you'd like one.

### DIFF
--- a/mysql-persistence/src/main/java/com/netflix/conductor/mysql/config/MySQLConfiguration.java
+++ b/mysql-persistence/src/main/java/com/netflix/conductor/mysql/config/MySQLConfiguration.java
@@ -18,8 +18,6 @@ import java.util.Optional;
 import javax.sql.DataSource;
 
 import org.conductoross.conductor.mysql.dao.MySQLFileMetadataDAO;
-import org.conductoross.conductor.mysql.dao.MySQLSkillMetadataDAO;
-import org.conductoross.conductor.mysql.dao.MySQLSkillPackageDAO;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration;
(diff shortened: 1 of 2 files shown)


manan164:
Closing — not needed. orkes-conductor already has `conductor-mysql-persistence` on its compile classpath (`api` scope), so `MySQLSkillMetadataDAO`/`MySQLSkillPackageDAO` can be constructed directly by the host; only the `@Bean` wiring was unreachable, not the classes. Doing it host-side instead, which avoids needing a new release.

The underlying inconsistency is still real if anyone wants it later: skill DAOs are declared inside `MySQLConfiguration` / `PostgresConfiguration` / `SqliteConfiguration`, each depending on a `*RetryTemplate` declared by that same class, so a host excluding one of those configs loses skill storage as collateral.

PR #1441, opened 2026-07-31
fix(agentspan): stop emitting DO_WHILE nested inside DO_WHILE for requiredTools

ling-senpeng13:
> **Draft.** Structurally verified and unit-tested; not yet exercised end-to-end against a live server. See *Not verified* below.

## Problem

Conductor does not support a `DO_WHILE` directly inside another `DO_WHILE`. The `requiredTools` path emitted exactly that — the retry wrapper is a `DO_WHILE`, and the agent's ReAct loop it retries is also a `DO_WHILE`, handed straight to `loopOver`:

```java
WorkflowTask loop = buildDoWhile(loopRef, termCondition, loopTasks, loopInputs);   // DO_WHILE
...
WorkflowTask outerLoop = buildDoWhile(outerLoopRef, outerCondition, List.of(loop, checkTask), outerInputs);
                                                                              // ^^^^ nested DO_WHILE
```

So declaring an agent with `requiredTools` produces an unsupported workflow shape. Nothing in the caller's control causes it — it's emitted by the compiler.

**This is the only place in the compiler that nests loops.** There is a single `DO_WHILE` factory (`buildDoWhile`, the one `setType("DO_WHILE")`), and across all its call sites this is the one that passes a `DO_WHILE` into a loop body. Every other loop body is built from non-loop tasks.

## Fix

Use the supported shape — a `DO_WHILE` inside a `SUB_WORKFLOW` — which this compiler **already uses for swarm agents**. The agent loop is wrapped in a `SUB_WORKFLOW` with an embedded definition, so the outer retry loop sees a `SUB_WORKFLOW` rather than a `DO_WHILE`.

### The subtlety: variable scope

The loop body reads workflow **variables** — `_agent_state`, `_last_tool_results`, `_stop_requested`, `_signal_injection`, and `_human_feedback` when approval is on. A sub-workflow starts with an *empty* variable scope, so a naive wrap would leave those silently reading empty.

The inner definition therefore begins with a `SET_VARIABLE` seeding them from sub-workflow inputs, wired from the parent's variables. **Seeding alone is sufficient** because every `SET_VARIABLE` in this compiler sits *outside* the loop body — the loop only reads these, so nothing needs writing back to the parent.

The check task previously read the inner loop's output directly; it now reads it via the sub-workflow's `outputParameters` mapping.

## Behaviour

Unchanged otherwise — same retry bound of 3, same condition. The `requiredTools`-empty branch already emitted a bare loop and is untouched, so no nesting is introduced there.

## Tests

`testCompileWithRequiredTools` is updated to the new shape and additionally asserts the seeded variables, the input wiring from parent variables, and the output mapping.

It also asserts the invariant the whole shape exists for, via a recursive walk that descends into embedded sub-workflow definitions:

```java
assertNoDoWhileNestedInDoWhile(wf);
```

**I verified that assertion actually bites**: putting the old nesting back (`List.of(loop, checkTask)`) fails the test with *"DO_WHILE 'filing_agent_loop' is nested directly inside another DO_WHILE"*. So it's a real guard, not decoration.

`563 tests pass`, spotless applied.

## Not verified

No end-to-end run against a live server. The compiled shape is asserted structurally, but the runtime semantics of seeded variables across the sub-workflow boundary should be confirmed by the agent e2e suite before this leaves draft. Two things worth watching there specifically:

- whether `_agent_state` seeded per outer-retry-attempt behaves as intended (each attempt re-seeds from the parent's value, so state does not accumulate across attempts — matching the previous behaviour, where the inner loop re-read the same parent variable, but worth confirming)
- the sub-workflow's own timeout, inherited from `createWorkflow` (60s), relative to long agent loops

## Context

This shape is what motivated adding nested-`DO_WHILE` support to the engine in orkes-io/orkes-conductor#3819. If `requiredTools` stops producing it, that engine change may not be needed — which also avoids the task-reference suffix-collision surface raised in review there.

### DIFF
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/AgentCompiler.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/AgentCompiler.java
@@ -789,8 +789,57 @@ public class AgentCompiler {
         // Prefill tool calls: execute before the loop so results are in LLM context
         allTasks.addAll(prefill.tasks());
 
-        // Required tools enforcement: wrap loop + check in outer DO_WHILE
+        // Required tools enforcement: retry the agent loop until every required tool has
+        // been called (bounded at 3 attempts).
+        //
+        // The retry is an outer DO_WHILE and the agent loop it retries is itself a
+        // DO_WHILE. Conductor does not support a DO_WHILE nested directly inside another
+        // DO_WHILE, so the inner loop is wrapped in a SUB_WORKFLOW — the documented
+        // workaround, and the shape this compiler already uses elsewhere for swarm agents.
         if (config.getRequiredTools() != null && !config.getRequiredTools().isEmpty()) {
+            String innerWfRef = toRef(config.getName()) + "_required_tools_inner";
+
+            // The loop body reads workflow VARIABLES (_agent_state, _last_tool_results,
+            // _stop_requested, _signal_injection, and _human_feedback when approval is
+            // on). Variables are per-workflow, so a sub-workflow starts with none: seed
+            // them from sub-workflow inputs before the loop runs, mirroring the parent's
+            // init_state. Every SET_VARIABLE in this compiler sits OUTSIDE the loop body,
+            // so the loop only reads them — seeding is sufficient and no write-back to
+            // the parent is required.
+            Map<String, Object> seedVars = new LinkedHashMap<>();
+            for (String v : initVars.keySet()) {
+                seedVars.put(v, "${workflow.input." + v + "}");
+            }
+            WorkflowTask seedState = new WorkflowTask();
+            seedState.setType("SET_VARIABLE");
+            seedState.setTaskReferenceName(innerWfRef + "_seed_state");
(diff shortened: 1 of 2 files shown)


ling-senpeng13:
Closing — the approach does not work as implemented.

Wrapping the agent loop in a SUB_WORKFLOW breaks parent-scope task references: tasks *after* the loop (e.g. `<agent>_resolve_output`) reference the LLM task, which moves inside the sub-workflow. Conductor rejects the definition at registration:

```
ConstraintViolationException: registerWorkflowDef.arg0: taskReferenceName:
  e2e_java_marker_guard_agent_llm for given task: e2e_java_marker_guard_agent_resolve_output
  input value: llm_result
```

Reproduced against conductor-oss standalone built from this branch: `Suite8bGuardrailsExtended.test_tool_output_detected_by_guardrail` fails with `API error 400: Validation failed`.

I accounted for workflow-variable scope but not for post-loop tasks referencing in-loop task refs. Making this work would require exposing those refs through the sub-workflow's outputParameters and rewriting every parent-scope reference to them — materially larger than the one-site change proposed here, and not obviously the right trade.

Not pursuing further for now.

PR #1444, opened 2026-07-31
fix(azure-foundry): fix OAuth scope, API version, and multi-part content extraction

shaileshpadave:
## Summary

- **OAuth scope**: `management.azure.com` → `cognitiveservices.azure.com` — wrong audience was causing HTTP 401 on all Azure Foundry calls
- **API version**: `2025-05-01` → `2024-05-01-preview` — the GA version returns HTTP 404 on Azure resources that haven't been upgraded yet
- **Content extraction**: iterate content parts to find the first `type=text` instead of blindly taking `content[0]` — assistants with code interpreter (e.g. analyst) return an `image_file` part first, causing empty text output

## Test plan

- [x] Greeter agent returns greeting text
- [x] Summarizer agent returns bullet-point summary
- [x] Analyst agent (with code interpreter) returns text analysis, not empty string


shaileshpadave:
Closing — superseded by #1445 which is based on latest main.

PR #1447, opened 2026-07-31
feat(azure-foundry,bedrock): Entra ID auth modes, agent tools forwarding, and AssumeRole support

shaileshpadave:
Adds Azure AI Foundry and AWS Bedrock agent support to Conductor — discovery, listing, and execution via normal Conductor workflow + AGENT task. Also lays the conductor-oss hooks for OBO identity passthrough (enterprise-only feature).

## What's in this PR

**Agent discovery** — set a secret with an `endpoint` key for Azure or `region` for Bedrock; those agents appear automatically in the Agent Definitions list alongside native Conductor agents.

**Auth** — credential-based. Azure supports API key, service principal, managed identity, and default credential chain from the secret. Bedrock supports static keys and AssumeRole. This is the default auth model for both OSS and enterprise.

**OBO (On-Behalf-Of) hooks — enterprise only** — when a cluster is SSO-wired to Entra ID, a human's login identity can flow through to Foundry instead of the stored service credential. The conductor-oss changes here are the plumbing only:
- `useCallerIdentity` + `userAssertion` fields on `ConductorAgentStartRequest` — the flag and token carrier
- `@JsonIgnore callerEntraToken` on `WorkflowModel` — never exposed via API; set by the enterprise SSO layer at workflow start
- OBO exchange in `AzureFoundryAgentClient.buildAuthState()` — when both fields are present, exchanges the user's Entra token for a Foundry-scoped token via Microsoft's OBO endpoint; falls back to credential-based auth silently if either is missing
- All existing credential auth paths unchanged — OSS clusters are unaffected

**UI** — provider logos (Azure, Bedrock, Orkes) appear next to each agent name. Filter chips (All / Azure Foundry / Conductor) sit below the results count and include the provider logo.

## Tested

Agent listing, logo rendering, credential-based execution — verified against a live `shailesh-ai-foundry` instance (10 Azure agents + 1 native agent). OBO logic covered by unit tests (`AzureFoundryAgentClientTest.OboAuthTest` — 4 cases: OBO bearer, missing-token fallback, flag-off bypass, incomplete-creds fallback).


agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
```diff
@@ -317,85 +419,196 @@ private static ConductorAgentState toState(String azureStatus) {
...
+        if (endpoint.contains("inference.ml.azure.com")) return ML_INFERENCE_SCOPE;
+        if (endpoint.contains("services.ai.azure.com")) return FOUNDRY_SCOPE;
+        return DEFAULT_SCOPE;
     }
 
     private String resolveEndpoint(ConductorAgentStartRequest request) {
-        String endpoint = rawConfig(request, "endpoint");
-        if (StringUtils.isBlank(endpoint)) {
-            endpoint = credentialResolutionService.resolve("AZURE_FOUNDRY_ENDPOINT");
+        String url = request.getAgentUrl();
```
v1r3n, line 524: This removes the previously supported `rawConfig.endpoint` configuration path. Existing Azure workflows that provide the documented endpoint there, rather than in the newly added `agentUrl`, now fail unless a global `AZURE_FOUNDRY_ENDPOINT` secret happens to be configured. Preserve `rawConfig.endpoint` as a fallback and add a regression test before making `agentUrl` mandatory.

v1r3n, line 524: `agentUrl` is workflow-controlled and reaches `post`/`get` unchanged, where the configured `api-key` or Bearer token is attached. This permits a workflow to direct an Azure credential to an arbitrary host. Validate the scheme and permitted Azure endpoint host(s), with an explicit configurable allowlist if private endpoints are supported, before creating authenticated requests.

agentspan/src/test/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClientTest.java
```diff
@@ -0,0 +1,425 @@
...
+import okhttp3.OkHttpClient;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for the pure routing, parsing, transformation, and auth-detection logic in {@link
+ * AzureFoundryAgentClient}. No Azure credentials or network calls required.
+ */
+@ExtendWith(MockitoExtension.class)
```
v1r3n, line 49: The added Azure coverage is Mockito-only and exercises helper methods, not the new authenticated HTTP flows. Add a configuration-gated real Azure Foundry integration test covering an actual request/response path; the repository review rules require real-service coverage for this provider integration when credentials are available.

v1r3n (changes requested):
Changes requested.

---
Reviewer note (changes requested):
see comments

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClient.java
shaileshpadave: resolveEndpoint() now checks agentUrl then rawConfig.endpoint then AZURE_FOUNDRY_ENDPOINT secret, with a regression test in AzureFoundryAgentClientTest.ResolveEndpoint.

shaileshpadave: added validateAzureHost() with an explicit allowlist of four Azure domain suffixes (.openai.azure.com, .cognitiveservices.azure.com, .services.ai.azure.com, .inference.ml.azure.com).

agentspan/src/test/java/org/conductoross/conductor/ai/agentspan/runtime/service/AzureFoundryAgentClientTest.java
shaileshpadave: Added AzureFoundryAgentClientIT.java with gated on CONDUCTOR_SECRET_CONDUCTOR_AZURE_SP, hits live Azure, just ran and passed locally.

PR #1459, opened 2026-08-03
fix: slot release must not wake popped-but-unacked messages (floor-restricted peek)

manan164:
Targets #1419's branch — proposed increment before it merges.

## Hazard found while verifying #1419

The Redis queue is a single zset: `pop.lua` bumps a popped message's score to `now+unackTimeout` **in place**, so `peekFirstIds` (`ZRANGEBYSCORE 0 +inf`) can return a message another worker is executing *right now* — and an in-flight score (`now+30s`) sorts **ahead** of a postponed sibling (`now+60s`). `resetOffsetTime` on that member causes an immediate duplicate delivery, while the intended sibling stays stranded.

Reproduced live (MySQL store + Redis queue, real `RedisQueueDAO`):

| member | before release | after release |
|---|---|---|
| `inflight-msg` (popped, unacked) | `+30000 ms` | **`−1 ms` — redeliverable mid-execution** |
| `postponed-sibling` (the actual target) | `+599999 ms` | `+599698 ms` — untouched |

This is not introduced by #1419 — the same peek+reset has run on redis-store deployments since #1105 — but #1419 activates it on the SQL-store deployments where it was previously inert, so it's worth closing in the same change. Exposure: concurrency-limited task types whose messages sit unacked during execution (async system tasks for their whole run; worker tasks in the pop→ack window).

## Change

- `QueueDAO.peekPostponedIds(queueName, minDelayMillis, count)` — only messages postponed at least `minDelayMillis` into the future are eligible for release. Default delegates to `peekFirstIds` (the SQL queues already exclude in-flight rows via `popped = false`).
- `BaseRedisQueueDAO` overrides it with a score floor (`ZRANGEBYSCORE now+minDelay +inf`).
- The facade release uses it, floor from new `conductor.app.concurrencySlotReleaseMinDelay` (default 35s — above the 30s unack windows).

Trade-off, stated plainly: a sibling whose remaining postpone is already below the floor is left to natural expiry (bounded, benign) — a missed wakeup costs seconds; a duplicate delivery of an in-flight task cannot be taken back.

## Verification

- `RedisQueueDAOPeekPostponedTest` (real Redis via testcontainers): floor-less peek returns the in-flight message first (the hazard); floored peek returns only the postponed sibling; in-flight-only queue → empty.
- `ExecutionDAOFacadeTest` 22/22, including new no-reset-when-only-in-flight case.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
+++ b/core/src/main/java/com/netflix/conductor/core/config/ConductorProperties.java
@@ -97,6 +97,15 @@ public class ConductorProperties {
     @DurationUnit(ChronoUnit.SECONDS)
     private Duration taskExecutionPostponeDuration = Duration.ofSeconds(60);
 
+    /**
+     * Only queue messages postponed at least this far into the future are woken up when a
+     * concurrency-limited task frees a slot. Must exceed every unack/visibility window in use
+     * (default system-task and worker pop windows are 30s): on queues that keep in-flight messages
+     * visible, a smaller value could wake a message another worker is still executing.
+     */
+    @DurationUnit(ChronoUnit.SECONDS)
+    private Duration concurrencySlotReleaseMinDelay = Duration.ofSeconds(35);
+
     /** Used to enable/disable the indexing of tasks. */
     private boolean taskIndexingEnabled = true;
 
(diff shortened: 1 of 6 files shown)


manan164:
Folded directly into #1419's branch instead.

PR #1462, opened 2026-08-03
fix: resolve persistence concurrency-limit wiring and add /api/health endpoint

v1r3n:
- [x] Bugfix
  - [ ] Feature
  - [ ] Refactoring (no functional changes, no api changes)
  - [ ] Build related changes
  - [ ] WHOSUSING.md
  - [ ] Other (please describe):

  NOTE: Please remember to run ./gradlew spotlessApply to fix any format violations.

  Changes in this PR

  - Fixes persistence configuration so ExecutionDAO receives the QueueDAO interface bean, allowing concurrency-limit behavior to initialize correctly for MySQL and Postgres.
  - Adds /api/health as an alias for the existing /health endpoint, supporting clients and deployments that use the API-prefixed route.

  Issue #

  N/A

PR #1463, opened 2026-08-04
test: await async decide in SubWorkflowRestartSpec setup instead of racing it

manan164:
- [x] Bugfix (test-only: CI flake)

## Failure

[CI run 30872035136](https://github.com/conductor-oss/conductor/actions/runs/30872035136) (on #1462's branch, unrelated to its change):

```
SubWorkflowRestartSpec > Test restart on the leaf in a 3-level subworkflow FAILED
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
    at SubWorkflowRestartSpec.setup(SubWorkflowRestartSpec.groovy:125)
```

## Root cause

`setup()` read `tasks.get(1).subWorkflowId` immediately after completing task 1. The decide that schedules the SUB_WORKFLOW task normally runs inline with that completion, but is deferred to the background sweeper when the workflow lock is contended — under CI load the raw read outruns it. Same load-sensitive race class as the `WorkflowRetryTests`/`WorkflowRerunTests` hardening in 555d7d069; this spec carried one more instance of the pattern (both the root and mid-level stages — the root stage just failed as an assertion instead of an exception when it lost the race).

## Change (test-only)

Both stages now wait (`PollingConditions`, 30s ceiling) for the SUB_WORKFLOW task to exist and for its `subWorkflowId` to be populated before dereferencing. The waits deliberately tolerate the system-task coordinator having already started the task — the manual `asyncSystemTaskExecutor.execute` now fires only when the task is still `SCHEDULED`, which is why the old `find{SCHEDULED}`-then-`if` could be legitimately null.

Positive eventually-waits only: zero added time on passing runs.

## Verification

10/10 consecutive green local runs of the spec (each a fresh JVM/gradle execution).

### DIFF
--- a/test-harness/src/test/groovy/com/netflix/conductor/test/integration/SubWorkflowRestartSpec.groovy
+++ b/test-harness/src/test/groovy/com/netflix/conductor/test/integration/SubWorkflowRestartSpec.groovy
@@ -21,6 +21,7 @@ import com.netflix.conductor.core.execution.tasks.SubWorkflow
 import com.netflix.conductor.test.base.AbstractSpecification
 
 import spock.lang.Shared
+import spock.util.concurrent.PollingConditions
 
 import static com.netflix.conductor.common.metadata.tasks.TaskType.TASK_TYPE_SUB_WORKFLOW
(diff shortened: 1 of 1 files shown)


manan164:
Consolidated into the single pipeline-fix PR per review preference.

PR #1464, opened 2026-08-04
test: re-enable 4 healed WorkflowRerunTests, refresh stale @Disabled reasons, harden load-flaky awaits

manan164:
- [x] Bugfix (test-only)

## Why

The retry/rerun/restart engine fixes (388fabfdc lineage, merged) silently repaired several behaviors that `WorkflowRerunTests` cases were `@Disabled` for — but nobody re-enabled the tests, and the stale disable reasons made the file look more broken than the engine is. Verified everything against a live server on current main.

## Changes (one file)

**Re-enabled — passing, 13/13 across repeated local full-suite runs:**
- fork-join rerun with DO_WHILE loop task (3 variants)
- SWITCH re-execution after rerun (sync-status variant)

**Still failing — `@Disabled` reasons rewritten to the accurate CURRENT failure modes** (each verified through 30–60s awaits; the awaited state genuinely never arrives):
- fork-join rerun: sibling branch not rescheduled (task list stays at 2)
- SUB_WORKFLOW-inside-FORK rerun ×2 (Ticket #7097): sibling branch child never spawns (`subWorkflowId` stays null)
- DO_WHILE rerun: task never re-decided from SCHEDULED to IN_PROGRESS
- SWITCH rerun: workflow completes without rescheduling the selected branch
- fork-join with wait/webhook/switch: never reaches the expected fork shape
- SWITCH-inside-DO_WHILE: flaky across consecutive runs

These are real engine gaps (orkes-parity, #7097 family) — the updated reasons + hardened diagnostics make the tests ready for whoever picks up that work: the racy one-shot reads (`subWorkflowId`, post-rerun snapshots) now await with explicit messages, so the failures report the gap directly ("sibling branch not rescheduled; tasks=2") instead of NPE/IndexOutOfBounds.

**Two load-flaky awaits in untouched enabled tests raised (5s/10s → 30s)** — both observed failing under repeated-run load while the suite was otherwise green; positive eventually-waits, zero cost on passing runs.

## Verification

Full `WorkflowRerunTests` against a live current-main server: 37 tests, 0 failures, 9 skipped (2 by-design + 7 documented gaps) — including 3/3 consecutive full-suite green runs with the final code, and the 4 re-enabled tests green in every run of every loop (13/13).

### DIFF
--- a/e2e/src/test/java/io/conductor/e2e/workflow/WorkflowRerunTests.java
+++ b/e2e/src/test/java/io/conductor/e2e/workflow/WorkflowRerunTests.java
@@ -449,7 +449,7 @@ public class WorkflowRerunTests {
 
     @Test
     @Disabled(
-            "Fork-join rerun from a completed branch task does not re-schedule sibling branches in conductor-oss")
+            "Rerun of a fork-join from a completed branch task does not re-schedule the sibling branch (task list stays at 2 for 30s; see the awaited diagnostic in the test)")
     @DisplayName("Check workflow fork join task rerun")
     public void testRerunForkJoinWorkflow() {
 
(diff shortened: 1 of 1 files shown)


manan164:
Consolidated into the single pipeline-fix PR per review preference.

PR #1474, opened 2026-08-04
fix(wait-task): accept ISO-8601 duration and fix misleading error message

### LINKED ISSUE #1310: WAIT task: misleading error 'date or duration is null' when ISO-8601 duration format is used

shaileshpadave:
## What

A WAIT task with `"duration": "PT1S"` (ISO-8601) failed with "Either date or duration is passed as null" — a message that implies the field was absent, when it was actually present but in an unsupported format.

Two changes:

1. **`DateTimeUtils.parseDuration()`** — after the custom format (`"1s"`, `"2m"`) fails to match, fall back to `java.time.Duration.parse()` so ISO-8601 durations like `"PT1S"`, `"P5D"`, `"PT1H30M"` work natively.

2. **`WorkflowTaskTypeConstraint`** — when `parseDuration()` still throws (genuinely invalid input), surface the actual error message instead of the misleading "passed as null" message.

Fixes #1310

## How to verify

1. Define a workflow with a WAIT task using ISO-8601 duration:
   ```json
   { "type": "WAIT", "name": "wait1", "taskReferenceName": "wait_ref",
     "inputParameters": { "duration": "PT1S" } }
   ```
2. Before fix: task fails immediately with "Either date or duration is passed as null".
3. After fix: task waits 1 second and completes normally.
4. To test the error message fix: use an invalid value like `"duration": "PT"` — the error now reads "Not valid duration: PT" instead of blaming null.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
+++ b/core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
@@ -36,13 +36,18 @@ public class DateTimeUtils {
 
     public static Duration parseDuration(String text) {
         Matcher m = DURATION_PATTERN.matcher(text);
-        if (!m.matches()) throw new IllegalArgumentException("Not valid duration: " + text);
-
-        int days = (m.start(1) == -1 ? 0 : Integer.parseInt(m.group(1)));
-        int hours = (m.start(2) == -1 ? 0 : Integer.parseInt(m.group(2)));
-        int mins = (m.start(3) == -1 ? 0 : Integer.parseInt(m.group(3)));
-        int secs = (m.start(4) == -1 ? 0 : Integer.parseInt(m.group(4)));
-        return Duration.ofSeconds((days * 86400) + (hours * 60L + mins) * 60L + secs);
+        if (m.matches()) {
+            int days = (m.start(1) == -1 ? 0 : Integer.parseInt(m.group(1)));
+            int hours = (m.start(2) == -1 ? 0 : Integer.parseInt(m.group(2)));
+            int mins = (m.start(3) == -1 ? 0 : Integer.parseInt(m.group(3)));
+            int secs = (m.start(4) == -1 ? 0 : Integer.parseInt(m.group(4)));
+            return Duration.ofSeconds((days * 86400) + (hours * 60L + mins) * 60L + secs);
+        }
+        try {
+            return Duration.parse(text);
+        } catch (java.time.format.DateTimeParseException e) {
+            throw new IllegalArgumentException("Not valid duration: " + text);
+        }
     }
 
     public static Date parseDate(String date) throws ParseException {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/netflix/conductor/core/utils/DateTimeUtils.java
```diff
@@ -36,13 +36,18 @@ public class DateTimeUtils {
...
+            int hours = (m.start(2) == -1 ? 0 : Integer.parseInt(m.group(2)));
+            int mins = (m.start(3) == -1 ? 0 : Integer.parseInt(m.group(3)));
+            int secs = (m.start(4) == -1 ? 0 : Integer.parseInt(m.group(4)));
+            return Duration.ofSeconds((days * 86400) + (hours * 60L + mins) * 60L + secs);
+        }
+        try {
+            return Duration.parse(text);
+        } catch (java.time.format.DateTimeParseException e) {
+            throw new IllegalArgumentException("Not valid duration: " + text);
+        }
```
nikhiln64, line 50: ## Context

This looks good. Accepting ISO 8601 durations through the Duration.parse fallback is clean, the regex and ISO paths do not overlap, and the new error message is a real improvement over the old date or duration is null wording. I also checked the message change at WorkflowTaskTypeConstraint.java:333 that passes e.getMessage() with the user text into buildConstraintViolationWithTemplate, and on Hibernate Validator 8 custom violation messages do not evaluate EL and an unbalanced brace only logs a warning, so there is no injection or interpolation crash there.

▎ Example: The Duration.parse fallback now accepts values the old regex never could, like a negative duration PT-1H or a zero or sub second duration PT0S or PT0.5S. The duration branch in WaitTaskMapper at 100 to 106 does not floor the result the way the until branch does at 114 to 117, so a negative duration sets callbackAfterSeconds to a negative value and a waitTimeout in the past, and the WAIT completes immediately instead of waiting.

## Suggestion

Flooring the computed seconds at zero in the duration branch, or rejecting non positive durations in parseDuration, would keep the two paths consistent. A couple of tests for the negative and zero and sub second cases, plus one assertion on the new constraint message, would round it out since the added tests only cover the happy path values.

## NOTE

None of this blocks the fix.

PR #1476, opened 2026-08-04
fix(#1437): strip whitespace from env-var API keys in AgentspanAIModelProvider

### LINKED ISSUE #1437: Provider API keys aren't trimmed/validated — trailing newline → cryptic "Unexpected char 0x0a in Authorization value"
### What happened

While testing agentspan agents against a local 3.32.0-rc.9 server, every OpenAI call failed at runtime with:

```
Task execution failed: Unexpected char 0x0a at 80 in Authorization value
```

`0x0a` is a newline. The server's `OPENAI_API_KEY` had a trailing `\n` (it was loaded from a file), so the auth header became `Bearer sk-…\n` and Java's HTTP client rejected it. A clean Anthropic key on the same server worked fine, so the path is otherwise healthy — this is purely the untrimmed key.

It's an easy state to land in: `export OPENAI_API_KEY=$(cat keyfile)` usually strips the newline, but plenty of common patterns don't (a `.env` line with a trailing newline, `echo`-ing into a file, some secret managers). The failure mode is cryptic and only visible at execution time.

### Root cause

Provider clients build the header as raw `"Bearer " + apiKey` with no sanitization — e.g. `ai/src/main/java/org/conductoross/conductor/ai/providers/openai/api/OpenAI*Api.java`. Keys are read either via each provider's `@ConfigurationProperties` (`OpenAIConfiguration`, `AnthropicConfiguration`, …) or from raw env vars through agentspan's `KnownProviderEnvVars` resolver. Nothing trims along the way.

### Angles to weigh (not prescribing a fix)

**Where a fix could live:**
- **At ingestion** — trim once where env/config → key is resolved (the `@ConfigurationProperties` setters and/or the `KnownProviderEnvVars` resolver). One place, and all ~13 providers inherit a clean key. Most robust, but it's a shared code path so worth a careful look.
- **At the header site** — `"Bearer " + apiKey.strip()` in the provider API classes. Localized and obvious, but it's ~13 providers × multiple call sites, so it's easy to miss one — better as a backstop than the primary fix.
- **Validate + surface** — have `/api/providers/status` report a key as `configured but invalid` (e.g. "contains whitespace") so the problem shows up at config time instead of mid-execution. Complements either of the above rather than replacing them.

**A judgement call for maintainers:** silently trimming *all* surrounding whitespace from a secret vs. rejecting a malformed key with a clear error message. Both are defensible — trimming is friendlier, rejecting is more explicit about "your key is wrong." Trailing-newline is the obviously-safe case; leading/embedded whitespace is more of a policy choice.

**Not a CLI concern:** worth noting the fix belongs server-side — for execution the key never passes through `conductor-cli`; the server reads it from its own env/credential store and builds the header. (The CLI's only useful role here would be reading a validated `/api/providers/status` in `doctor` so it warns early.)

### Repro

1. `export OPENAI_API_KEY="$(printf 'sk-yourkey\n')"` (note the trailing newline)
2. Start the server; run any agent/LLM task that hits OpenAI.
3. Task fails: `Unexpected char 0x0a … in Authorization value`.

Related: #1416 (self-hosted model base-url handling) is the same "provider credential/config handling is fragile" theme.

shaileshpadave:
## Summary

Fixes #1437 — trailing newlines in API keys read from environment variables (e.g. `export KEY=$(cat keyfile)`) caused a cryptic runtime error:

```
Task execution failed: Unexpected char 0x0a at 80 in Authorization value
```

## What changed

**`AgentspanAIModelProvider.getSystemEnv()`** now strips surrounding whitespace from `System.getenv()` results. The direct `System.getenv()` call inside `getModel()` is also routed through `getSystemEnv()` for consistency.

A new package-private `readRawEnv()` hook makes the trimming testable without PowerMock; existing tests that override `getSystemEnv()` are unaffected.

## How to verify

Start the server with a key that has a trailing newline:
```bash
export ANTHROPIC_API_KEY="$(printf 'sk-ant-yourkey\n')"
java -jar conductor-server.jar --agentspan.embedded=true --conductor.integrations.ai.enabled=true
```
Run any agent task — it should succeed instead of failing with `Unexpected char 0x0a`.

Verified locally: server started with `ANTHROPIC_API_KEY` ending in `0x0a`, agent `01_basic_agent.py` completed successfully.

New tests in `AgentspanAIModelProviderTest`:
- `getSystemEnv_stripsTrailingNewline`
- `getSystemEnv_stripsLeadingAndTrailingWhitespace`
- `getSystemEnv_returnsNullWhenEnvVarAbsent`


v1r3n (changes requested):
doing .strip() is not scalable across all the model providers.  I think its the user's responsibility to ensure api keys are trimmed.

v1r3n (approved):

PR #1479, opened 2026-08-05
fix: stamp/preserve createTime

### LINKED ISSUE #1450: UI: Created Time and Owner Email are blank on the Agent Definitions page
## Problem

On **Agent Definitions** (`/agents`), the **Created Time** and **Owner Email** columns are blank.

- **Owner Email** is empty for *every* row, including agents created seconds earlier.
- **Created Time** is empty for older agent definitions (`adk_greeter`, `agent_handoff`, `aj-greeter`, `worker_beta`) but populated for ones created just now (`ztest_17_tools_nested` → `2026-07-31 17:43:06`).

So the columns render but the data is missing or not being mapped. With 257 definitions on the cluster, "who created this and when" is the main way to tell test agents from real ones — right now there's no way to attribute an agent to a person.

## Steps to reproduce

1. Open **Agent Definitions** on a cluster with existing agents
2. Observe the **Created Time** and **Owner Email** columns

## Expected

- **Owner Email** shows the user or application that created the agent definition
- **Created Time** shows a timestamp for all definitions, not only recently created ones

## Actual

- **Owner Email**: blank on all rows
- **Created Time**: blank on older rows, populated on new ones

## Notes / likely cause

Agent definitions compile to Conductor workflow definitions, which carry `createTime` / `createdBy` (and `ownerEmail` on `WorkflowDef`). Two things worth checking:

1. Whether the UI is reading `ownerEmail` at all, or reading a field the agent-definition API doesn't return
2. Whether older definitions were created before those fields were persisted — if so, the column should show a placeholder rather than empty, and older rows need backfilling

Observed on the Orkes demo cluster (`orkes-demo.orkesconductor.io`), server 3.32.0-rc.10.

kowser-orkes:
Pull Request Title: fix: preserve createTime on WorkflowDef update in MetadataServiceImpl

Pull Request type
----

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

### Root cause:

- `MetadataServiceImpl.updateTaskDef()` preserves `createTime` on update
- `MetadataServiceImpl.updateWorkflowDef()` does not: it stamps `updateTime` only and passes the caller-supplied `createTime` directly to the DAO
- Agent definitions are persisted as `WorkflowDef`s — this asymmetry is the direct cause of #4311

### Fix:

- `MetadataServiceImpl.updateWorkflowDef()` fetches the existing `WorkflowDef` via `metadataDAO.getWorkflowDef(name, version)`
- Preserves the stored `createTime` when present, stamps a new value only when unset
- Mirrors the existing `updateTaskDef()` pattern
- Preserves the existing upsert semantics of `updateWorkflowDef` — does not throw when the row is missing, unlike `updateTaskDef`
- `createdBy` is intentionally left unchanged; a related gap was ruled working-as-intended by the maintainer in the closed conductor-oss/conductor#1450
- `AgentService.upsertWorkflowDef()`: removed the now-redundant manual `createTime` merge, superseded by the fix above

Tests added
----

- `MetadataServiceTest.testUpdateWorkflowDefPreservesCreateTime`
    - Registering a `WorkflowDef` stamps `createTime`
    - Updating it with a different caller-supplied value preserves the original
- `AgentSpanDeploymentContractEndToEndTest.deployStampsCreateTimeAndRedeployPreservesIt`
    - Deploying an agent stamps a non-zero `createTime`
    - Redeploying it preserves that value

Alternatives considered
----

- Push the guarantee into all five `MetadataDAO` backends (Redis, Postgres, MySQL, Sqlite, Cassandra)
    - Rejected: `MetadataServiceImpl` is the sole production caller of these DAO methods, so a single fix point is sufficient
    - Would have duplicated the same merge logic five times
- Fix only in `AgentService.upsertWorkflowDef()`
    - Rejected: does not cover other or future callers that bypass `AgentService`
- Merge the Postgres/MySQL `created_on` column into the hydrated object on read
    - Rejected: asymmetric, as Redis/Cassandra/Sqlite have no equivalent column
- Add a dedicated `MetadataDAO` method (e.g. `touchCreateTime`)
    - Rejected: no new capability is needed

### DIFF
--- a/core/src/main/java/com/netflix/conductor/service/MetadataServiceImpl.java
+++ b/core/src/main/java/com/netflix/conductor/service/MetadataServiceImpl.java
@@ -124,6 +124,12 @@ public class MetadataServiceImpl implements MetadataService {
      * @param workflowDef Workflow definition to be updated
      */
     public void updateWorkflowDef(WorkflowDef workflowDef) {
+        long createTime =
+                metadataDAO
+                        .getWorkflowDef(workflowDef.getName(), workflowDef.getVersion())
+                        .map(WorkflowDef::getCreateTime)
+                        .orElse(0L);
+        workflowDef.setCreateTime(createTime == 0 ? System.currentTimeMillis() : createTime);
         workflowDef.setUpdateTime(System.currentTimeMillis());
         metadataDAO.updateWorkflowDef(workflowDef);
         metadataChangeListener.onWorkflowDefUpdated(workflowDef);
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentService.java
@@ -886,19 +886,14 @@ public class AgentService {
 
     /**
      * Persist an agent-generated workflow with the metadata lifecycle used by normal workflow
-     * definitions. Direct DAO updates bypass {@link MetadataService} and leave a newly deployed
-     * agent's create time at zero, which prevents the definitions table from rendering it.
+     * definitions, routing to {@link MetadataService#registerWorkflowDef}/{@link
+     * MetadataService#updateWorkflowDef} so the create/update {@link
+     * org.conductoross.conductor.core.listener.MetadataChangeListener} events fire correctly.
+     * {@code createTime} itself no longer needs juggling here — {@link
+     * MetadataService#updateWorkflowDef} preserves it on update.
      */
     private void upsertWorkflowDef(WorkflowDef def) {
-        Optional<WorkflowDef> existing =
-                metadataDAO.getWorkflowDef(def.getName(), def.getVersion());
-        if (existing.isPresent()) {
-            // Older direct-DAO deployments have no create time; repair it on their next deploy.
-            if (existing.get().getCreateTime() == 0) {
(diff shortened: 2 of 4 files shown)


v1r3n (changes requested):
should be in the service layer not in the dao layer

v1r3n:
The changes should not be in the DAO layer.  There is metadata service interface that captures all the requirements to add ownerapp, timestamps etc, use that so its applicable to all the DAOs.

v1r3n:
Pushed candidate `fdc3ed66ec1453be12042e41ac0f4fec55dc536f` for review after bounded remediation. Local verification did not pass (`failed`); this is not a success claim. Review the verifier evidence before merging.

core/src/main/java/com/netflix/conductor/service/MetadataServiceImpl.java
```diff
@@ -124,6 +124,12 @@ public TaskDef getTaskDef(String taskType) {
      * @param workflowDef Workflow definition to be updated
      */
     public void updateWorkflowDef(WorkflowDef workflowDef) {
+        long createTime =
```
v1r3n, line 127: this does not make sense, why are we doing get?

v1r3n (changes requested):
see comments

v1r3n:
closing this, I will open another PR

PR #1480, opened 2026-08-05
fix(ai): a retried LLM chat task should keep its conversation history

ling-senpeng13:
## The bug

Retrying an `LLM_CHAT_COMPLETE` task in an agent workflow loses the entire conversation. The retried attempt sees only the static `[system, user]` template from the workflow definition, not the history the first attempt had.

## Why it happens

Conversation history is never expressed as a `${...}` reference in the workflow definition. `AgentChatCompleteTaskMapper` assembles it **imperatively in Java** at scheduling time — walking the workflow's completed tasks, extracting tool results, condensing if the context window demands it — and writes the result into the task's `inputData`.

Retry doesn't re-run the mapper. It copies the task and re-resolves the *definition*:

```java
TaskModel taskToBeRetried = task.copy();              // history is here...
taskToBeRetried.getInputData().putAll(taskInput);     // ...and clobbered by re-resolving
                                                      //    the definition's inputParameters
```

Since the definition's `inputParameters` contain only the static template, `putAll` overwrites the assembled conversation with it. The history existed solely in the first attempt's `inputData`, and that's what gets replaced.

## Fix
The mapper strips `messages` and `tools` from the task's own copy of the definition before it's persisted, so at retry time `getWorkflowTask().getInputParameters()` doesn't have them, the re-resolved `taskInput` in `taskToBeRescheduled()` doesn't have them, and `putAll` can't overwrite the assembled values which is the root cause of this bug.

`AIModelTaskMapper#detachAssembledInputFromDefinition(TaskModel, String... keys)`, shared by both chat-complete mappers.

## Testing
- groovy test
- It is covered in https://github.com/conductor-oss/python-sdk/pull/462. Verified in the local environment before the test PR is merged.

### DIFF
--- a/ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
+++ b/ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
@@ -12,6 +12,7 @@
  */
 package org.conductoross.conductor.ai.tasks.mapper;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
@@ -179,5 +180,44 @@ public abstract class AIModelTaskMapper<T extends LLMWorkerInput> implements Tas
                 return;
             }
         }
+    }
+
+    /**
+     * Preserves the chat input assembled during initial task mapping when the task is retried.
+     *
+     * <p>A retry copies the previous task input and then resolves the workflow task's input
+     * parameters again. That would replace the assembled conversation and tools with their original
+     * definitions. Removing those parameters from the task's private copy of the definition
+     * prevents that overwrite.
+     *
+     * <p>The workflow task is copied first because its original definition may be shared by later
+     * retries or other tasks in the workflow.
+     *
+     * <p>This method is best-effort: if the copy fails, the original definition is left intact and
+     * a retry will lose the assembled input.
+     */
+    protected void preserveAssembledChatInputOnRetry(TaskModel task, String... assembledKeys) {
+        WorkflowTask sharedDefinition = task.getWorkflowTask();
+        if (sharedDefinition == null || sharedDefinition.getInputParameters() == null) {
+            return;
+        }
+        WorkflowTask taskOwnedDefinition;
+        try {
+            taskOwnedDefinition = objectMapper.convertValue(sharedDefinition, WorkflowTask.class);
+        } catch (IllegalArgumentException e) {
+            log.warn(
+                    "Could not copy the task definition for {} — a retry of this task would fall back"
+                            + " to the static template",
+                    task.getTaskId(),
+                    e);
+            return;
+        }
+        Map<String, Object> inputParameters =
+                new HashMap<>(taskOwnedDefinition.getInputParameters());
+        for (String key : assembledKeys) {
+            inputParameters.remove(key);
+        }
+        taskOwnedDefinition.setInputParameters(inputParameters);
+        task.setWorkflowTask(taskOwnedDefinition);
     }
 }
--- a/test-harness/src/test/groovy/com/netflix/conductor/test/integration/LlmChatRetryContextSpec.groovy
+++ b/test-harness/src/test/groovy/com/netflix/conductor/test/integration/LlmChatRetryContextSpec.groovy
@@ -1,0 +1,243 @@
+/*
+ * Copyright 2026 Conductor Authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.netflix.conductor.test.integration
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.test.context.TestPropertySource
+
+import com.netflix.conductor.common.metadata.tasks.TaskDef
+import com.netflix.conductor.common.metadata.tasks.TaskResult
+import com.netflix.conductor.common.metadata.tasks.TaskType
+import com.netflix.conductor.common.metadata.workflow.WorkflowDef
+import com.netflix.conductor.common.metadata.workflow.WorkflowTask
+import com.netflix.conductor.core.dal.ExecutionDAOFacade
+import com.netflix.conductor.model.TaskModel
+import com.netflix.conductor.model.WorkflowModel
+import com.netflix.conductor.test.base.AbstractSpecification
+
+import groovy.json.JsonOutput
+
+/**
+ * <p>Conversation history is not carried by {@code ${...}} references. The LLM_CHAT_COMPLETE task
+ * mapper reconstructs it at scheduling time by walking the workflow's completed tasks. Retry is a
+ * different code path — it copies the task and re-resolves the carried definition's
+ * inputParameters over it (WorkflowExecutorOps#taskToBeRescheduled) — so it can hand the model a
+ * bare template with none of the tool exchanges that already happened. The model, seeing what
+ * looks like a new request, re-issues a tool call it has already made.
+ */
+@TestPropertySource(properties = ["conductor.integrations.ai.enabled=true"])
+class LlmChatRetryContextSpec extends AbstractSpecification {
+
+    private static final String CHAT_WORKFLOW = 'retry_llm_chat_context'
+    private static final String TOOL_WORKFLOW = 'retry_tool_input_control'
+
+    private static final String CUSTOMER_TOOL = 'lookup_customer'
+    private static final String WEATHER_TOOL = 'fetch_weather'
+    private static final String CHAT_REF = 'chat'
+
+    private static final String SYSTEM_MESSAGE = 'Answer using the customer lookup you already ran.'
+    private static final String USER_MESSAGE = "What's the weather where customer 4471 lives?"
+
+    /** Distinctive token from the tool result. Asserting on it proves the ACTUAL prior exchange
+     * was preserved — a merely tool-shaped message would say nothing about which call it describes. */
+    private static final String TOOL_RESULT_MARKER = 'Seattle'
+
+    @Autowired
+    ExecutionDAOFacade executionDAOFacade
+
+    def setup() {
+        registerTaskDef(TaskType.LLM_CHAT_COMPLETE.name())
+        registerTaskDef(CUSTOMER_TOOL)
+        registerTaskDef(WEATHER_TOOL)
+        metadataService.registerWorkflowDef(chatWorkflowDefinition())
+        metadataService.registerWorkflowDef(toolWorkflowDefinition())
+    }
+
+    def cleanup() {
+        unregister(CHAT_WORKFLOW)
+        unregister(TOOL_WORKFLOW)
+    }
+
+    def "a retried LLM chat task keeps its assembled conversation"() {
+        given: "a chat task scheduled after a tool call has already completed"
+        String workflowId = startWorkflow(CHAT_WORKFLOW, 1, 'retry-llm-chat-context', [:], null)
+        workflowTestUtil.pollAndCompleteTask(CUSTOMER_TOOL, 'worker',
+                [customer_id: '4471', name: 'Jane Doe', city: TOOL_RESULT_MARKER])
+
+        TaskModel firstAttempt = awaitChatAttempt(workflowId, 0)
+
+        and: "the fresh dispatch did receive the tool exchange — the behaviour a retry must match"
+        List<Map<String, Object>> assembled = messages(firstAttempt)
+        assert roles(assembled) != ['system', 'user']
+        assert hasToolExchange(assembled)
+
+        and: "the keys are dropped from the task's own copy, not from the shared definition"
+        WorkflowTask runtimeTemplate = workflowModel(workflowId)
+                .workflowDefinition
+                .getTaskByRefName(CHAT_REF)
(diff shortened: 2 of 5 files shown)


manan164 (changes requested):
Does this happen with all worker task that we lose input parameters on retry?

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
```diff
@@ -180,4 +181,61 @@ private void threadPreviousResponseId(TaskMapperContext context, TaskModel curre
...
+     * behaviour (a later retry loses the assembled input).
+     */
+    protected void detachAssembledInputFromDefinition(TaskModel task, String... assembledKeys) {
+        try {
+            WorkflowTask sharedDefinition = task.getWorkflowTask();
+            if (sharedDefinition == null || sharedDefinition.getInputParameters() == null) {
+                return;
+            }
+            boolean anyPresent = false;
+            for (String key : assembledKeys) {
```
NicholasDCole: It seems to me like we're placing fix in the wrong layer. 
This is a clean-up after all other aspects of task execution creation --- that means something is broken upstream

ling-senpeng13:
> Does this happen with all worker task that we lose input parameters on retry?

Yes — re-resolving `inputParameters` on retry is the designed behaviour for every worker task, not
something specific to LLM tasks:

```java
taskToBeRetried.getInputData().putAll(taskInput);   // taskToBeRescheduled
```

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
ling-senpeng13: > It seems to me like we're placing fix in the wrong layer. This is a clean-up after all other aspects of task execution creation --- that means something is broken upstream

I understand it is confusing.

The normal flow persists the assembled conversation. 

Retry doesn't re-run the mapper. `WorkflowExecutorOps.taskToBeRescheduled` It copies the task — assembled conversation and all — then re-resolves the definition's `inputParameters` and overlays the result:

```java
TaskModel taskToBeRetried = task.copy();              // history is here...
taskToBeRetried.getInputData().putAll(taskInput);     // ...and clobbered by re-resolving
                                                      //    the definition's inputParameters
```

`putAll` overwrites only the keys re-resolution produces, and those are exactly the keys the definition declares. messages is declared there — as the authored `[system, user]` template — so it overwrites the assembled conversation.

The fix removes `messages/tools` from the task's own copy of the definition. Re-resolution then produces no such keys, `putAll` has nothing to overwrite with, and the assembled values survive.

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
```diff
@@ -180,4 +181,61 @@ private void threadPreviousResponseId(TaskMapperContext context, TaskModel curre
...
+     * DO_WHILE iterations) always map from the definition's own instance, so they are unaffected.
+     *
+     * <p>Best-effort: a failure must not fail scheduling — the fallback is the pre-existing
+     * behaviour (a later retry loses the assembled input).
+     */
+    protected void detachAssembledInputFromDefinition(TaskModel task, String... assembledKeys) {
+        try {
+            WorkflowTask sharedDefinition = task.getWorkflowTask();
+            if (sharedDefinition == null || sharedDefinition.getInputParameters() == null) {
+                return;
```
NicholasDCole: Can we rename method and define the arguments statically?

```
  private static final Set<String> ASSEMBLED_CHAT_INPUT_KEYS =
          Set.of("messages", "tools");

  protected final void preserveAssembledChatInputOnRetry(TaskModel task) {
      // Existing deep-copy and removal logic using ASSEMBLED_CHAT_INPUT_KEYS.
  }
```
Something like this I feel is more clear

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
```diff
@@ -180,4 +181,61 @@ private void threadPreviousResponseId(TaskMapperContext context, TaskModel curre
...
+     * {@code $}{...} patterns must never be re-interpreted) and does not grow the task payload.
+     *
+     * <p>The deep copy is load-bearing: {@code task.getWorkflowTask()} is the shared instance from
+     * the cached {@link com.netflix.conductor.common.metadata.workflow.WorkflowDef}; mutating it
+     * would leak into every other execution using that definition. New schedules (including
+     * DO_WHILE iterations) always map from the definition's own instance, so they are unaffected.
+     *
+     * <p>Best-effort: a failure must not fail scheduling — the fallback is the pre-existing
+     * behaviour (a later retry loses the assembled input).
+     */
```
NicholasDCole, line 198:
```suggestion

  /**
   * Preserves the chat input assembled during initial task mapping when the task is retried.
   *
   * A retry copies the previous task input and then resolves the workflow task's input parameters
   * again. That would replace the assembled conversation and tools with their original definitions.
   * Removing those parameters from the task's private copy of the definition prevents that overwrite.
   *
   * The workflow task is copied first because its original definition may be shared by other
   * workflow executions.
   */
     
   
```

NicholasDCole: Suggesting this to be more readable and simple explanation of above

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
```diff
@@ -180,4 +181,61 @@ private void threadPreviousResponseId(TaskMapperContext context, TaskModel curre
...
+    protected void detachAssembledInputFromDefinition(TaskModel task, String... assembledKeys) {
+        try {
+            WorkflowTask sharedDefinition = task.getWorkflowTask();
```
NicholasDCole: Code is correct here. I might suggest little simpler version: 

```
      WorkflowTask definition = task.getWorkflowTask();
      if (definition == null || definition.getInputParameters() == null) {
          return;
      }

      Map<String, Object> definitionInput = definition.getInputParameters();
      if (ASSEMBLED_CHAT_INPUT_KEYS.stream().noneMatch(definitionInput::containsKey)) {
          return;
      }

      WorkflowTask taskDefinition =
              objectMapper.convertValue(definition, WorkflowTask.class);

      Map<String, Object> retryableInput =
              new HashMap<>(taskDefinition.getInputParameters());
      ASSEMBLED_CHAT_INPUT_KEYS.forEach(retryableInput::remove);

      taskDefinition.setInputParameters(retryableInput);
      task.setWorkflowTask(taskDefinition);
```

We don't need to catch exception here as well. That's already covered in the calling locations -- can rely on their error-handling I feel

NicholasDCole (changes requested):
Couple suggestions. Please check

manan164 (approved):

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
ling-senpeng13: I have removed the general exception handling would like to keep `IllegalArgumentException` from `objectMapper.convertValue()` so that we can properly fall back the default behavior.

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
ling-senpeng13: I applied your suggestion with some minor changes. PTAL

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
```diff
@@ -180,4 +181,53 @@ private void threadPreviousResponseId(TaskMapperContext context, TaskModel curre
...
+                anyPresent = true;
+                break;
+            }
+        }
+        if (!anyPresent) {
+            return;
+        }
+        WorkflowTask taskOwnedDefinition;
+        try {
+            taskOwnedDefinition = objectMapper.convertValue(sharedDefinition, WorkflowTask.class);
```
v1r3n, line 206: what is the reason for doing this?

v1r3n (changes requested):

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
ling-senpeng13: After thinking this a bit more, it is better for the caller to decide what keys to be removed. The list can change in the future.

NicholasDCole: It is additional complexity -- behavior is same from all calling sites. Should not be need to extend list or have different list based on calling site

ai/src/main/java/org/conductoross/conductor/ai/tasks/mapper/AIModelTaskMapper.java
NicholasDCole: I thought originally that we should do fix in a different layer, but that might require changes in core code since we are doing this : 
https://github.com/conductor-oss/conductor/blob/1d9524f297cbc8e394eda75ddb9299b3736f89ae/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java#L606-L613


I also don't see why the copy is necessary here. Can't it be as simple as below? @ling-senpeng13 

```
private static final Set<String> ASSEMBLED_CHAT_INPUT_KEYS =
          Set.of("messages", "tools");
    protected final void preserveAssembledChatInputOnRetry(TaskModel task) {
        WorkflowTask sharedWorkflowTask = task.getWorkflowTask();
        if (sharedWorkflowTask == null || sharedWorkflowTask.getInputParameters() == null) {
            return;
        }

        sharedWorkflowTask.getInputParameters().keySet().removeAll(ASSEMBLED_CHAT_INPUT_KEYS);
    }
```
What is purpose of copying over workflow task?

NicholasDCole: If we need to do copy, then can we do more simple method here? 
```
    protected final void preserveAssembledChatInputOnRetry(TaskModel task) {
        WorkflowTask sharedWorkflowTask = task.getWorkflowTask();
        if (sharedWorkflowTask == null || sharedWorkflowTask.getInputParameters() == null) {
            return;
        }

        Map<String, Object> inputParametersForRetry =
                new HashMap<>(sharedWorkflowTask.getInputParameters());

        if (!inputParametersForRetry.keySet().removeAll(ASSEMBLED_CHAT_INPUT_KEYS)) {
            return;
        }

        WorkflowTask taskLocalWorkflowTask = new WorkflowTask();
        BeanUtils.copyProperties(sharedWorkflowTask, taskLocalWorkflowTask, "inputParameters");
        taskLocalWorkflowTask.setInputParameters(inputParametersForRetry);

        task.setWorkflowTask(taskLocalWorkflowTask);
    }
```
It is currently far too verbose

NicholasDCole: It might be needed, but our tests don't currently prove that the workflow's runtime template are not modified
https://github.com/conductor-oss/conductor/blob/1d9524f297cbc8e394eda75ddb9299b3736f89ae/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java#L1374

test-harness/src/test/groovy/com/netflix/conductor/test/integration/LlmChatRetryContextSpec.groovy
```diff
@@ -0,0 +1,226 @@
...
+        workflowTestUtil.pollAndCompleteTask(CUSTOMER_TOOL, 'worker',
+                [customer_id: '4471', name: 'Jane Doe', city: TOOL_RESULT_MARKER])
+
+        TaskModel firstAttempt = awaitChatAttempt(workflowId, 0)
+
+        and: "the fresh dispatch did receive the tool exchange — the behaviour a retry must match"
+        List<Map<String, Object>> assembled = messages(firstAttempt)
+        assert roles(assembled) != ['system', 'user']
+        assert hasToolExchange(assembled)
+
```
NicholasDCole, line 83: Can we add a case for this?        

```
 and: "the task-local retry snapshot did not modify the workflow's runtime template"
        WorkflowTask runtimeTemplate = workflowModel(workflowId)
                .workflowDefinition
                .getTaskByRefName(CHAT_REF)
        assert runtimeTemplate.inputParameters.messages*.message == [SYSTEM_MESSAGE, USER_MESSAGE]
        assert runtimeTemplate.inputParameters.tools*.name == [WEATHER_TOOL]
        assert !firstAttempt.workflowTask.inputParameters.containsKey('messages')
        assert !firstAttempt.workflowTask.inputParameters.containsKey('tools')
        assert firstAttempt.inputData.tools*.name == [WEATHER_TOOL]
```

v1r3n (changes requested):
does not look right

ling-senpeng13:
> does not look right

Normal scheduling calls the task mapper, which assembles the conversation — including the tool
call and its response — into the task's inputData, and that is what gets persisted. The task
definition's inputParameters still holds the authored versions: messages as the [system, user] template, and tools as declared.

Retry does not call the task mapper. It copies the persisted task, re-resolves inputParameters
from the static definition, and putAlls the result over the copy — so the authored values
overwrite the persisted messages and tools.

The retried attempt then reaches the model with no record of the tool it already called, so the
model calls it again — a duplicate execution of a possibly side-effecting tool.

The summary: Retry rebuilds a task's input from its definition rather than reusing it — correct only while the input is derivable from the definition. Any mapper that assembles input in Java breaks that assumption, so it must remove those keys from its task's copy of the definition, leaving re-resolution nothing to overwrite them with.

v1r3n (changes requested):
tbd

bradyyie:
Will solve this later speaking viren as we can maybe roll this up into getHistory instead and have it propgate that way

Issue #1303, opened 2026-07-16
[UI] Stale "AgentSpan" branding in the Create Agent dialog


deeptireddy-lab:
**Found by:** Deepti
**Area:** Agents UI / branding

**Description**
The **Create Agent** dialog still references "AgentSpan" throughout, which is no longer the name to surface.

**Steps to reproduce**
1. Open the Conductor UI
2. Go to the **Agents** section
3. Click **Create Agent**

**Actual** — the dialog shows stale `AgentSpan` branding:
- Title: "BUILD AN AGENT WITH THE **AGENTSPAN** SDK"
- Body: "…they're built with the **AgentSpan** SDK or CLI…"
- Button: "View **AgentSpan** on GitHub"

**Expected**
No `AgentSpan` references — use the current naming (Conductor / Agents).

**Environment**
OSS getting-started UI. Screenshot available (attach in-thread).

_Related: empty Agents-tab branding issue (filed separately)._

kunalrawat425:
Working on removing the stale AgentSpan references from the Create Agent dialog. Will update branding to use current Conductor naming. PR incoming.

PR #1481, opened 2026-08-05
fix(ui): suppress HANDOFF badge on leaf agents; drop stale AgentSpan branding

### LINKED ISSUE #1453: UI: leaf agents show a false HANDOFF strategy badge (SDK default leaking into the diagram)
## Problem

The Agent Definition diagram shows a **HANDOFF** strategy badge and a turn count on leaf agents that have no strategy and no sub-agents. It reads as configured behaviour when it is just an unset default.

## Repro

In `ztest_15_three_level` (orkes-demo), `ztest_15a` and `ztest_15e` are plain single agents:

```python
A("15a", "Classify the incident in one line.")   # no strategy, no sub-agents
A("15e", "Write the postmortem summary.")        # no strategy, no sub-agents
```

The diagram renders both as:

```
ztest_15a   AGENT   [HANDOFF]  25 turns
ztest_15e   AGENT   [HANDOFF]  25 turns
```

## Cause

`Agent.__init__` defaults `strategy` to `Strategy.HANDOFF` and `max_turns` to `25` regardless of whether sub-agents exist:

```python
>>> a = Agent(name='leaf_probe', model='openai/gpt-4o-mini', instructions='x')
>>> a.strategy    -> Strategy.HANDOFF
>>> a.agents      -> []
>>> a.max_turns   -> 25
```

The serializer is already aware this is meaningless for leaves — `config_serializer.py` only emits `strategy` on the wire when the agent has sub-agents:

```python
has_sub_agents = (bool(agent.agents)
                  or getattr(agent, "planner", None) is not None
                  or getattr(agent, "fallback", None) is not None)
config = {..., "strategy": agent.strategy if has_sub_agents else None, ...}
```

So the UI is reading the SDK object default rather than the serialized definition, or falling back to a default when `strategy` is null.

## Why it matters

"HANDOFF" on a node with zero sub-agents states something false — handoff to what? Anyone reading the diagram to understand a multi-agent design will misread which nodes actually coordinate. In `ztest_15_three_level` it makes two of the three visible nodes look like coordinators when only `ztest_15rou` is one.

## Expected

Show no strategy badge when `agents` is empty and no `planner`/`fallback` is set — mirroring the `has_sub_agents` logic the serializer already uses. Same for the turn count: `25 turns` on a single-shot leaf agent is noise.

Server 3.32.0-rc.10, orkes-demo cluster.

shaileshpadave:
Fixes #1453, #1303, #1304.

**What changed:**
- Leaf agents in the diagram no longer show a false HANDOFF badge and "25 turns" — those only appear now when an agent actually has sub-agents
- Replaced stale "AgentSpan" references in the Agents and Skills empty states with "Conductor Agents"

**Tests:** 3 new unit tests covering the leaf/coordinator badge logic. All existing agent page tests pass.


shaileshpadave:
Before:
image

After:
image

Before:
image

After:
image

v1r3n (approved):

PR #1482, opened 2026-08-05
fix(ui): correct Run Agent navigation and Go Back from error page

### LINKED ISSUE #1317: [3.32.0-rc.9] Clicking on ‘Run Agent’ in Agent Executions gives a 404 and then clicking Go Back goes to Workflow executions
**Details**
Conductor version: 3.32.0-rc.9

**To Reproduce**
1. Go to the Agent Executions page
2. Click on the Run Agent button. It will show a 404

Image

3. Click `GO BACK` and it will go to the Workflow Executions page instead of Agent Executions

Image

shaileshpadave:
Fixes #1317.

## What changed

**`AgentSearch.tsx` — Run Agent button gives a 404**

Replaced `usePushHistory` with `useNavigate` directly for the "Run Agent" button, matching the pattern already used in `AgentDefinitions.tsx` for the same route. The `usePushHistory` wrapper (via `url-parse`) was unnecessary here — it's intended for multi-cluster stack-switching — and was unreliable when navigating to `/runAgent` from the executions page.

**`Error.tsx` — Go Back lands on Workflow Executions instead of Agent Executions**

`navigate.length` is the function's parameter count (always `1` in React Router v6 because `options` defaults to `{}`), not browser history length. This made the condition `navigate.length <= 1` permanently `true`, so every "Go Back" click on any error page called `navigate("/")` (Workflow Executions) instead of `navigate(-1)`.

Changed to `window.history.length`, which is the correct check.

## Test plan

- [ ] Agent Executions → click **Run Agent** → lands on Run Agent page (no 404)
- [ ] Run Agent → click **Close** → returns to Agent Executions
- [ ] Navigate to a non-existent URL → click **GO BACK** → returns to the previous page, not Workflow Executions

### DIFF
--- a/ui-next/src/components/ui/Error.test.tsx
+++ b/ui-next/src/components/ui/Error.test.tsx
@@ -1,0 +1,42 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { ThemeProvider } from "@mui/material/styles";
+import { MemoryRouter, Route, Routes } from "react-router";
+import { describe, it, expect, vi } from "vitest";
+import theme from "theme/theme";
+import Error from "./Error";
+
+const mockNavigate = vi.fn();
+vi.mock("react-router", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("react-router")>();
+  return { ...actual, useNavigate: () => mockNavigate };
+});
+
+function renderAtKey(locationKey: string) {
+  return render(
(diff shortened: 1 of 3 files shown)


shaileshpadave:
Closing — the Run Agent 404 fix is hard to validate end-to-end in browser since the underlying navigation bug is also fixed in this PR. Will revisit.

PR #1485, opened 2026-08-05
fix(ui): model picker on Run Agent form (#1451)

### LINKED ISSUE #1451: UI: Model override on Run Agent is free-text with no way to pick an available model
## Problem

On the **Run Agent** page, **Model override** is a free-text input with placeholder *"Use the deployed agent model"*. There is no way to discover or pick a model — you have to already know the exact provider/model string and type it correctly.

Every other field on the form is constrained (**Agent** is a dropdown), so this is the one input where a typo silently changes behavior.

## Steps to reproduce

1. Open **Run Agent**
2. Select any agent
3. Click **Model override**

## Expected

A picker listing the models actually available on this cluster — sourced from the configured AI/LLM integrations and the Model Library — so you can select rather than guess. Free text as a fallback is fine, but the selectable list should be the default path.

## Actual

An empty text box. No dropdown, no autocomplete, no link to the Model Library, no indication of the expected format (`provider/model`, e.g. `openai/gpt-4o-mini`).

## Why it matters

1. **Undiscoverable** — nothing on the form tells you the format or which models the cluster has. The user has to leave the page to find out.
2. **Silent failure mode** — a typo or a model not configured on this cluster fails at execution time with a server-side model-resolution error, which reads as an agent failure rather than a bad input.
3. **It defeats the feature's purpose** — the field exists to let you A/B a different model per execution. If you have to already know the exact string, the people most likely to use it are the least likely to reach for it.

## Suggested fix

Populate a dropdown from the cluster's configured integrations (the same source the Model Library uses), grouped by provider, with the deployed agent model shown as the default. Keep free text as an "other" option for models not yet in the library.

Observed on the Orkes demo cluster (`orkes-demo.orkesconductor.io`), server 3.32.0-rc.10.

shaileshpadave:
Fixes #1451

The Model override field was a plain text box with no hint of what to type. Now it fetches the AI model integrations configured on the cluster and offers them as autocomplete suggestions in `provider/model` format (e.g. `openai/gpt-4o`).

Free text entry still works for models not in the list. If no AI integrations are configured, it degrades gracefully to an empty autocomplete (same UX as before).

### DIFF
--- a/ui-next/src/pages/agent/RunAgent.tsx
+++ b/ui-next/src/pages/agent/RunAgent.tsx
@@ -7,13 +7,15 @@ import SectionHeader from "components/layout/SectionHeader";
 import { ConductorAutoComplete } from "components/ui/inputs/ConductorAutoComplete";
 import ConductorInput from "components/ui/inputs/ConductorInput";
 import SectionContainer from "components/ui/layout/SectionContainer";
-import { useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { Helmet } from "react-helmet";
 import { useLocation, useNavigate } from "react-router";
 import { AGENT_EXECUTIONS_URL } from "utils/constants/route";
-import { useAction, useFetch } from "utils/query";
+import { useAction, useAuthHeaders, useFetch } from "utils/query";
 import { useLocalStorage } from "utils";
 import { v4 as uuidv4 } from "uuid";
+import { fetchWithContext, useFetchContext } from "plugins/fetch";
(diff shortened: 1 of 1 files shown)


nthmost-orkes:
Superseded by #1519, which merged on 2026-08-10 and adds the model picker to the Run Agent form, resolving the same issue (#1451, now closed). The fix is already on main.

If I've closed this in error please respond — this is a quick cleanup sweep and I'm happy to reopen. Thanks for the contribution!

PR #1498, opened 2026-08-06
fix: write and compare sqlite index timestamps in UTC (#1497)

### LINKED ISSUE #1497: SQLite index stores start_time in local time but searches it as UTC — time-range filters hide recent executions
## Summary

The SQLite index writes `start_time` in the JVM local timezone. Search converts the query value to UTC. The two do not match. Time-range searches drop recent records.

## Impact

- The UI "Last N hours" presets return Empty.
- `conductor agent execution --since 1h` returns nothing.
- `/api/workflow/search?query=startTime><ms>` under-reports.

The error equals the JVM UTC offset. E.g.: my machine runs `America/Asuncion` (-04:00), so records stay hidden for 4 hours. UTC machines do not show the bug.

## Steps to reproduce

1. Set the machine timezone to a non-UTC zone.
2. Start the server with the default SQLite persistence (`3.32.0-rc.23`).
3. Start a workflow.
4. Open the UI and select "Last 1 hour". The list is empty.

## Cause

Write path — `SqliteIndexDAO.java:98`:

```java
Timestamp startTime = Timestamp.from(Instant.from(ta));
```

`java.sql.Timestamp` serializes with the JVM default zone. The driver stores local time.

Read path — `SqliteIndexQueryBuilder.millisToUtc:153`:

```java
ZonedDateTime startDate = ZonedDateTime.ofInstant(Instant.ofEpochMilli(startTimeMilli), ZoneOffset.UTC);
```

The query value is UTC.

mp-orkes:
Fixes #1497 - take a look at the screen recording for a repro of the issue.

## What changed

`SqliteIndexDAO` wrote `start_time`/`update_time` via `java.sql.Timestamp.toString()`, which renders in the **JVM default zone**, while `SqliteIndexQueryBuilder` rendered the search bound in **UTC**. These columns are TEXT-compared, so every time-range search was wrong by the host's UTC offset.

Both sides now use one canonical format, `yyyy-MM-dd HH:mm:ss.SSS` in UTC, applied to all four index timestamp columns (`workflow_index` and `task_index` × `start_time`/`update_time`).

`V6__index_timestamps_to_utc.sql` converts existing rows.

## How to test

```
./gradlew :conductor-sqlite-persistence:test
```

The suite runs under `TZ=America/Los_Angeles` (set in `sqlite-persistence/build.gradle`). This is deliberate and load-bearing — CI runs UTC, where a local-time/UTC mismatch is invisible and every one of these tests passes vacuously.

To see it red, check out the first commit alone (tests without the fix): 15 failures, including the reported symptom as `expected:<1> but was:<0>`..

### DIFF
--- a/sqlite-persistence/build.gradle
+++ b/sqlite-persistence/build.gradle
@@ -35,4 +35,8 @@ dependencies {
 test {
     //the MySQL unit tests must run within the same JVM to share the same embedded DB
     maxParallelForks = 1
+    // Issue #1497 regression tests need a non-UTC zone to be meaningful; CI runs UTC, where a
+    // local-time/UTC mismatch is invisible. TZ reaches both the forked JVM's default zone and
+    // the native SQLite library's localtime, so they agree.
+    environment 'TZ', 'America/Asuncion'
 }
--- a/sqlite-persistence/build.gradle
+++ b/sqlite-persistence/build.gradle
@@ -38,5 +38,5 @@ test {
     // Issue #1497 regression tests need a non-UTC zone to be meaningful; CI runs UTC, where a
     // local-time/UTC mismatch is invisible. TZ reaches both the forked JVM's default zone and
     // the native SQLite library's localtime, so they agree.
-    environment 'TZ', 'America/Asuncion'
+    environment 'TZ', 'America/Los_Angeles'
 }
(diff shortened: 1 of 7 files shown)


mp-orkes:
This is happening with SQLite. 
On this screen recording you can see that I run a workflow, then search within last 15min, last 30min... and get nothing until last 12h. Which is wrong.  


https://github.com/user-attachments/assets/42966099-ae7e-4e1a-990d-371cf4ebe322

v1r3n (approved):

PR #1516, opened 2026-08-10
fix(ui-next): sort Agent Name on the column that is actually indexed

### LINKED ISSUE #1514: Sorting the agent name in ascending and descending order does not work reliably - this is on the agent executions page

manan164:
Fixes #1514

`handleSort` was renaming the column id `workflowType` to `workflowName` before sending it. `workflowName` isn't a real field on any backend — not in the sqlite/postgres `VALID_FIELDS`, not in the ES mappings — so the sort was dropped and rows came back in whatever order the query plan yielded. On Elasticsearch it's worse: the unmapped field is rejected outright and the search returns **HTTP 500**, so the page fails to load rather than mis-ordering. **This PR deletes that rename;** the column id now goes through unchanged. The same bad remap existed on the workflow executions and task search pages, so all five call sites now share one helper.

### Before — clicking the name column sends `workflowName:DESC` → 500, table empty
before

### After — sends `workflowType:DESC` → 200, rows sorted
after

Same URL, same filters, same server; only the `handleSort` line differs. (Captured on the Workflow Executions page, which shares the identical remap — the environment had no agent executions to show.)

Frontend-only, no backend behaviour change. Tests added at both layers (`buildSortParam`, sqlite + postgres query builders), each confirmed failing before the fix.

Not addressed here: `workflow_type` is non-unique and no tiebreaker is appended, so rows sharing an agent name can still shuffle across page boundaries under offset paging.

### DIFF
--- a/postgres-persistence/src/test/java/com/netflix/conductor/postgres/util/PostgresIndexQueryBuilderTest.java
+++ b/postgres-persistence/src/test/java/com/netflix/conductor/postgres/util/PostgresIndexQueryBuilderTest.java
@@ -725,4 +725,26 @@ public class PostgresIndexQueryBuilderTest {
         inOrder.verify(mockQuery).addParameter(0);
         verifyNoMoreInteractions(mockQuery);
     }
+
+    @Test
+    void shouldSortOnWorkflowTypeWhichBacksTheAgentNameColumn() throws SQLException {
+        PostgresIndexQueryBuilder builder =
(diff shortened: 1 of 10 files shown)


manan164:
Closing this since https://github.com/conductor-oss/conductor/pull/1515 takes care of this.

PR #1517, opened 2026-08-10
fix(persistence): index end_time so End Time can be sorted and filtered

manan164:
End Time is sortable and has a date-range filter, but `end_time` was never a column on `workflow_index`/`task_index`, so the builders dropped it from both the `ORDER BY` and the `WHERE`: the sort returned unordered rows and the filter was silently ignored. Adds the column, index, write path and back-fill on sqlite + postgres. Fixes task search too, since it shares the same allow-list.

`NOT NULL DEFAULT epoch` rather than nullable: SQLite sorts NULL lowest and Postgres highest, so running executions would otherwise land at opposite ends of `endTime:DESC` depending on backend.

**Existing executions are back-filled** from `json_data` — terminal executions are never re-indexed, so otherwise they'd sort as unfinished forever. Both back-fills survive a malformed value rather than aborting. Postgres needs more care than V13.2 took: `to_timestamp(..., 'YYYY-MM-DD"T"HH24:MI:SS.MS')` raises `invalid value "Z" for "MS"` on an `endTime` with no milliseconds, and `applyDataMigrations` defaults to `true`, so that would abort the migration and stop the server booting. V18.1 uses a `::timestamptz` cast (accepts any ISO-8601 form, reads trailing `Z` as UTC) behind a regex guard, so a non-timestamp value skips the row instead of failing the statement.

**Before merging:** V18.1 is a single unbatched `UPDATE` over the whole index table, and the `CREATE INDEX` is not `CONCURRENTLY` so it blocks writes while building. Both are worth timing against a production-sized copy; `applyDataMigrations=false` defers the back-fill if you'd rather run it out of band. (V14 and V17 create indexes the same way, so that part isn't new here.)

Tests assert actual returned row order and actual filtering through `searchWorkflowSummary`, not just the generated SQL; both fail correctly when `end_time` is removed from `VALID_FIELDS`. Back-fills verified against a 3913-workflow sqlite DB (0 mismatches, index used for the sort) and against seeded postgres covering values with/without milliseconds, with offsets, and empty/malformed/absent.
endtime-WITHOUT-fix

endtime-WITH-fix


bradyyie:
https://github.com/conductor-oss/conductor/issues/1514

bradyyie (approved):
Tested locally and approved for OSS - working and filtering on Postgres

PR #1520, opened 2026-08-10
fix(agentspan): JOIN only compacts outputs containing state keys

### LINKED ISSUE #1499: On conductor-oss 3.32.0-rc.23/24, the JOIN inside an agent's DO_WHILE loop returns empty output even though its forked tool tasks all COMPLETED
On conductor-oss 3.32.0-rc.23/24, the JOIN inside an agent's DO_WHILE loop returns empty output even though its forked tool tasks all COMPLETED, so tool results never re-enter the LLM context and the agent re-requests the same tools every turn until max_turns. Confirmed on execution 4173ff62-0ccb-4dae-ac8e-f4c01fee6384 (04_http_and_mcp_tools.py)
manually recreated on latest on oss.

- https://github.com/orkes-io/orkes-conductor/issues/3838 — the original bug report seems very similar

Image

If you run 04_http_and_mcp_tools.py, you will see the issue. It happens on many other examples still in oss.

NicholasDCole:
Closes https://github.com/conductor-oss/conductor/issues/1499

Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

- **JOIN no longer drops unmarked tool outputs.**
  - Preserve full outputs from fork branches without `state`/`_state_updates` (e.g. MCP, HTTP tool results).
  - Continue compacting state-bearing branches to those two merge keys.
  - Add unit and e2e coverage proving an unmarked MCP result survives JOIN and state merge.
- **Swarm agents can respond after their own tool calls.**
  - Only suppress a sub-agent's post-tool-call turn when a parent `on_tool_result` handoff watches a tool *that agent* owns, not whenever any handoff targets any tool — previously this could end an agent's turn before its LLM saw its tool's result.
  - Add a `ctx_inject` task to the resumed per-turn loop so it sees prior state/tool-results instead of the bare prompt, matching existing single-agent loop wiring.
- **UI fixes for agent execution diagrams.**
  - Fix `AgentExecutionDiagram` rendering the literal string "undefined" as the response sublabel for tool-call-only turns.
  - Render handoffs (explicit transfer, condition-based, or HANDOFF-strategy routing) between real sub-agent turns, which previously showed nothing.
#### Before

MCP Tool Output are lost on join --- MCP tools may/may not be called repeatedly . Not deterministic and recovery is possible by LLM: <br>

image

#### After
MCP Tool output are not lost on Join --- MCP tools not called repeatedly: <br>

image

#### Before:

Billing Swarm Agent is not able to do a transfer after a tool result comes. LLM task is not possible after tool result. (this is because we defined AfterToolResult at all in the example agent). <br>
image

Context is not present: <br>

image

#### After:

Billing Swarm Agent is able to do a transfer explicitly : <br>
image

Context is present: <br>
image

#### Before:

Handoffs do not render: <br>
image

#### After:

Handoffs are showing:  <br>
image


NicholasDCole:
Validated all above scenarios using this python example of a swarm using AfterToolResult: 

```
"""Swarm handoff + shared state — regression repro for a JOIN/state-compaction bug.

SWARM's OnToolResult handoff (17/64_swarm*.py) matches against a tool's
result on the fork/join for the turn. A tool that writes to context.state
(51_shared_state.py) gets that folded into its own output as
result["_state_updates"] (see conductor.ai.agents.runtime._dispatch).

Server bug (fixed in orkes-conductor PR #3925): when a tool result carried
both a return value and _state_updates, JOIN's state compaction kept
_state_updates/state but dropped the tool's name and result. OnToolResult
never matched, the handoff never fired, and `support` looped on
lookup_customer until max_turns.

This also covers two follow-on fixes via a three-agent chain
(support -> account_specialist -> billing_specialist):

    - returnAfterToolResults is scoped per-agent to whether *that agent's*
      tool is being watched by a handoff, not to "a handoff exists anywhere
      in the swarm" (the old bug). support and account_specialist's tools
      are each watched, so their post-tool turn is correctly skipped.
    - billing_specialist's tool (get_account_balance) is deliberately NOT
      watched by any handoff. It needs a real second turn to answer, and
      that turn should still see the name/plan gathered earlier — proving
      ctx_inject reaches the swarm sub-agent loop, not just single-agent.

Runs with the default full transfer mesh (no allowed_transitions): every
agent can transfer to every other agent, so the declared handoffs are
racing against the model just choosing to transfer instead.

Requirements:
    - Conductor server with LLM support
    - CONDUCTOR_SERVER_URL=http://localhost:8080/api in .env or environment
    - CONDUCTOR_AGENT_LLM_MODEL=openai/gpt-4o-mini in .env or environment
"""

from conductor.ai.agents import Agent, AgentRuntime, Strategy, tool
from conductor.ai.agents.handoff import OnToolResult
from conductor.ai.agents.tool import ToolContext
from settings import settings


# Writing to context.state is what triggers the _state_updates wrapping.

@tool
def lookup_customer(customer_id: str, context: ToolContext = None) -> dict:
    """Look up a customer's name by ID."""
    name = "Alice" if customer_id == "c-42" else "Unknown"
    context.state["customer_name"] = name
    return {"customer_id": customer_id, "customer_name": name}


@tool
def get_customer_plan(context: ToolContext = None) -> dict:
    """Retrieve the customer's subscription plan, from shared state."""
    context.state["plan"] = "Premium"
    return {
        "customer_name": context.state.get("customer_name", "not found"),
        "plan": "Premium",
    }


@tool
def get_account_balance(context: ToolContext = None) -> dict:
    """Retrieve the customer's account balance, from shared state."""
    context.state["balance"] = "$1,204.50"
    return {
        "customer_name": context.state.get("customer_name", "not found"),
        "plan": context.state.get("plan", "not found"),
        "balance": "$1,204.50",
    }


billing_specialist = Agent(
    name="billing_specialist",
    model=settings.llm_model,
    instructions=(
        # No handoff watches get_account_balance — the control case, so this
        # agent must reach a real second turn to answer.
        "You are a billing specialist. Call get_account_balance to retrieve "
        "the customer's balance, then summarize the full record — name, "
        "plan, and balance — for the customer in one or two sentences."
    ),
    tools=[get_account_balance],
    max_turns=3,
)

account_specialist = Agent(
    name="account_specialist",
    model=settings.llm_model,
    instructions=(
        # Not saying "then transfer to billing_specialist" on purpose — the
        # compiler pastes these instructions into the transfer tool
        # description peers see, and it already handed off via OnToolResult.
        "You are an account specialist. Call get_customer_plan to retrieve "
        "the customer's plan."
    ),
    tools=[get_customer_plan],
    max_turns=3,
)

support = Agent(
    name="support",
    model=settings.llm_model,
    instructions=(
        # Kept generic rather than echoing the user's "look up the
        # customer" wording — that overlap made peers' transfer tools look
        # more relevant than their own tools, so models transferred instead
        # of calling them.
        "You are front-line customer support. If the customer's plan and "
        "balance are already in context, the lookup is done — do not call "
        "lookup_customer again. Otherwise, when given a customer ID, call "
        "lookup_customer with it."
    ),
    tools=[lookup_customer],
    agents=[account_specialist, billing_specialist],
    strategy=Strategy.SWARM,
    handoffs=[
        OnToolResult(
            tool_name="lookup_customer",
            result_contains="Alice",
            target="account_specialist",
        ),
        # result_contains is documented as optional, but the server
        # validator currently rejects on_tool_result without it.
        OnToolResult(
            tool_name="get_customer_plan",
            result_contains="Premium",
            target="billing_specialist",
        ),
    ],
    # No allowed_transitions — full transfer mesh, SWARM's default.
    max_turns=10,
)


if __name__ == "__main__":
    with AgentRuntime() as runtime:
        print("=" * 60)
        print("  Swarm handoff chain gated on state-mutating tool results")
        print("=" * 60)
        result = runtime.run(support, "Can you look up customer c-42 for me?")
        result.print_result()

        output = str(result.output)
        if "Alice" in output and "1,204.50" in output:
            print(
                "[OK] billing_specialist received the full chain: "
                "name -> plan -> balance"
            )
        elif "Alice" in output:
            print(
                "[PARTIAL] handoff reached account_specialist but the chain "
                "stopped before billing_specialist saw the balance."
            )
        else:
            print(
                "[FAIL] handoff never fired — support likely re-ran "
                "lookup_customer until max_turns. This is the "
                "JOIN/state-compaction bug."
            )

        # Production pattern:
        # 1. Deploy once during CI/CD:
        # runtime.deploy(support)
        # CLI alternative:
        # runtime.deploy(agent) from a release script
        #
        # 2. In a separate long-lived worker process:
        # runtime.serve(support)

```

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/MultiAgentCompiler.java
```diff
@@ -1685,6 +1685,37 @@ private WorkflowDef compileSwarmAgentWorkflow(
...
+                        "state", "${workflow.variables._agent_state}",
+                        "signals", "${workflow.variables._signal_injection}",
+                        "toolResults", "${workflow.variables._last_tool_results}",
+                        "maxSize", 32768,
+                        "maxValueSize", 4096,
+                        "expression", JavaScriptBuilder.contextInjectionScript()));
+        @SuppressWarnings("unchecked")
+        List<Object> llmMessages = (List<Object>) llmTask.getInputParameters().get("messages");
+        for (int mi = 0; mi < llmMessages.size(); mi++) {
+            if (llmMessages.get(mi) instanceof Map<?, ?> msg && "user".equals(msg.get("role"))) {
```
kowser-orkes: NIT: Lets not use index like this, prefer `for(entry : list)` syntax and return the index. And later use the index to mutate the state

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/MultiAgentCompiler.java
```diff
@@ -2163,15 +2194,29 @@ private List<WorkflowTask> buildSwarmCaseTasks(
...
                 parent.getHandoffs() != null
                         && parent.getHandoffs().stream()
                                 .anyMatch(
                                         handoff ->
                                                 handoff != null
                                                         && "on_tool_result"
-                                                                .equals(handoff.getType()));
+                                                                .equals(handoff.getType())
+                                                        && subToolNames.contains(
+                                                                handoff.getToolName()));
```
kowser-orkes: My brain hurts when trying to read this part

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -84,19 +84,15 @@ public boolean execute(
 
             TaskModel.Status taskStatus = forkedTask.getStatus();
 
-            // Only add to task output if it's not empty. For Conductor-Agents agent executions,
-            // copy
-            // only the agent merge keys (compact) to keep the JOIN payload small; otherwise copy
-            // the full fork output (default Conductor behavior).
+            // Only add to task output if it's not empty. For Conductor-Agents, preserve marked
+            // tool outputs with their identity, compact state-only branches, and pass through
+            // unmarked tool outputs for the ReAct state merge.
```
kowser-orkes: Lets clean comments like these

core/src/main/java/com/netflix/conductor/core/execution/tasks/Join.java
```diff
@@ -153,21 +149,31 @@ private static boolean isAgentExecution(WorkflowModel workflow) {
...
         Map<String, Object> output = forkedTask.getOutputData();
-        Map<String, Object> compact = new LinkedHashMap<>();
+
+        // Tool identity takes precedence because a tool output may also contain state updates.
         Object agentToolName = forkedTask.getInputData().get("_agent_tool_name");
         if (agentToolName != null) {
-            compact.put("_agent_tool_name", agentToolName);
-            compact.put("_agent_tool_output", output);
-            return compact;
+            Map<String, Object> toolOutput = new LinkedHashMap<>();
```
kowser-orkes, line 178: Can we return immutable maps?

agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/MultiAgentCompiler.java
```diff
@@ -1685,6 +1685,37 @@ private WorkflowDef compileSwarmAgentWorkflow(
...
+        // Context injection: without this, a resumed turn only ever sees the original prompt with
+        // no record of what happened on prior turns, so the LLM has no way to know it already
+        // made a tool call and just repeats it until max_turns instead of reading the result and
+        // answering. Mirrors the ctx_inject wiring in AgentCompiler's own loop-building.
+        String ctxInjectRef = agent.getName() + "_ctx_inject";
+        WorkflowTask ctxInject = new WorkflowTask();
+        ctxInject.setType("INLINE");
+        ctxInject.setTaskReferenceName(ctxInjectRef);
+        ctxInject.setInputParameters(
+                Map.of(
```
kowser-orkes: seems like a constant

kowser-orkes (dismissed):

bradyyie (approved):

PR #1523, opened 2026-08-11
fix(retry): preserve LLM chat conversation across task retry

manan164:
## The bug
An `LLM_CHAT_COMPLETE` task's `messages` (conversation history) and `tools` are assembled imperatively by its `TaskMapper` from the workflow's **completed tasks** — they are not derivable from the task definition's `inputParameters`.

Both retry paths `copy()` the failed task and then re-resolve the definition's `inputParameters` over it with a blind merge, **skipping the mapper**:
- `WorkflowExecutorOps.taskToBeRescheduled` — manual retry / terminate+retry
- `DeciderService.retry` — automatic retry within `retryCount`

So the retried chat is dispatched with only the bare `[system, user]` template — none of the tool/assistant turns that already happened — and the model re-issues a tool call it already made.

## The fix (generic — no task type or input key special-cased)
On retry, **re-map the task through its `TaskMapper`** instead of re-resolving the definition alone. The mapper rebuilds the input from current workflow state (regenerating the conversation via `getHistory`) exactly as first-scheduling does, and still re-resolves `${...}` (e.g. `CPEWF_TASK_ID`), preserving existing retry semantics.

- Extract a filter-free `DeciderService.mapTask(...)` from `getTasksToBeScheduled` (passing `taskDefinition` + the generated `taskId` into `getTaskInput`, matching orkes-conductor#3641); add `remapRetriedTaskInput(...)`.
- `DeciderService.retry` and `WorkflowExecutorOps.taskToBeRescheduled` apply the re-mapped input, falling back to the previous definition re-resolution if the re-map yields nothing for the task.

## Verified
- Chat conversation preserved on **manual** retry (terminate→retry) ✓
- Chat conversation preserved on **automatic** retry (`retryCount` transient fail) ✓
- Non-chat regression: a SIMPLE task's `${CPEWF_TASK_ID}` still re-resolves to the new attempt's id ✓

Alternative to #1480 (which strips the keys in the mapper). This keeps the fix in the retry paths and is generic to any mapper-assembled input.

### DIFF
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/ai/AgentChatCompleteTaskMapper.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/ai/AgentChatCompleteTaskMapper.java
@@ -147,6 +147,9 @@ public class AgentChatCompleteTaskMapper extends AIModelTaskMapper<ChatCompletio
             validateRunnableConversation(chatCompletion);
             ensureEndsWithUserMessage(chatCompletion, taskModel);
             ensureJsonOutputUserMessage(chatCompletion);
+            // Detach the assembled conversation from the task's definition copy so a retry can't
+            // overwrite it by re-resolving the template (see AIModelTaskMapper).
+            detachAssembledInputFromDefinition(taskModel, "messages", "tools");
         } catch (Exception e) {
             if (e instanceof TerminateWorkflowException) {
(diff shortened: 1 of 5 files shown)


manan164:
Closing as duplicate: the LLM chat retry fix here (detach messages/tools from the task's definition copy in the mapper) is the same mechanism already in #1480 (`preserveAssembledChatInputOnRetry`). Consolidating on #1480.

Issue #1280, opened 2026-07-13
Spring Boot 4 compatibility: server artifacts hard-reference Boot-3-only FQCNs (Jackson2ObjectMapperBuilderCustomizer, DataSourceAutoConfiguration)


MaRsOnIx:
### The feature, motivation and pitch

**Describe the problem**

We embed the Conductor server modules (`conductor-core`, `conductor-rest`, `conductor-postgres-persistence`, `conductor-scheduler-*`) inside our own Spring Boot application. After upgrading the host application to Spring Boot 4.1 (Spring Framework 7 / Jackson 3), the application fails at startup, because the published Conductor artifacts were compiled against Spring Boot 3-only class locations that no longer exist in Boot 4.x:

1. `conductor-common` — `com.netflix.conductor.common.config.ObjectMapperBuilderConfiguration`:

```java
// common/src/main/java/com/netflix/conductor/common/config/ObjectMapperBuilderConfiguration.java
import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer; // L15

@Bean
public Jackson2ObjectMapperBuilderCustomizer conductorJackson2ObjectMapperBuilderCustomizer() { ... } // L28
```

In Boot 4.x this SPI was moved out of `spring-boot-autoconfigure` into the optional `spring-boot-jackson2` compatibility module **under a new package** (`org.springframework.boot.jackson2.autoconfigure.Jackson2ObjectMapperBuilderCustomizer`). Adding `spring-boot-jackson2` therefore does not satisfy the already-compiled reference — the old FQCN no longer exists on any classpath → `NoClassDefFoundError` at context startup. On top of that, `spring-boot-jackson2` is deprecated since Boot 4.0 as an explicit stop-gap, with removal expected around the Spring Framework 7.1 timeline — so there is no long-term path on the consumer side.

2. `conductor-postgres-persistence` and `conductor-sqlite-persistence`:

```java
// postgres-persistence/.../config/PostgresConfiguration.java (L29, L47)
// sqlite-persistence/.../config/SqliteConfiguration.java (L29, L55)
import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;

@Import(DataSourceAutoConfiguration.class)
```

[...]

PR #1531, opened 2026-08-12
Upgrade to Spring Boot 4.1.0

rq-jwhitlock:
# Upgrade to Spring Boot 4.1.0

Moves the build from Spring Boot 3.5.x to **4.1.0** (Spring Framework 7.0.x) on JDK 21.

Opening this to start a conversation as much as to land code — if Spring Boot 4 isn't on the
roadmap for the `3.32.x` line yet, I'm happy to hold it, retarget it, or split it into a
series of smaller PRs. It's been validated end to end (details below) and is in production
use downstream, so it's a working starting point either way.

## What Spring Boot 4 requires

**Per-technology autoconfigure modules.** Spring Boot 4 splits the single
`spring-boot-autoconfigure` jar into per-technology modules, so classes that previously
arrived transitively are now direct dependencies, and their packages moved:

| Spring Boot 3 | Spring Boot 4 |
|---|---|
| `boot.autoconfigure.jdbc` | `boot.jdbc.autoconfigure` |
| `boot.autoconfigure.flyway` | `boot.flyway.autoconfigure` |
| `boot.actuate.autoconfigure.metrics` | `boot.micrometer.metrics.autoconfigure` |
| `boot.autoconfigure.http` | `boot.http.converter.autoconfigure` |

Modules importing those classes now declare the corresponding `spring-boot-<technology>`
artifact directly rather than relying on a transitive.

**Jackson 2 and 3 coexist.** Spring Boot 4 ships Jackson 3 (`tools.jackson`) as the default
JSON mapper, while Jackson 2 (`com.fasterxml.jackson`) stays on the classpath. Conductor's
model classes carry Jackson 2 annotations, and `jackson-annotations` is the artifact both
versions read. Jackson 3's `JsonMapper$Builder` static initializer needs `@JsonSerializeAs`
from `jackson-annotations` >= 2.21, so the version-alignment rule in `build.gradle` now
**exempts `jackson-annotations`** and defers to the BOM. Clamping it below 2.21 fails web
server startup with `NoClassDefFoundError`.

**`OkHttp3ClientHttpRequestFactory` was removed in Framework 7.** `conductor-ai` used it, so
this adds a trimmed local equivalent (`OkHttpClientHttpRequestFactory` in
`org.conductoross.conductor.ai.http`) ported from Framework 6.2's implementation, with tests
covering it via `MockWebServer`. Swapping to `JdkClientHttpRequestFactory` instead would
silently drop any interceptor configured on the supplied `OkHttpClient`, which the AI module
relies on — there's a test that specifically guards that behaviour.

**Dependency floors** that Spring Boot 4 needs: Flyway 12.4.0, Jedis 7.4.1, httpclient5
5.6.1, `jackson-bom` 2.21.4. `spring-retry` is no longer managed by the Spring Boot BOM, so
it's pinned explicitly (2.0.13) — 23 unversioned declarations across 17 modules would
otherwise fail to resolve. Adds `junit-platform-launcher` as `testRuntimeOnly`, which Gradle
requires for the JUnit Platform version Spring Boot 4 manages.

## Second commit: BOM overrides that became downgrades

This is worth reviewing on its own. `springboot-bom-overrides.gradle` pinned patched versions
as CVE floors against the Spring Boot 3.3 BOM. Under 4.1.0 the BOM already manages all of
them at or above those floors, so **keeping the overrides downgrades three libraries**:

| override | SB4.1.0 BOM | effect |
|---|---|---|
| `spring-framework` 6.2.19 | 7.0.8 | downgrade — **breaks the compile** |
| `spring-security` 6.5.11 | 7.1.0 | downgrade |
| `netty` 4.1.135.Final | 4.2.15.Final | downgrade |
| `log4j2` 2.25.4 | 2.25.4 | no-op |
| `kafka` 3.9.1 | 4.2.1 | intentional — **kept** |

The `spring-framework` one is not a subtle regression: Spring Boot 4 *is* Framework 7, so
pinning 6.2.19 reverts the framework wholesale and fails to compile, because
`AbstractStreamingClientHttpRequest` is public in 7.x and package-private in 6.2.x. The Kafka
pin stays, since Kafka 4.x drops broker APIs the event-queue client still targets — it's a
compatibility constraint rather than a security floor, and the comment now says so, to keep
the two kinds of pin from being confused again.

## Verification

`./gradlew test -x :conductor-test-harness:test --continue` on `main` @ `1bad2c88d`:
**3722 tests, 2 failures**, both pre-existing and unrelated:

- `PostgresLockDAOTest.testLockAcquisitionAndRelease` — "Lease expiration is too early".
  A timing flake comparing container `now()` against host `Instant.now()`. **Fails with an
  identical message on unmodified `main`**, and passes in isolation.
- `PostgresConfigurationDataMigrationTest.dataMigrationIsNotAppliedWhenDisabled` — "Detected
  applied migration not resolved locally: 13.2". Passes in isolation; `V13.2` lives in the
  separate `migration_postgres_data` location, so this is a shared-database ordering artifact.

For reference, unmodified `main` fails **6** of 124 in `conductor-postgres-persistence` (that
same lock flake plus five `PostgresGrpcEndToEndTest` context-load failures), so this branch is
not making that module worse.

`spotlessCheck` is clean. Not yet exercised: the `test-harness` module and the Cassandra and
ES8 paths.

## Notes

Happy to split this into reviewable commits per concern — autoconfigure module moves, Jackson
alignment, the OkHttp factory, and the BOM override cleanup are all independent. Also glad to
rebase whenever it suits; I've had to move the base three times already while preparing it, so
I'm expecting that.


rq-jwhitlock:
This is a larger PR because it's upgrading to Spring Boot 4. I have been looking, but didn't see any action yet on that (though I saw comments requesting it in slack). Since I've done a lot of SB4 upgrades for my own company recently, I went ahead and did it here. I built it in my own environment, then tested it successfully in one of our services that implements it.

nthmost-orkes:
Hi @rq-jwhitlock -- thanks for the submission, and while I understand the basis for the idea, I have to close this PR on two different grounds.

First off, we have to follow some fairly complex internal planning for dependency and architecture shifts that have to do with dual support for OSS and Orkes (enterprise).

Second, though perhaps more pertinently here -- we really need issues filed on the repo before we are willing to accept external contributions.  That's how we can integrate external submissions with our internal planning considerations.

Thanks for using Conductor OSS and for the submission.  Please (again) feel free to submit an issue proposing this work.  It's unlikely we'll want a PR for this right away, but over time the issue itself may gather more support from other users and find its way into our planning schedule.

rq-jwhitlock:
Hi @nthmost-orkes , thanks for the response.  

I certainly understand planning, and architecture shifts. I also understand you have procedures around acceptance of external contributions.  I get all that. 

However, let me add more context around why this is needed, why I contributed when I did, and why you might be brushing this off a little too quickly. 

Spring Boot 3 was end-of-support on June 30, 2026. That means security vulnerabilities will be found that cannot/will not be fixed. There are some CVEs currently, right now, in the SB3 line that will concern any company who is concerned about breaches/infiltrations (that should be everybody). 

The conductor-oss latest versions you are building, including most recent release candidates, are still built on SB3. Your libraries will not function with an application or service that has been upgraded to SB4.  So there's currently no workaround. 

I went to your github repo for conductor-oss, and found that an issue HAD ALREADY been entered by a user. And, I also found that you, Naomi, logged this as a bug. Here is the issue:
https://github.com/conductor-oss/conductor/issues/1280

Even though I wasn't the originator of the issue, I did respond in the thread, adding that I also have the problem and immediate need.

However, your labeling it a bug, but no other follow-ups, seemed to indicate nothing is being done from your team.  I contributed because I had the ability and the need. 

It's not really a bug--that's minimizing the issue. Instead, it should be a normal upgrade path for any serious organization  wanting to create, maintain, and expand viable products. It's not something users should have to report, and vote on for priority.  As of right now, our security reviews have identified your libraries out of compliance. 

Again, I understand closing the PR, due to your procedures and potential architecture changes to which I'm not privy. But  doing so without providing even a hint of what may help me, that other user who first reported the issue, and any other users who NEED an upgrade like this perhaps without even knowing yet, is a huge concern. Can I get more idea of the SB4 roadmap, and when I can hope for some traction on this?

nthmost-orkes:
@rq-jwhitlock OK, thanks -- the labelling shouldn't concern you, we don't prioritize work based on labels.

bradyyie:
https://github.com/conductor-oss/conductor/pull/1573

PR #1537, opened 2026-08-13
fix(build): bump springdoc to 2.8.6 for Spring Framework 6.2 compatibility

mp-orkes:
`GET /api-docs` returns 500, so Swagger UI is unusable:

```
NoSuchMethodError: 'void ControllerAdviceBean.<init>(java.lang.Object)'
```

springdoc 2.1.0 calls `ControllerAdviceBean(Object)`, removed in Spring Framework 6.2. `8f78c0a78` bumped Spring Boot 3.3.5 → 3.5.14 (Framework 6.1 → 6.2) and left `revSpringDoc` at 2.1.0. It links lazily on first call, so the server starts clean and only `/api-docs` breaks.

## Notes

- Nothing caught this because no test had ever requested the document.
- The test uses a local `@RestControllerAdvice` instead of `ApplicationExceptionMapper` — springdoc only needs some advice bean present, and importing the production one into a context breaks `ApplicationExceptionMapperTest`, which installs a JVM-wide static `LoggerFactory` mock into the shared Gradle test JVM.
- `kafka/build.gradle` drops two unused test deps. mockserver pulls the pre-jakarta `io.swagger.core.v3:swagger-core`, which duplicates springdoc's `swagger-core-jakarta` packages and is missing `ObjectMapperFactory.createJson31()` — enough to break the module's Spring context under 2.8.6. Nothing in `kafka/src` references mockserver or testcontainers, so removing them leaves only the jakarta artifacts on the classpath.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -62,7 +62,7 @@ ext {
     revJq = '0.0.13'
     revJsr311Api = '1.1.1'
     revMockServerClient = '5.12.0'
-    revSpringDoc = '2.1.0'
+    revSpringDoc = '2.8.6'
     revOrkesQueues = '2.0.0.rc3'
     revPowerMock = '2.0.9'
     revProtogenAnnotations = '1.0.0'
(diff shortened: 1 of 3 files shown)


nthmost-orkes (approved):
get it

PR #1540, opened 2026-08-14
fix(ui): escape non-ASCII chars in workflow save body to prevent WAF corruption

shaileshpadave:
Workflows with em-dash characters (—) fail to save or clone from the UI on LinkedIn's environment. The em-dash is 3 bytes in UTF-8 (E2 80 94) — their corporate proxy strips bytes in the 0x80-0x9F range, which corrupts the sequence and makes Jackson reject the body with a JSON parse error.

Direct API and SDK calls work fine because they don't route through the browser's proxy. The corporate WAF/proxy sits in the browser's network path — it intercepts HTTP traffic from the browser but curl, Postman, and SDK calls go directly to the server without touching it. The intermittent success on retry is also consistent with this — LinkedIn runs a fleet of proxy nodes and not every node strips the bytes, so retrying sometimes hits a clean one.

The fix is simple: before sending any JSON body, escape all non-ASCII characters as \uXXXX sequences. The body becomes pure ASCII, nothing for the proxy to corrupt, and Jackson decodes the escapes back transparently on the server side. Applied it in fetchWithContext so all outgoing requests are covered in one place. Added a JSON.parse guard so non-JSON bodies like raw prompt templates are left untouched.

Covers save, update, and clone paths. Unit tests for the encoding function + e2e integration tests for both save and clone.

Full writeup: https://docs.google.com/document/d/1r1B_HyzaUpwYb2MaCz0kN-RUjUMAS9BNKsZVRgupmO0/edit?usp=sharing


c4lm:
~~Please feature flag this, off by default~~ if you mean to keep it, ensure all paths are covered.
What about SDKs? This looks like a problem on the target and not in conductor

ui-next/src/pages/definition/confirmSave/state/services.ts
```diff
@@ -45,7 +47,7 @@ export const createWorkflow = async (
           ...authHeaders,
         },
 
-        body: editorChanges,
+        body: asciiSafeJson(editorChanges),
```
ajlai-orkes: since it's already in fetchWithContext aren't you essentially calling asciiSafeJson twice?

ajlai-orkes:
please add e2e integration tests

ui-next/src/pages/definitions/dialog/CloneWorkflowDialog.tsx
```diff
@@ -136,11 +136,13 @@ const CloneWorkflowDialog = ({
           createWorkflowAction.mutate({
             method: HTTPMethods.POST,
             path: WORKFLOW_METADATA_BASE_URL,
-            body: JSON.stringify({
-              ...clonedWorkflow,
-              version: 1,
-              name: newName,
-            }),
+            body: asciiSafeJson(
```
ajlai-orkes: same here, asciiSafeJson is already in place in fetchWithContext so this is now redundant

ui-next/src/plugins/fetch.ts
```diff
@@ -40,6 +41,10 @@ export async function fetchWithContext(
 ): Promise<any> {
   const newParams = { ...fetchParams };
 
+  if (typeof newParams.body === "string") {
+    newParams.body = asciiSafeJson(newParams.body);
```
ajlai-orkes: is this valid for every use of fetchWithContext? there should be a json check as well. it looks like 
in cloudTemplates.ts, cloud template prompt save posts the raw template string:
```
await fetchWithContext(
  `/prompts/${prompt.name}${toMaybeQueryString(promptObj)}`,
  ...
    body: prompt.template,
```

ui-next/src/utils/strings.ts
```diff
@@ -48,3 +48,13 @@ export const getSequentiallySuffix = ({
...
 export const toUpperFirst = (str: string) => _upperFirst(_lowerCase(str));
+
+// Escape non-ASCII chars so a JSON string is ASCII-safe before HTTP transit.
+// Some WAF/proxy deployments strip C1 bytes (0x80–0x9F) from the raw body,
+// corrupting multibyte UTF-8 sequences (e.g. em-dash E2 80 94) and causing
+// Jackson to throw "JSON parse error: Invalid UTF-8 middle byte".
+export function asciiSafeJson(json: string): string {
+  return json.replace(/[-￿]/g, (ch) =>
+    `\\u${ch.charCodeAt(0).toString(16).padStart(4, "0")}`,
+  );
```
ajlai-orkes: LLM suggestion:
Works, but the class is literal U+0080–U+FFFF in the file. Prefer /[\u0080-\uffff]/g so the source itself stays ASCII-safe and readable in diffs/reviews.

ui-next/src/pages/definition/confirmSave/state/services.ts
shaileshpadave: Removed call-site asciiSafeJson

ui-next/src/pages/definitions/dialog/CloneWorkflowDialog.tsx
shaileshpadave: Removed call-site asciiSafeJson

ui-next/src/plugins/fetch.ts
shaileshpadave: Added JSON.parse guard: only encodes when body is valid JSON

ui-next/src/utils/strings.ts
shaileshpadave: Changed to /[\u0080-\uFFFF]/g

shaileshpadave:
> This looks like a problem on the target and not in conductor

Agreed it's caused by LinkedIn's network, not a conductor bug. The fix is on our side because customers can't change their corporate WAF/proxy configuration. The encoding is transparent, Jackson decodes \u2014 back to — on the server, nothing changes in storage or behavior.

shaileshpadave:
SDKs and direct API calls aren't affected here — the corporate WAF/proxy only sits in the browser's network path. Browser requests from the UI go through it, curl/Postman/SDK calls go directly to the server. So this is purely a UI-side fix and no SDK changes are needed.

Agreed it's a problem on LinkedIn's infrastructure, but customers can't change their corporate proxy config (and even if they could, rolling it out uniformly across a fleet of proxy nodes takes time — the customer also noted that retrying sometimes works, which points to inconsistent behavior across nodes). The encoding is fully transparent: Jackson decodes — back to — on the server, nothing changes in what gets stored or returned.

All paths are covered — applied it in fetchWithContext with a JSON.parse guard, so every outgoing JSON body goes through it in one place.

ajlai-orkes (approved):

PR #1543, opened 2026-08-15
fix(a2a): add authentication check to A2AAgentServerResource endpoints

shaileshpadave:
The 3 A2A server endpoints had no auth check at all — anyone could list, discover, or invoke agents without being authenticated.

Added `@PreAuthorize("isAuthenticated()")` to `listAgents`, `agentCard`, and `jsonRpc` in `A2AAgentServerResource`.


shaileshpadave:
Closing — conductor-oss has no Spring Security wired, so @PreAuthorize is a no-op here. Auth enforcement belongs in orkes-conductor via OrkesA2AAgentServerResource.

PR #1550, opened 2026-08-19
fix(core): correct getTaskDuration over-counting and guard malformed retry chains

manan164:
`getTaskDuration` recursed down the `retriedTaskId` chain as `return s + getTaskDuration(s, ...)`, passing the running sum down and then adding it again, so every hop double-counted everything before it.

| retry chain durations | reported | correct |
|---|---|---|
| 10, 20 | 40 | 30 |
| 10, 20, 30 | 100 | 60 |

The value feeds `Monitors.recordTaskExecutionTime(..., includeRetries=true)`, so that metric is inflated by a compounding factor for every retried task.

Now walks the chain iteratively and sums once. A visited set makes a malformed chain harmless — a self-reference or cycle previously recursed until the stack overflowed, killing the thread mid task-update.

### DIFF
--- a/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
+++ b/core/src/main/java/com/netflix/conductor/core/execution/WorkflowExecutorOps.java
@@ -1966,12 +1966,17 @@ public class WorkflowExecutorOps implements WorkflowExecutor {
     }
 
     private long getTaskDuration(long s, TaskModel task) {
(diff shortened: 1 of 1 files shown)


manan164:
Closing for now — will revisit later. Deferring the call on whether to keep the cumulative include-retries duration metric (and fix its double-count + cycle guard) or drop it entirely.

manan164:
Closing this PR since we will take this later while auditing all task related metrics.

PR #1551, opened 2026-08-19
feat(common): add TaskDef.regionDurable and RerunWorkflowRequest.rerunParentDownstreamTasks flags

manan164:
Two independent request/definition flags, both defaulting to `false` so existing behaviour is unchanged. Neither does anything on its own — each is read by the engine path that implements it.

### `TaskDef.regionDurable`

Marks a task type whose updates should be replicated region-durably in an active-active deployment — committed locally and synchronously confirmed to the peer region(s) before ack, with asynchronous convergence as fallback.

### `RerunWorkflowRequest.rerunParentDownstreamTasks`

Rerunning from a task inside a sub-workflow can either re-run just that workflow, or also re-run the downstream tasks of every ancestor. The second is useful when a corrected result has to flow up through the parents, but it re-runs tasks that already **completed**, so it cannot be the default. With the flag off, a rerun continues to touch only the workflow named in the request, leaving ancestors' completed tasks alone.

### Note on proto

Neither field is annotated `@ProtoField`, matching the existing `enforceSchema` precedent. An annotated field needs a matching entry in its `.proto` and in `AbstractProtoMapper`; adding the annotation without both would silently drop the value over gRPC. Both flags are JSON/REST-only for now.

### DIFF
--- a/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
+++ b/common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
@@ -157,6 +157,7 @@ public class TaskDef extends Auditable {
      * before the caller is acked (falling back to asynchronous convergence if the sync quorum is
      * missed). Defaults to false (locally durable only).
      */
+    @ProtoField(id = 26)
     private boolean regionDurable;
 
     public TaskDef() {}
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/netflix/conductor/common/metadata/tasks/TaskDef.java
```diff
@@ -151,6 +151,15 @@ public enum RetryLogic {
...
     private boolean enforceSchema;
 
+    /**
+     * When true, updates to tasks of this type are replicated region-durably in an active-active
+     * deployment: the update is committed locally and synchronously confirmed to the peer region(s)
+     * before the caller is acked (falling back to asynchronous convergence if the sync quorum is
+     * missed). Defaults to false (locally durable only).
+     */
+    @ProtoField(id = 26)
+    private boolean regionDurable;
```
v1r3n, line 161: can we follow the same enum as workflow start request for consistency mode?

PR #1559, opened 2026-08-21
feat(common): add rerunParentDownstreamTasks to RerunWorkflowRequest

manan164:
Rerunning from a task inside a sub-workflow can either re-run just that workflow, or also re-run the downstream tasks of every ancestor. The second is useful when a corrected result has to flow up through the parents, but it re-runs tasks that already completed, so it can't be the default.

Adds the flag with `@ProtoField(id = 6)`, defaulting to `false` — the existing behaviour. No engine change here; this only makes the option expressible on the request.


manan164:
Folded into #1551, which carries both flags.

PR #1573, opened 2026-08-25
(WIP)Upgrade to Spring Boot 4.1.1 and Jackson 3

bradyyie:
Upgrades the framework stack to Spring Boot 4.1.1 and migrates the codebase from Jackson 2 to
Jackson 3. This brings Spring Framework 7, Spring Security 7, Tomcat 11, Jakarta EE 11, springdoc 3
and Spring AI 2.0 with it.

Versions now come from the Spring Boot BOM unless there is a stated reason not to. Overrides that
existed only because an older BOM lagged behind have been removed: framework, security, log4j2,
netty, kafka, groovy, tomcat, flyway, micrometer, awaitility and jackson.

- Elasticsearch, because Conductor ships selectable ES6/7/8 and OpenSearch backends and the BOM
  tracks a client that will not talk to the older clusters
- Testcontainers, because the BOM moved to 2.x and that migration is its own piece of work


bradyyie:
https://github.com/conductor-oss/conductor/issues/1280

bradyyie:
Branch will be used to create 4.0.0-alpha.X fixes made here or merges to this branch from other branches will be included

The reason for this:
- is to allow OSS to continue with 3.X.X versions and fixes without blocking anyone.
- Allows for alpha images for end users to use as pre-release

Fixes and PRs to this branch i am happy to approve HOWEVER please DO NOT file issues about this branch. Fix it or wait for it to be fixed 😄

PR #951, opened 2026-04-02
PR-951 Restore Redis Sentinel auth and support multiple sentinel addresses in redis-lock

rajeshwar-nu:
## Summary

This PR fixes two issues with Redis Sentinel connectivity in Conductor:

1. **Restores Sentinel auth** — PR #927 refactored Redis connectivity to Jedis 6 but broke Sentinel authentication, causing `All sentinels down` failures on secured deployments.
2. **Supports multiple sentinel addresses in redis-lock** — The `CONDUCTOR_REDIS_LOCK_SERVER_ADDRESS` config now accepts semicolon-separated sentinel endpoints for improved HA.

### Missing Sentinel auth (from PR #927 refactor)

After the Jedis 6 refactor, `RedisSentinelConfiguration` applied auth to the master client config but built a separate Sentinel client config without auth. In environments where Sentinel requires authentication, Conductor fails at startup:

```
All sentinels down, cannot determine where <master> is running.
```

### Single sentinel address in redis-lock

The redis-lock module's Sentinel config only accepted a single sentinel address. If that sentinel was down, the lock client couldn't discover the master — even though other sentinels were healthy. The cluster mode already supported multiple addresses via comma splitting, but sentinel mode did not.

### Sentinel auth restoration
- Reuse the configured credentials for Sentinel-node connections
- Reuse the configured SSL settings for Sentinel-node connections
- Keep the current `sentinelMasterName` behavior unchanged
- Add focused regression tests for the Sentinel client config
- Update the module README so the Sentinel behavior matches the implementation

### Multiple sentinel address support
- Split `serverAddress` on semicolons in the Sentinel path of `RedisLockConfiguration`, matching how cluster mode already splits on commas
- This allows configuring multiple sentinel endpoints, e.g.:
  ```
  CONDUCTOR_REDIS_LOCK_SERVER_ADDRESS=redis://sentinel-0:26379;redis://sentinel-1:26379
  ```

### Merge conflict resolution
- Merged latest `upstream/main` into `fix/redis-sentinel-backcompat`
- Resolved the conflict in `RedisSentinelConfiguration` by keeping the newer `RedisSentinelCondition` conditional wiring from `main`
- Preserved this PR's Sentinel backcompat behavior: separate master and Sentinel client configs, both sharing configured auth and SSL behavior

## Validation

- `./gradlew spotlessApply`
[...]

PR #927, opened 2026-03-26
Refactor redis and upgrade jedis

v1r3n:
Pull Request type
----
- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] WHOSUSING.md
- [ ] Other (please describe):

**NOTE**: Please remember to run `./gradlew spotlessApply` to fix any format violations.

Changes in this PR
----

_Describe the new behavior from this PR, and why it's needed_
Issue #

Alternatives considered
----

_Describe alternative implementation you have considered_

PR #1578, opened 2026-08-27
feat(redis-sentinel): add conductor.redis.sentinelPassword, no AUTH to sentinels by default

### LINKED ISSUE #1577: Redis Sentinel: connection fails when sentinels have no requirepass (Jedis 6 rejects AUTH error reply)
## Describe the bug

Since #951, the sentinel client config reuses the master password (4th segment of
`conductor.redis.hosts`) for the AUTH sent to the sentinel nodes themselves. When the
sentinels do **not** have `requirepass` configured (a very common deployment — master
protected, sentinels open on a trusted network), the sentinel replies with an error to
the AUTH command. Jedis 6 treats that error reply as fatal, so the connection to the
sentinel fails and Conductor cannot start.

## Details

- Affected: `redis-configuration` module, `RedisSentinelConfiguration.createSentinelClientConfig(...)`
- Redis Sentinel without `requirepass` rejects `AUTH`: `ERR Client sent AUTH, but no password is set`
- Jedis 6 surfaces this as a connection failure (older Jedis tolerated it)
- Workaround today: set `requirepass` on every sentinel to the same value as the master password — not always possible/desirable

## Proposed fix (preferred): explicit `conductor.redis.sentinelPassword` property

Add a dedicated property for sentinel-node credentials, matching Spring Data Redis /
Lettuce `spring.data.redis.sentinel.password` semantics:

- `conductor.redis.sentinelPassword` **unset (default)** → no AUTH is sent to sentinels
- `conductor.redis.sentinelPassword` **set** → AUTH with that value (plus `conductor.redis.username` if set)
- Master/data-node auth is unchanged — still the 4th segment of `conductor.redis.hosts`

Note on backward compatibility: deployments that rely on the current behavior
(sentinels protected with the *same* password as the master) would need to set
`conductor.redis.sentinelPassword` explicitly after upgrading. This is a one-line
config migration, and it makes the previously-implicit coupling explicit.

I have this implemented with tests and can open a PR immediately if the approach is
agreed: https://github.com/jerrynieRe/conductor/tree/feat/redis-sentinel-password

## Alternative considered: fallback to master password

Keep the #951 behavior as the default (sentinel AUTH falls back to the master
password when `sentinelPassword` is unset) and only change behavior when the new
property is set, with some sentinel value (e.g. empty string) meaning "no AUTH".

This is fully backward compatible, but it does not fix the default-configuration bug
above (unauthenticated sentinels still break out of the box), and giving the empty
string a special meaning is awkward. Happy to go this way instead if maintainers
prefer strict compatibility.

## Environment

- Conductor: main
- Jedis: 6.x (as pinned by the project)
- Redis Sentinel topology with `requirepass` on the master only

jerrynieRe:
Fixes #1577.

Since v3.30.0, Conductor can't start against a really common Redis Sentinel setup (password-protected master, unauthenticated sentinels). Two things landed together in that release: the Jedis 3->6 upgrade (#927) and #951 (which made the sentinel client reuse the master password for AUTH). Jedis 6 treats the sentinel's "no password is set" reply as a connection failure, so startup dies with "All sentinels down".

This adds a dedicated `conductor.redis.sentinelPassword` and stops sending AUTH to sentinels by default:

- unset (default): no AUTH sent to sentinels
- set: AUTH with that value (plus `conductor.redis.user` if set)
- master/data-node auth is unchanged (still the 4th segment of `conductor.redis.hosts`)

This matches Spring Data Redis / Lettuce (`spring.data.redis.sentinel.password`), and it also matches what Conductor's own redis-lock module already does (Redisson only AUTHs sentinels when you explicitly set a sentinel password -- it never reuses the data-node one). So this just brings the Jedis path in line with the Redisson path we already ship.

Heads up on backward compat: if your sentinels are protected with the same password as the master, you now need to set `conductor.redis.sentinelPassword` explicitly. It's a one-line config change (it makes the previously implicit coupling explicit).

Verified against a live topology (password-protected master + unauthenticated sentinel): the new default connects and reads/writes fine, and the old behavior (AUTH to sentinel) fails with "All sentinels down" as expected. Unit tests cover no-auth-by-default, blank-is-no-auth, and sentinelPassword-when-set.


v1r3n:
@nthmost-orkes does this make sense?

jerrynieRe:
To add some context on why this change makes sense (we hit this in a real deployment):

**What happens today.** Since #951, the sentinel client config reuses the master password (4th segment of `conductor.redis.hosts`) for the AUTH sent to the sentinel nodes themselves. Our sentinels are provided by a managed Redis Sentinel service and have no `requirepass` — and there is no way for us to set one. A sentinel without `requirepass` answers AUTH with an error, and Jedis 6 fails hard on it during `initSentinels`, so every sentinel is marked down and the server cannot start:

```
2026-08-31 08:58:02.102 [main] WARN  redis.clients.jedis.providers.SentineledConnectionProvider - Could not get master address from <sentinel-host>:26379.
redis.clients.jedis.exceptions.JedisDataException: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
	at redis.clients.jedis.Protocol.processError(Protocol.java:110) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.authenticate(Connection.java:576) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.helloAndAuth(Connection.java:546) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.providers.SentineledConnectionProvider.initSentinels(SentineledConnectionProvider.java:181) ~[jedis-6.0.0.jar!/:?]
	at com.netflix.conductor.redis.config.RedisSentinelConfiguration.createUnifiedJedis(RedisSentinelConfiguration.java:91) ~[conductor-redis-configuration-3.30.2-SNAPSHOT.jar!/:?]
	... (full sanitized log at the end of this comment)

Caused by: redis.clients.jedis.exceptions.JedisConnectionException: All sentinels down, cannot determine where <master-name> is running.
```

Note this is a startup failure with no config workaround on the Conductor side — the only fix today is changing the sentinel deployment itself (set `requirepass` on every sentinel to the master's password), which is impossible on managed services.

**Why a separate property rather than reusing the master password.** Redis treats master auth and sentinel auth as independent: `requirepass` on the master says nothing about what the sentinels require. Clients that model this — Spring Boot / Lettuce (`spring.data.redis.sentinel.password`), redis-py (`sentinel_kwargs`) — all use a dedicated sentinel credential that defaults to "send no AUTH". This PR mirrors that semantic: master auth stays in `conductor.redis.hosts`, sentinel auth is opt-in via `conductor.redis.sentinelPassword`.

**Backward compatibility.** The only setups affected are those whose sentinels require the *same* password as the master (the implicit coupling introduced by #951). They keep working by setting `conductor.redis.sentinelPassword` to that password — a one-line migration that makes the coupling explicit. If you'd rather keep strict compatibility, the fallback variant (default to the master password when `sentinelPassword` is unset) is described in #1577 and I'm happy to rework the PR that way — it just leaves the unauthenticated-sentinel case broken by default, which is the common topology.

<details>
<summary>Full sanitized log</summary>

```
2026-08-31 08:58:01.364 [main] INFO  com.netflix.conductor.redis.config.RedisSentinelConfiguration - Starting conductor server using redis_sentinel, master=<master-name>, sentinels=[<sentinel-host>:26379], SSL=false
2026-08-31 08:58:02.102 [main] WARN  redis.clients.jedis.providers.SentineledConnectionProvider - Could not get master address from <sentinel-host>:26379.
redis.clients.jedis.exceptions.JedisDataException: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
	at redis.clients.jedis.Protocol.processError(Protocol.java:110) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Protocol.process(Protocol.java:158) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Protocol.read(Protocol.java:221) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.protocolRead(Connection.java:384) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:397) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.getStatusCodeReply(Connection.java:306) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.authenticate(Connection.java:576) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.helloAndAuth(Connection.java:546) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.initializeFromClientConfig(Connection.java:470) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Connection.<init>(Connection.java:77) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.Jedis.<init>(Jedis.java:76) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.providers.SentineledConnectionProvider.initSentinels(SentineledConnectionProvider.java:181) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.providers.SentineledConnectionProvider.<init>(SentineledConnectionProvider.java:101) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.JedisSentineled.<init>(JedisSentineled.java:36) ~[jedis-6.0.0.jar!/:?]
	at com.netflix.conductor.redis.config.RedisSentinelConfiguration.createUnifiedJedis(RedisSentinelConfiguration.java:91) ~[conductor-redis-configuration-3.30.2-SNAPSHOT.jar!/:?]
	... (Spring bean creation frames omitted)

2026-08-31 08:58:02.110 [main] WARN  org.springframework.boot.web.servlet.context.AnnotationConfigServletWebServerApplicationContext - Exception encountered during context initialization - cancelling refresh attempt: ...
    Error creating bean with name 'createUnifiedJedis' defined in class path resource [com/netflix/conductor/redis/config/RedisSentinelConfiguration.class]:
    Failed to instantiate [redis.clients.jedis.UnifiedJedis]: Factory method 'createUnifiedJedis' threw exception with message:
    All sentinels down, cannot determine where <master-name> is running.

Caused by: redis.clients.jedis.exceptions.JedisConnectionException: All sentinels down, cannot determine where <master-name> is running.
	at redis.clients.jedis.providers.SentineledConnectionProvider.initSentinels(SentineledConnectionProvider.java:209) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.providers.SentineledConnectionProvider.<init>(SentineledConnectionProvider.java:101) ~[jedis-6.0.0.jar!/:?]
	at redis.clients.jedis.JedisSentineled.<init>(JedisSentineled.java:36) ~[jedis-6.0.0.jar!/:?]
	at com.netflix.conductor.redis.config.RedisSentinelConfiguration.createUnifiedJedis(RedisSentinelConfiguration.java:91) ~[conductor-redis-configuration-3.30.2-SNAPSHOT.jar!/:?]
```

</details>

nthmost-orkes:
> @nthmost-orkes does this make sense?

Yeah, this does make sense.  #951 introduced not-fully-intended behavior that just surfed in with related changes.  

So while this will be a minor (like mosquito-sized) breaking change for a handful of people who may have started using the 5-months-old way of doing things, it brings the behavior back in line with what we actually want.

nthmost-orkes (dismissed):
This seems good, thanks for catching this.  What we need is a documentation treatment (you can throw your LLM at it, nbd), because:

* redis-configuration/README.md line 101 still describes the old #951 behavior 
* the properties table is missing a sentinel-password row

jerrynieRe:
Thanks for the review @nthmost-orkes! Documentation updated in 932043a:

* Rewrote the `RedisSentinelConfiguration` bullet (the old line 101) to describe the new behavior: sentinels reuse SSL settings but not the data-node password; AUTH goes to sentinels only when `sentinel-password` is set.
* Added a `sentinel-password` row to the connection properties table, and surfaced the property in the Sentinel quick-start example as a commented line.
* One drive-by fix while verifying the table against the code: `ignore-ssl` is honored in cluster **and** sentinel modes (`createBaseClientConfigBuilder` applies the trust-all context for sentinel too), so the "cluster mode only" note was corrected. Happy to drop that hunk if you'd rather keep this PR strictly scoped.

nthmost-orkes (approved):
Confirmed end-to-end, g2g

Releases where the erroneous behavior occurs: v3.30.0 and every release since.

PR #1591, opened 2026-08-31
[CCOR-13369]fix:Carry filters across the executions SQL format toggle

najeebkp:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
### Issue 
[CCOR-13369](https://orkes.atlassian.net/browse/CCOR-13369?atlOrigin=eyJpIjoiMDI1MGNmMzliNTU3NDU2MzljNjhmYzQ5ZTE2NjE1MTkiLCJwIjoiaiJ9)SQL format toggle does not populate the query box with the equivalent query for active filters (Workflow Executions)

### Fixes
Switching SQL format on left the query box empty and dropped the filters only basic search has a control for — workflow name, id, correlation ids, idempotency key, modified times, exclude sub-executions — from the request too, so results quietly widened. They are now translated into query text, and read back out when switching off. A query basic search cannot express prompts before being discarded. Status and the time bounds keep their own controls as the source of truth unless the query mentions them.

### Changes

- `basicFilterQuery.ts` (new) — translates the basic-only filters to query
  clauses and back, plus the merge policy for switching off.
- Toggling on seeds the SQL box via the `useQueryState` default, so the first
  search already carries the clauses instead of firing unfiltered and correcting.
- Toggling off parses the query back into the fields. Anything basic search
  cannot express (`OR`, grouping, unknown fields) prompts before being discarded,
  and discarding clears the param.
- Parsing is all-or-nothing: a query that is only partly understood applies
  nothing, rather than silently dropping the rest.
- Status and the start/end time bounds have controls in both modes, so they are
  only overwritten when the query mentions them — otherwise switching back wiped
  a status picked in the dropdown.

Extends the SQL format toggle fix to the Agents and Tasks execution search: switching it on seeds the query box with the equivalent query for the active filters, switching it off reads a representable query back into the fields (prompting first when it isn't), and Reset clears every filter rather than only the current mode's. Agents reuse the workflow clause builder; Tasks needed their own for a different field set. The task page holds both modes in one component, so its seed is recomputed on the toggle instead of riding on useQueryState's default, which is captured once on first render.

### Demo
Screen Recording 2026-08-31 at 6 02 49 PM

Enterprise UI Playwright Tests
----
Every PR automatically triggers the enterprise UI Playwright E2E test suite.
Tests run against conductor-ui `main` by default. To test against a different
conductor-ui branch, add this line anywhere in the PR description:

```
conductor-ui-branch: my-feature-branch
```


v1r3n (approved):

ui-next/src/pages/executions/workflowSearchComponents/AdvancedSearch.tsx
```diff
@@ -119,7 +120,35 @@ export default function AdvancedSearch({
...
+  const [basicWorkflowType] = useQueryState<string[]>("workflowType", []);
+  const [basicWorkflowId] = useQueryState("workflowId", "");
+  const [basicCorrelationIds] = useQueryState<string[]>("correlationIds", []);
+  const [basicIdempotencyKey] = useQueryState<string[]>("idempotencyKey", []);
+  const [basicModifiedFrom] = useQueryState("modifiedFrom", "");
+  const [basicModifiedTo] = useQueryState("modifiedTo", "");
+  const [basicExcludeSubExecutions] = useQueryState(
+    "excludeSubExecutions",
+    false,
+  );
```
ajlai-orkes: I'd create one useWorkflowSearchFilters() hook that owns every useQueryState key for this. The filters are meant to be bidirectional right?

ajlai-orkes:
This is only updating workflow search. What about Agents and Task executions search? Last time I wanted to modify workflow basic/advanced search, Deepti wanted to keep things consistent across all of them

ui-next/src/pages/executions/WorkflowSearch.tsx
```diff
@@ -77,6 +84,22 @@ export default function WorkflowPanel({
...
+  const [, setWorkflowType] = useQueryState<string[]>("workflowType", []);
+  const [, setWorkflowId] = useQueryState("workflowId", "");
+  const [, setCorrelationIds] = useQueryState<string[]>("correlationIds", []);
+  const [, setIdempotencyKey] = useQueryState<string[]>("idempotencyKey", []);
+  const [, setModifiedFrom] = useQueryState("modifiedFrom", "");
+  const [, setModifiedTo] = useQueryState("modifiedTo", "");
+  const [, setExcludeSubExecutions] = useQueryState(
+    "excludeSubExecutions",
+    false,
+  );
```
ajlai-orkes: likewise, here there's a bunch of repeated useQueryState. 

A separate issue is that because of the useQueryState, typing in the search fields is very slow due to the re-rendering. I really want to do a whole UI overhaul of these search filters

ui-next/src/pages/executions/workflowSearchComponents/basicFilterQuery.ts
```diff
@@ -0,0 +1,244 @@
...
+  const filters: ParsedBasicFilters = {};
+
+  for (const rawClause of trimmed.split(/\s+AND\s+/i)) {
+    const clause = rawClause.trim();
+    if (clause === "") {
+      continue;
+    }
+
+    const inMatch = clause.match(/^(\w+)\s+IN\s*\(([^()]*)\)$/i);
+    const opMatch = clause.match(/^(\w+)\s*(=|>|<)\s*(.*)$/);
```
dfont-orkes: this is introducing new and slightly different clause parsing. we should try to not add more differing clause parsing in the code with this PR.

this new parser is similar to the one in cliSearch.ts, but compared to it there are some differences. for example, the case insensitivity means that a lower case "and" splits, which diverges from what the Java code will do. also, in an `IN` if there are any parenthesis the entire match fails. this one does seem to parse `correlationId IN ('a,b')` more intuitively here than the Java code, but for compatibility reasons we may need to match exactly what the backend does in this PR. these are just some examples, I would suggest stress testing these regexes heavily.

In addition to stress-testing the regexes for compliance with other regexes in the code I would recommend turning this regex clause parsing into its own module which `basicFilterQuery.ts` would use. Ideally this would have been done earlier to officially establish how the query parsing should work (even better, there would be an official grammar instead of regex parsing), but introducing it now is still worthwhile. And then if it won't blow up the PR's size I would recommend then having cliSearch.ts use this new shared module instead of defining its own regexes. If it would be a very large effort, or if it would introduce too many changes in behavior at once, then at least put a TODO on it to have it refactored to use the shared module.

ui-next/src/pages/executions/workflowSearchComponents/AdvancedSearch.tsx
najeebkp: Updated

ui-next/src/pages/executions/workflowSearchComponents/basicFilterQuery.ts
najeebkp: Updated

ui-next/src/pages/executions/WorkflowSearch.tsx
najeebkp: Updated to a hook. and I think we should make the UI revamp for the searches in a separate ticket.

ajlai-orkes:
> This is only updating workflow search. What about Agents and Task executions search? Last time I wanted to modify workflow basic/advanced search, Deepti wanted to keep things consistent across all of them

@najeebkp bump on this. Can you check with @deeptireddy-lab?

najeebkp:
Added the same fix for agents and tasks search

Issue #1118, opened 2026-05-21
Epic: Port /api/schema resource from Orkes for input/output schema management


nthmost-orkes:
## Overview

Port the `/api/schema` REST resource (and its supporting service + persistence) from Orkes Conductor to OSS, so that input/output schemas can be defined once, stored centrally, and referenced from task definitions (`taskDef.inputSchema` / `outputSchema`). The UI already calls `/api/schema` for the Simple Task input/output schema picker, but the endpoint does not exist in OSS — today the picker has to be hidden behind a feature flag.

## Problem Statement

`SchemaDef` exists in OSS (`conductor/common/src/main/java/com/netflix/conductor/common/metadata/SchemaDef.java`) and `DeciderService` already consumes `taskDef.inputSchema` / `outputSchema` for validation. What's missing is the management surface: a REST resource, a service, and persistence so users can CRUD schemas and the UI can list them in pickers.

Orkes ships exactly this surface in `io.orkes.conductor.server.rest.SchemaResource` + `io.orkes.conductor.integration.service.SchemaService`. UI, Java SDK, and Python SDK all hit `/api/schema` with specific payload shapes and JSON field names. Diverging on path or shape would force every client to branch on backend type.

## Parity Strategy

Mirror the Orkes layout exactly, swapping only the namespace (`io.orkes.conductor.*` → `org.conductoross.conductor.*`).

| Orkes (source of truth) | OSS port |
|---|---|
| `io.orkes.conductor.server.rest.SchemaResource` | `org.conductoross.conductor.rest.SchemaResource` |
| `io.orkes.conductor.integration.service.SchemaService` | `org.conductoross.conductor.service.SchemaService` |
| `@RequestMapping("/api/schema")` | identical |
| Methods: `save`, `getSchemaByNameWithLatestVersion`, `getSchemaByNameAndVersion`, `getAllSchemas`, `deleteSchemaByName`, `deleteSchemaByNameAndVersion` | identical names + signatures |
| Query params: `?short=`, `?newVersion=` | identical |

## Orkes vs OSS: What Changes

| Component | Orkes | OSS | Change |
|---|---|---|---|
| REST path + payloads | `/api/schema` + `SchemaDef` | Same | Keep |
| `SchemaDef` model | `com.netflix.conductor.common.metadata.SchemaDef` | Same (already in OSS) | Keep |
| DAO interface | `SchemaDAO` (org-scoped) | Same signatures, OSS passes `"default"` orgId | Simplify |
[...]

PR #1592, opened 2026-08-31
Add schema registry to OSS Conductor

ambiorix2099:
Pull Request type
----
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

Adds a schema registry to OSS Conductor: storage on MySQL, PostgreSQL, SQLite and Redis, six REST endpoints at `/api/schema`, a management screen, and engine validation each definition opts into. Cassandra is deliberately not implemented.

It closes a gap that has been visible for a while: OSS Conductor references schemas in three places and serves them in none. `SchemaDef` ships and attaches to task and workflow definitions, the JSON Schema validator ships, the docs describe attaching schemas, and the five schema pickers in the UI query `/api/schema`, which no OSS server implements. They render as empty dropdowns and fail silently.

| Before: `/api/schema` returns 404 | After: the registry is served |
|---|---|
| Input and output schema pickers with an empty version field | The same pickers with the schema name and Version 2 resolved |

- **Storage** in `meta_schema_def`, keyed on (name, version), with new versions allocated against the unique constraint so two simultaneous saves cannot silently overwrite each other. Cassandra is the deliberate omission: `SchemaDAO` is a required dependency with no default implementation, so a Cassandra-backed server fails at startup rather than accepting writes it cannot store. That backend already fails to start on OSS for an unrelated reason, so no working deployment changes.
- **Six endpoints at `/api/schema`**, matching the contract the shipped SDK clients were generated from. No SDK changes needed.
- **`SchemaDAO` carries the enterprise interface method for method**, as asked in review, plus one addition: `createSchemaIfAbsent`. Version allocation goes through the unique constraint on (name, version) rather than reading the maximum and writing max + 1, so two simultaneous saves cannot silently overwrite each other; a writer that loses the race re-reads and retries, and reports a conflict rather than a lost write. There is a four-writer concurrency test per backend on it. The enterprise interface has no equivalent because that implementation allocates without locking. `findAllVersionsByName`, `getAllShortenedSchemas` and `deleteAllByNames` are present for parity and have no caller here yet.
- **A Definitions > Schemas screen**: list, clone, delete, and a JSON editor that saves in place or as a new version.
- **Engine validation at four points**: workflow input at start, task input at scheduling, task output at update, workflow output at completion. There is no server property; a payload is checked when the definition's own `enforceSchema` is set and a schema is attached. `enforceSchema` defaults to `false` on a Task Definition and `true` on a Workflow Definition, so attaching a schema to a workflow is enough on its own.

The Schemas list with two schemas, their types, latest versions and version counts

A schema of a type this server stores but cannot validate opens read-only, with a note saying so.

An AVRO schema shown read-only, with a notice explaining it is stored but not validated by this server

Also repairs three defects in the AI layer's output-schema validation. Each one turned a documented feature into a hard failure, so none of the three can break a working path.

What this does not do:

- `externalRef` is stored and returned, never dereferenced. AVRO and protobuf round-trip; only JSON is validated.
- System-task output and externalized task output are not validated.
- A schema reference the registry does not hold silently skips validation. The payload goes through unchecked; the only signal is the `schema_registry_miss` counter.
- A registered schema the validator cannot parse does the same, and is logged as an error. Omitting `$schema` is enough to trigger it.
- The Task Definition form does not flag an unregistered schema reference. The other three picker surfaces do. The enterprise UI pins a baseline of that form with its pickers unflagged, so flagging it here turns that repo red until the pin moves.

### How to test

```bash
# server + bundled UI, built from this branch. API on 8000, UI on 8127.
docker compose -f docker/docker-compose-postgres.yaml up -d --build
```

The registry is at `http://localhost:8127/schemas`, under Definitions > Schemas. Register a schema there or over the API, and the input/output pickers on the Simple Task, Yield Task, Workflow Properties and Task Definition forms list it.

Enforcement needs no restart and no server setting. Attach a schema to a definition and set `enforceSchema` on it; it applies from that definition's next execution. A JSON Schema needs a `$schema` line, or the server cannot parse it and skips validation.

```bash
curl -X POST http://localhost:8000/api/schema -H 'Content-Type: application/json' -d '[{
  "name":"customer_input","version":1,"type":"JSON","data":{
    "$schema":"https://json-schema.org/draft/2020-12/schema",
    "type":"object","properties":{"name":{"type":"string"}},"required":["name"]}}]'
```

With that attached as a workflow `inputSchema`, a violating start is rejected before the execution exists:

```bash
curl -X POST http://localhost:8000/api/workflow/<workflow> -H 'Content-Type: application/json' -d '{"nickname":"ada"}'
# {"status":400,"message":"Schema validation failed $.name: is missing but it is required","retryable":false}
```

Set `enforceSchema` to `false` on the same definition and the identical request starts normally. A conforming workflow input whose task input breaks the task's own schema fails at the task boundary instead:

```
workflow FAILED
task step FAILED_WITH_TERMINAL_ERROR
  Schema validation failed $.name: is missing but it is required
```

Verified on this branch against PostgreSQL, MySQL and SQLite, and through the Java, Python, Go and JavaScript schema clients.

Issue #1118

Alternatives considered
----

- Enforcement on by default, or with no property at all. Rejected: workflows that succeed today would start failing on upgrade.
- An in-memory `SchemaDAO` as the default so the service always starts. Rejected: SQLite is the default backend, so it would accept writes, report success and lose them on restart.
- Registering the UI screen through the plugin registry. Rejected: the registry has no precedence mechanism, so an OSS registration would displace the commercial schema editor. Cost of importing the dialogs directly instead is that the picker's inline edit and preview buttons stay unavailable here.
- Matching the commercial table name. Rejected: `meta_schema_def` differs deliberately, which removes the fresh-install collision when both products share a database.

Enterprise UI Playwright Tests
----
Runs against conductor-ui `main` by default; no `conductor-ui-branch:` line needed.


ui-next/src/components/inputs/ConductorNameVersionField.tsx
```diff
@@ -101,8 +101,14 @@ export const ConductorNameVersionField = forwardRef<
...
                 });
               }}
+              // Flag against a list the server actually answered with, empty or
+              // not: a name absent from an empty listing is still a name this
+              // server does not hold. What must not flag is a list that never
+              // arrived, since a request in flight or an endpoint that answers
+              // 404 would otherwise paint every field carrying a reference red.
               error={
                 showErrorIfItemNotInList &&
+                isSuccess &&
```
ajlai-orkes: this will no longer show error if isSuccess is false now though which I don't think is what we necessarily want

ambiorix2099: Good catch. Changed the gate from `isSuccess` to `(isSuccess || isError)`

Empty-but-successful listings still flag, which is what the input-schema-validation test depends on.

nthmost-orkes (dismissed):
So, couple of **OSS/Orkes convergence type observations** -- there are a few method renames that may have downstream consequences

Calling in @v1r3n and @bradyyie to look:  SchemaService in this PR calls findAllVersionsByName, getAllShortenedSchemas, and deleteAllByNames — none of which exist on the OSS interface. So Orkes cannot compile against this interface without either re-adding those methods or rewriting its service.

Probably fine but worth calling out:  I note that `createSchemaIfAbsent` is an OSS-side addition that probably improves correctness over what's implemented on the Orkes side.   Something worth calling out more prominently in the PR description.

ui-next/src/components/inputs/ConductorNameVersionField.tsx
```diff
@@ -101,8 +101,16 @@ export const ConductorNameVersionField = forwardRef<
...
+              // failed one leaves the reference unverifiable, which is worth
+              // showing. What must not flag is a request still in flight or one
+              // react-query has not started (a disabled query is idle, not
+              // loading), since either would paint every field that carries a
+              // reference red on the way to the answer.
               error={
                 showErrorIfItemNotInList &&
+                (isSuccess || isError) &&
                 value != null &&
                 !options.some(({ name }) => name === value?.name)
```
ajlai-orkes: should we always show error if isError is true? like:
```
(isError || (showErrorIfItemNotInList &&
                isSuccess &&
                value != null &&
                !options.some(({ name }) => name === value?.name))
```

ajlai-orkes: for ease of reading maybe do something like:
```
const hasInvalidValue =
  showErrorIfItemNotInList &&
  isSuccess &&
  value != null &&
  !options.some(({ name }) => name === value.name);

const hasError = isError || hasInvalidValue;
```

and then this error prop can be just 
```
error={hasError}
```

ui-next/src/components/inputs/ConductorNameVersionField.tsx
```diff
@@ -73,6 +73,23 @@ export const ConductorNameVersionField = forwardRef<
...
+    //
+    // Kept behind showErrorIfItemNotInList, including for a failed request: the
+    // Task Definition form and the react-hook-form wrapper do not opt in, and
+    // flagging them on a listing failure would put a red field on a form that has
+    // never claimed to check its references.
+    const hasUnverifiableReference =
+      showErrorIfItemNotInList &&
+      (isSuccess || isError) &&
+      value != null &&
+      !options.some(({ name }) => name === value?.name);
```
ajlai-orkes: you should still have a `const hasError = isError || hasUnverifiableReference`  like I had posted above as if not all the conditions are fulfilled and `isError` is true, this won't show an error

ambiorix2099:
@nthmost-orkes
Missing methods were added to the interface now. They were 

- `findAllVersionsByName`
- `getAllShortenedSchemas`
- `deleteAllByNames`

`createSchemaIfAbsent` has been removed

ui-next/src/components/inputs/ConductorNameVersionField.tsx
ambiorix2099: Done.

v1r3n (approved):

bradyyie:
Ran and tested , no obvious issues or system failures found !

bradyyie (approved):

shaileshpadave (approved):
lgtm, 
Fixed few AI generated comments
ran bunch of tests successfully.

ui-next/src/components/inputs/ConductorNameVersionField.tsx
ambiorix2099: Done: `hasError = isError || hasInvalidValue`, same as your follow-up.

redis-persistence/src/main/java/org/conductoross/conductor/redis/dao/RedisSchemaDAO.java
```diff
@@ -0,0 +1,171 @@
...
+                Comparator.comparing(SchemaDef::getName).thenComparingInt(SchemaDef::getVersion));
+        return schemas;
+    }
+
+    @Override
+    public int deleteByNameAndVersion(String name, Integer version) {
+        Objects.requireNonNull(version, "Schema version cannot be null");
+        Long removed = jedisProxy.hdel(nsKey(SCHEMA_DEF, name), String.valueOf(version));
+        // The name is only listable while some version of it survives.
+        if (jedisProxy.hkeys(nsKey(SCHEMA_DEF, name)).isEmpty()) {
```
dfont-orkes, line 98: there appears to be a similar race here that's has more specific pre-conditions (so should be less likely). if a request to just delete one schema version coincides with a POST /api/schema in such a way that the delete deletes from SCHEMA_DEF, with the POST call happening between the HKEYS and the SREM calls, then a similar issue will happen where the listings of schemas will be missing whatever version was posted but loading by the specific version will still work.

redis-persistence/src/main/java/org/conductoross/conductor/redis/dao/RedisSchemaDAO.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return removed == null ? 0 : removed.intValue();
+    }
+
+    @Override
+    public int deleteAllByName(String name) {
+        // DEL reports keys removed, not fields, so the count comes from the hash's own length —
+        // the same number the SQL backends' row count reports.
+        Long fields = jedisProxy.hlen(nsKey(SCHEMA_DEF, name));
+        jedisProxy.del(nsKey(SCHEMA_DEF, name));
```
dfont-orkes, line 109: there appears to be a potential race condition here where one actor issues a DELETE /api/schema/foo then another actor issues a POST /api/schema which ends up doing an SADD between the DEL and the SREM here. The consequence is that `SCHEMA_DEF_NAMES` will no longer contain the schema but if queried for a specific version of the schema it will find it due to looking in the SCHEMA_DEF hash. I'm not sure what the likelihood is of this happening.

manan164 (approved):

PR #1595, opened 2026-09-01
Add size-based review-policy gate (extra approval for large PRs)

nthmost-orkes:
## What

Adds a **review-policy** gate: PRs that change **≥ 2000 lines** require **2 approvals**. Normal-size PRs are unaffected.

GitHub branch protection can only enforce a *fixed* minimum number of approvals — it can't vary that number by diff size. This workflow adds the size-based tier on top, while keeping its signal cleanly separate from CI.

## Signals are separated from CI on purpose

A review gate that shows a red X alongside the test checks reads like a broken build and discourages reviewers before they engage. So this is designed so the approval requirement is never confused with a failing test:

- **Named as policy, not a test.** The blocking check is `review-policy / approvals`. A red row there is unmistakably a review gate.
- **Primary signal is a label + sticky comment.** On a large under-approved PR the workflow adds a `needs: 2 approvals` label (plus `size: XL`) and posts a sticky comment with the current count. Reviewers see the requirement in the labels/timeline, not only in the checks row. Both are cleared automatically once satisfied.
- **Baseline stays native.** The baseline 1-approval requirement is delegated to native branch protection (GitHub's "Review required" banner), so normal-size PRs show a clean green check here.

## How it works

- Triggers on PR open/sync/reopen/ready-for-review (`pull_request_target`) and on review submitted/dismissed, so it re-evaluates when approvals change.
- `pull_request_target` is used so the token is writable on **fork** PRs (needed to manage the label + comment). The workflow **never checks out or executes PR code** — it only reads PR metadata via the API — so this is the safe use of that trigger.
- Sums `additions + deletions` across changed files, **excluding generated / vendored / binary files** so they don't push a PR into the large tier:
  - Lockfiles: `ui-next/pnpm-lock.yaml`, `ui/package-lock.json`, `ui/yarn.lock`, any `*.lock`
  - Vendored: `server/src/main/resources/swagger-ui/**`, any `**/vendor/**`
  - Generated / snapshots: `**/generated/**`, `**/__snapshots__/**`, `*.snap`, `*.min.js`, `*.min.css`
  - Binary assets: images, fonts, PDFs
- Counts the **latest non-comment review per user** as the approval tally.
- Skips draft PRs (and clears any stale label).

## Tuning

Thresholds live at the top of the script:

```js
const LARGE_PR_THRESHOLD = 2000;   // changed lines (adds + dels)
const APPROVALS_FOR_LARGE = 2;
const APPROVALS_DEFAULT = 1;        // enforced by branch protection, not this check
```

## To enforce

1. Set the repo's native branch protection / ruleset for `main` to require **1 approval** (the baseline).
2. Add `review-policy / approvals` as a **required status check**. It gates merge only for the large tier.

## Note

The one honest limitation: any signal that actually *blocks* merge shows up as a red required check — GitHub has no "blocking but not red" channel (`neutral`/`skipped` are treated as passing). This design keeps that unavoidable red clearly labeled as a review gate and moves the everyday, at-a-glance signal into a label + comment.


nthmost-orkes:
Self review: kinda hate it?   Still massaging here, don't at me yet.

v1r3n (approved):

Issue #1583, opened 2026-08-28
Every `SUB_WORKFLOW` is reported as a handoff, so an agent invoked as a tool is never a tool call


ambiorix2099:
## Bug Description

`agent_tool` compiles to `SUB_WORKFLOW`, and the listener emits a `handoff` event for any
`SUB_WORKFLOW` without checking which of the two it is. So calling another agent as a tool
produces a handoff event and no tool event.

Both readings break. "Did this agent hand off to X?" answers yes when X was only called as a tool,
and "was tool X called?" finds nothing. The emitted handoff target is not an agent name either;
it is the tool task's reference name, which is built from the LLM provider's tool-call ID.

Separately, every sub-workflow emits two `handoff` events with different targets.

## Root Cause

All in
`agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/service/AgentEventListener.java`.

**Unconditional handoff.** `onTaskScheduled` (`:78-84`) emits `AgentSSEEvent.handoff(...)` for any
`SUB_WORKFLOW`, with the target parsed from the reference name. `ToolCompiler.TYPE_MAP` maps
`agent_tool` to `SUB_WORKFLOW` (`ToolCompiler.java:146`), and the compiled dispatch confirms it
(`runtime/util/JavaScriptBuilder.java:725-728`).

The target is wrong as well. A tool task's reference name is
`call_<provider-tool-call-id>_<fork-index>__<iteration>`. `extractHandoffTarget` (`:487-512`)
strips the `__<iteration>` suffix at `:491`, then fails to match either the strategy pattern or the
index pattern, so it returns `call_<provider-tool-call-id>_<fork-index>` as the handoff target.

**Duplicate emission.** `onWorkflowStartedIfEnabled` emits a second handoff at `:222` for any child
workflow with a parent, which includes every agent-tool sub-workflow. One target comes from the
task reference name, the other from `childName.replaceAll("_wf$", "")`.

The comment at `:205-206` justifies the second site by asserting that `onTaskScheduled` is not
called for system tasks like `SUB_WORKFLOW`. That is not correct. `SubWorkflow.isAsync()` returns
`true` (`core/src/main/java/com/netflix/conductor/core/execution/tasks/SubWorkflow.java:328-330`),
so the task is queued and `taskStatusListener.onTaskScheduledIfEnabled(task)` fires at
`WorkflowExecutorOps.java:2130`.

## Steps to Reproduce

[...]

PR #1610, opened 2026-09-08
Fix tool-call selection and naming in the agent event listener

### LINKED ISSUE #1584: `isToolTask` denylists four real tool kinds, and its default branch admits any worker task as a tool
## Bug Description

Tool-call SSE events are wrong in both directions.

Four real tool task types are on the denylist, so no `tool_call` or `tool_result` event is
emitted for them: `mcp`, `human`, `http`/`api` and `agent_tool` tools. And the `default` branch
applies no effective filter, so a task that is not a tool emits a spurious pair. `LIST_MCP_TOOLS`,
the discovery task compiled into every MCP-using agent, is reported as a tool call named
`list_mcp_tools`.

So a consumer asking "was tool X called?" can get the wrong answer either way, and one asking
"was tool X *not* called?" gets a confident yes.

## Root Cause

`isToolTask` (`AgentEventListener.java:403-429`) denylists twelve task types and treats
everything else as a tool.

**Four denylist entries are real tool types.** `SUB_WORKFLOW` (`:419`), `HUMAN` (`:420`), `HTTP`
(`:422`) and `CALL_MCP_TOOL` (`:423`). `ToolCompiler.TYPE_MAP` (`ToolCompiler.java:140-153`)
compiles real tools onto exactly those: `http` and `api` to `HTTP`, `mcp` to `CALL_MCP_TOOL`,
`agent_tool` to `SUB_WORKFLOW`, `human` to `HUMAN`. That is five SDK tool types across four task
types.

**The `default` branch filters nothing.** `:427` returns
`"SIMPLE".equals(taskType) || task.getTaskDefinition().isPresent()`. The second disjunct reads as
"has a registered TaskDef", but `MetadataMapperService.java:120-135` populates a `TaskDef` for any
task with a non-blank name, synthesizing an ad-hoc one when the DAO has none, under a gate
(`:195-200`) that never looks at task type. So it is true for essentially every task, and any
worker task type not explicitly denylisted is treated as a tool. `LIST_MCP_TOOLS`
(`ai/.../MCPWorkers.java:58`) is the case that bites today; the structural problem is that a new
worker task type becomes a tool by default, with nobody deciding.

## Steps to Reproduce

Register an agent with one MCP tool and subscribe to its SSE stream.

- A spurious `tool_call`/`tool_result` pair arrives named `list_mcp_tools`, from the compiled
  discovery task.
- No `tool_call` event arrives for the MCP tool itself.

## Expected Behavior

Allowlist on task type off `TYPE_MAP`, plus the worker case where an executed SIMPLE task's
`taskType` is the task's own name. Keep the existing `_fw_` reference skip.

**The allowlist needs a naming change alongside it.** The enrich script replaces
`inputParameters` wholesale for three tool kinds, dropping the `method` key that
`LLMHelper.java:407` had put there:

- HTTP: `JavaScriptBuilder.java:716` sets `{http_request: hrequest}`, and `hrequest.method`
  (`:709`) is the HTTP verb, not the tool name.
- agent-tool: `:740-747` sets `prompt`, `session_id` and `__today__` only.
- `PULL_WORKFLOW_MESSAGES`: `:789` sets `{batchSize}` only.

MCP keeps `method` explicitly (`:722`), and HUMAN and the media/RAG kinds keep it by merging
`tc.inputParameters`. So with the denylist removed, `resolveToolName` (`:439-461`) would return
`"http"` for an HTTP tool and `"sub_workflow"` for an agent tool via its `:455-456` branch. The
only key set for every dispatched tool is `_agent_tool_name` (`:797`), which `resolveToolName`
does not read. Resolution order should be `inputData._agent_tool_name`, then `inputData.method`,
then `taskDefName`.

## Additional Notes

Verified against `origin/main` @ `ce46eac33`.

**The allowlist alone does not fix HTTP or agent-as-tool, and that is a separate change.**
`onTaskCompleted` (`:99`) is only reachable from `TaskStatusListener.onTaskCompletedIfEnabled`,
whose sole production call site is `WorkflowExecutorOps.java:1115` inside `updateTask(TaskResult)`,
the worker and task-update path. Async system tasks complete through `AsyncSystemTaskExecutor`
(`:200`, `:245`) and never notify the listener. `HttpTask.isAsync()` returns `true`
(`http-task/.../HttpTask.java:229`) and so does `SubWorkflow.isAsync()`
(`core/.../tasks/SubWorkflow.java:328-330`). So removing those two from the denylist changes
nothing on its own; they need a completion notification for async system tasks. `CALL_MCP_TOOL`
and `HUMAN` are unblocked by the denylist fix alone.

The dispatch script and tool-spec compilation are each duplicated
(`JavaScriptBuilder.java:1430-1540`, `ToolCompiler.java:1597-1658`), so a fix likely applies
twice. Marker-based identification was tried and reverted in `09842ab16`, which is why this is
framed as a selection change rather than a new marker.

Four SDKs reconstruct tool calls from the polled task list and each has its own version of this
problem (filed against `python-sdk`, `javascript-sdk`, `java-sdk`, `csharp-sdk`). Getting
selection and naming right here is what makes a shared rule worth stating: a tool call is
identified by task type, and its name comes from `inputData._agent_tool_name` or `taskDefName`,
never from the task reference name.

ambiorix2099:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

## Summary

- `isToolTask` identified a tool call by exclusion: twelve task types denylisted, everything else a tool. It was wrong in both directions.
- Root cause: four denylist entries are real tool kinds, since `ToolCompiler.TYPE_MAP` compiles `mcp`, `human`, `http`/`api` and `agent_tool` onto `CALL_MCP_TOOL`, `HUMAN`, `HTTP` and `SUB_WORKFLOW`. The default branch then filtered nothing, because `getTaskDefinition().isPresent()` is true for any named task once `MetadataMapperService` synthesizes an ad-hoc `TaskDef`.
- Fix: allowlist on `_agent_tool_name`, the task types a declared tool compiles to, and the worker case. Naming reads `_agent_tool_name`, then `method`, then `taskDefName`, never the reference name, which carries provider-controlled ids. An unmarked `SUB_WORKFLOW` counts as a strategy handoff rather than an agent-as-tool call, the one type here meaning two things. Compile-time markers were tried and reverted in `09842ab16`, so this stays a selection change.

## User impact

A human tool ran and returned its answer without ever appearing as a tool call in the agent's event stream. One `human` tool named `ask_human`, same agent, against servers built from this branch and from its merge-base:

| | events for the turn |
|---|---|
| before | `thinking`, `waiting`, `thinking`, `done` |
| after | `thinking`, `waiting`, `tool_call` (`ask_human`), `tool_result` (`ask_human`), `thinking`, `done` |

Worker tools are unaffected: `get_weather` reports the same pair either side.

## Changes

- `ToolCompiler.java`: add `COMPILED_TOOL_TASK_TYPES`, derived from `TYPE_MAP` and the media types that fall back to their own upper-cased name, excluding `SIMPLE`. The name is qualified because `AgentChatCompleteTaskMapper` holds a private `TOOL_TASK_TYPES` that does include `SIMPLE`, for a different decision.
- `AgentEventListener.java`: swap the denylist for that allowlist plus any custom task type, `TaskType.of` reporting `USER_DEFINED`. That covers worker tools and tools whose config names their own type, and excludes every platform type, `LIST_MCP_TOOLS` included. Reorder `resolveToolName` and drop the reference-name fallback.
- `AgentEventListenerTest.java`: eight cases over MCP and HUMAN naming, the discovery task, agent-as-tool against a handoff, a config-named task type both with and without the dispatch marker, nine orchestration types, and `_fw_` wrappers.

Fixes #1584

### DIFF
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
@@ -170,7 +170,7 @@ public class ToolCompiler {
      * all. {@code SimpleTaskMapper} rewrites an executed SIMPLE task's type to the task's own name,
      * so a worker tool is recognised by that instead.
      */
-    public static final Set<String> TOOL_TASK_TYPES =
+    public static final Set<String> COMPILED_TOOL_TASK_TYPES =
             Stream.concat(
                             TYPE_MAP.values().stream(),
                             MEDIA_TOOL_TYPES.stream().map(t -> t.toUpperCase(Locale.ROOT)))
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
@@ -156,19 +156,8 @@ public class ToolCompiler {
                     Map.entry("pull_workflow_messages", "PULL_WORKFLOW_MESSAGES"));
 
     /**
-     * Conductor task types a declared tool compiles to, for consumers that have to tell a tool
-     * invocation apart from the rest of a compiled agent workflow.
-     *
-     * <p>Derived from {@link #TYPE_MAP} and from the media types that fall back to their own
-     * upper-cased name, so adding a tool kind to the map extends this set with it.
-     *
-     * <p>It is a floor rather than the closed set of what a tool can compile to. A media or RAG
-     * tool's own config may name its task type, which nothing static can enumerate; those tasks
-     * carry the dispatch script's {@code _agent_tool_name} input instead.
-     *
-     * <p>{@code SIMPLE} is deliberately excluded: a worker tool is not identifiable by task type at
-     * all. {@code SimpleTaskMapper} rewrites an executed SIMPLE task's type to the task's own name,
-     * so a worker tool is recognised by that instead.
+     * Task types a declared tool compiles to. Excludes SIMPLE, whose executed task carries the
+     * tool's own name as its type.
      */
     public static final Set<String> COMPILED_TOOL_TASK_TYPES =
             Stream.concat(
--- a/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
+++ b/agentspan/src/main/java/org/conductoross/conductor/ai/agentspan/runtime/compiler/ToolCompiler.java
@@ -18,9 +18,12 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.conductoross.conductor.ai.agentspan.runtime.util.JavaScriptBuilder;
 import org.conductoross.conductor.common.metadata.agent.GuardrailConfig;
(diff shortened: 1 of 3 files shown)


ambiorix2099:
MCP, HTTP and `SUB_WORKFLOW` tools still report nothing after this change: all three complete as async system tasks, and `onTaskCompleted` is only reachable from `WorkflowExecutorOps.updateTask(TaskResult)`, so `AsyncSystemTaskExecutor` never notifies the listener. Confirmed on a live MCP agent.

#1583 stays open on that gap, which sits in core rather than here and is not a follow-on to this change.

NicholasDCole (approved):

Issue #1528, opened 2026-08-12
CVEs in Conductor-OSS 3.32.0-rc.25


peter-mcclonski:
Thanks to the great work from the dev team, the CVE count has dropped markedly after #1490 got resolved!  However, the following issues remain:

```
Vulnerabilities found for image myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;: total - 12, critical - 1, high - 11, medium - 0, low - 0
Image: myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;
Package: libssh2
Version: 1.11.1-1+deb13u1
Path: N/A
Status: open
Severity: high
CVE: ['CVE-2026-58050']
Description: ['CVE-2026-58050'] - 
----------------------------------------
Image: myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;
Package: nginx
Version: 1.26.3-3+deb13u7
Path: N/A
Status: open
Severity: high
CVE: ['CVE-2026-60005']
Description: ['CVE-2026-60005'] - 
----------------------------------------
Image: myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;
Package: io.netty_netty-codec-http2
Version: 4.1.135.Final
Path: /app/libs/conductor-server.jar/netty-codec-http2-4.1.135.Final.jar
Status: fixed in 4.2.16.Final, 4.1.136.Final
Severity: high
CVE: ['CVE-2026-56819']
Description: ['CVE-2026-56819'] - 
----------------------------------------
Image: myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;
Package: org.elasticsearch_elasticsearch
Version: 7.17.29
Path: /app/libs/conductor-server.jar/elasticsearch-7.17.29.jar
Status: fixed in 9.2.2, 9.1.8, 8.19.8,...
Severity: high
CVE: ['CVE-2025-37731']
Description: ['CVE-2025-37731'] - Improper Authentication in Elasticsearch PKI realm can lead to user impersonati
----------------------------------------
Image: myprivateregistry.org/conductoross/conductor:3.32.0-rc.25;
Package: io.micrometer_micrometer-core
Version: 1.15.11
Path: /app/libs/conductor-server.jar/micrometer-core-1.15.11.jar
Status: fixed in 1.16.6, 1.15.12
Severity: high
CVE: ['CVE-2026-40983', 'CVE-2026-40984']
Description: ['CVE-2026-40983'] - 
[...]

PR #1613, opened 2026-09-09
fix(docker): move server image to eclipse-temurin:21-jre-alpine base (#1528)

nthmost-orkes:
The last batch of CVEs from Peter's 3.32.1 scan in conductor-oss/conductor#1528 are OS-level packages in the server image's `debian:stable-slim` final stage -- nothing a Java dependency bump can reach, and debian-slim can't patch them forward:

| Package | CVE | Note |
|---|---|---|
| `perl` | CVE-2026-13221 (**critical**), CVE-2026-57432 | pulled in transitively on debian-slim |
| `libssh2` | CVE-2026-58050 | curl dependency chain |
| `nginx 1.26.3` | CVE-2026-60005 | debian-slim caps nginx at 1.26.x |

## Change

Switch the final stage of both server Dockerfiles to `eclipse-temurin:21-jre-alpine`:

- **perl and libssh2 are gone entirely** (not in the alpine image), and alpine ships **nginx 1.30.x**.
- The temurin JRE is preinstalled, so the only package we `apk add` is `nginx`.
- The healthcheck switches from `curl` to busybox **`wget`** (built into alpine), which also removes curl's `libcurl`/`libssh2` chain from the image.

### Alpine-specific adjustments (all smoke-tested)

- nginx config now lands in `/etc/nginx/http.d/` (alpine's `include` dir) instead of debian's `conf.d/`, plus `mkdir -p /run/nginx` for the pid dir.
- healthcheck **and** nginx `proxy_pass` use `127.0.0.1` instead of `localhost`. Under musl, `localhost` resolves `::1` first; busybox wget has no happy-eyeballs fallback so it fails against the IPv4-bound server, and nginx would burn a failed `::1` upstream attempt (502-then-retry + error-log noise) on every proxied request. `127.0.0.1` avoids both. This file is used only by these two server images, so the blast radius is contained.

### Compose healthchecks

Because the alpine image no longer ships `curl`, the docker-compose files that healthcheck the server on `:8080/health` were still calling a binary that no longer exists inside the container — the check would never pass, and any stack that gates on `service_healthy` (e.g. `ui-next/docker-compose.integration.yml`) would hang and go red on an unrelated PR. Swapped all 12 server healthchecks to the same busybox `wget` form the Dockerfile `HEALTHCHECK` uses (`wget -q -O /dev/null http://127.0.0.1:8080/health`). The elasticsearch `:9200/_cluster/health` checks keep `curl` — they run in the unchanged `elastic`/`opensearch` containers. (Thanks @ambiorix2099 for catching this.)

One behavior note: busybox `wget` is **stricter** than the previous `curl -I -XGET` (which had no `-f`, so it exited `0` even on an HTTP 500 — effectively a TCP-connectivity check). `wget` exits non-zero on any 4xx/5xx. That's an improvement, but it means a container whose `/health` returns non-200 while the process is up now flips `unhealthy` where it previously stayed green. `/health` returns 200 when the server is up, so no change in practice.

## Verification

Smoke-tested against the built alpine image (arm64):

- ✅ nginx **1.30.4** installs, config parses from `http.d/`, binds `:5000`, workers start, serves static root
- ✅ busybox `wget` healthcheck works against a live listener (with `127.0.0.1`; confirmed `localhost` fails on musl)
- ✅ `proxy_pass` resolves IPv4-only — verified no `[::1]` upstream attempts in the nginx error log
- ✅ java 21 present; `perl`, `libssh2`, and `curl` all absent
- ✅ `sqlite-jdbc` 3.49.0.0 bundles `Linux-Musl/{aarch64,x86_64}` natives, so the zero-config SQLite default backend loads under musl
- ✅ `eclipse-temurin:21-jre-alpine` is multi-arch (linux/amd64 + linux/arm64/v8), so the multi-arch image build is unaffected

### Full end-to-end (now verified — full multi-stage build, arm64)

Built the real `Dockerfile` from source (`PREBUILT=false`: gradle JAR + pnpm UI) and ran it:

- ✅ **Server JAR boots under musl** — default SQLite backend: HikariPool connects, Flyway migrates, `SQLite 3.49`, "Started Conductor in 4.4s". No `UnsatisfiedLinkError` / native-load failures anywhere in the boot log. Also booted clean on **postgres** (Flyway on PostgreSQL 16.13).
- ✅ **`/health` → 200** `{"healthy":true}`; the busybox `wget` healthcheck (both the Dockerfile `HEALTHCHECK` and the compose `test:`) returns `exit=0`.
- ✅ **Compose gating works** — `docker-compose-ui-e2e.yaml` (postgres + server): server flips to `healthy` in ~10s, so `depends_on: condition: service_healthy` resolves. This is the exact path that would have hung on the old curl healthcheck.
- ✅ **End-to-end UI → `/api` proxy** — nginx serves the UI (`GET /` → 200) and proxies `/api`, `/actuator`, `/swagger-ui` to `127.0.0.1:8080` (all 200). nginx error log is empty — **no `[::1]` upstream attempts**, confirming the IPv4 `proxy_pass` fix.
- ✅ **Functional workflow round-trip** — registered + ran a workflow whose INLINE task evaluates graaljs; it `COMPLETED` with the expected output, so the GraalVM polyglot engine runs under musl too.

(`/actuator/health` returns 503 when ES is absent because its `elasticsearch` indicator is `DOWN` — same on debian; the healthchecks target `/health`, not `/actuator/health`, so they're unaffected.)

## Scope

The two server images (`Dockerfile` → `:latest`, `Dockerfile.next` → `:next`) plus the 12 docker-compose files whose server healthcheck had to follow curl → wget. The builder stage (`azul/zulu-openjdk-debian:21`) is unchanged — it's discarded and doesn't ship. The standalone UI image (`docker/ui/Dockerfile`) has its own nginx config and is out of scope.

**User impact:** Operators scanning the Conductor server image will see the remaining OS-level CVEs — including the one critical (`perl`) — disappear, and the image gets smaller since the JRE no longer installs on top of a general-purpose base. No change to how the server is run or configured.

Fixes the last of the tractable items in conductor-oss/conductor#1528.


docker/server/Dockerfile
```diff
@@ -79,12 +83,14 @@ COPY --from=builder /conductor/server/build/libs/*boot*.jar /app/libs/conductor-
...
-COPY docker/server/nginx/nginx.conf  /etc/nginx/conf.d/default.conf
+COPY docker/server/nginx/nginx.conf  /etc/nginx/http.d/default.conf
 
 # Copy the files for the server into the app folders
 RUN chmod +x /app/startup.sh
 
-HEALTHCHECK --interval=60s --timeout=30s --retries=10 CMD curl -I -XGET http://localhost:8080/health || exit 1
+# Uses 127.0.0.1 (not localhost): busybox wget resolves localhost to ::1 first and does not fall
+# back to IPv4 the way curl did, so localhost would fail against the IPv4-bound server.
+HEALTHCHECK --interval=60s --timeout=30s --retries=10 CMD wget -q -O /dev/null http://127.0.0.1:8080/health || exit 1
```
ambiorix2099, line 93: This drops curl (ran the base image to check: busybox wget only), but the compose files still healthcheck the server with it.
Nothing breaks today, though ui-next/docker-compose.integration.yml gates app on the server being healthy, and its workflow only runs on UI changes.
So the next UI PR gets a hang and a red build it didn't cause.

Can we swap curl for wget in the 12 compose files as part of this?

test: ["CMD", "wget", "-q", "-O", "/dev/null", "http://127.0.0.1:8080/health"]

No concerns with the base image change itself.

ambiorix2099 (dismissed):
No concerns with the base image change itself, but commented on a potential issue.

docker/server/Dockerfile
nthmost-orkes: ooh good catch!

nthmost-orkes: Done in 140775a. Swapped curl → busybox wget in all 12 tracked compose files that healthcheck the server on `:8080/health`, using the exact form the Dockerfile HEALTHCHECK already uses:

```
test: ["CMD", "wget", "-q", "-O", "/dev/null", "http://127.0.0.1:8080/health"]
```

Note I used `127.0.0.1` rather than `localhost` — same musl reason as the Dockerfile: busybox wget resolves `localhost` to `::1` first with no IPv4 fallback, so `localhost` would fail against the IPv4-bound server inside the alpine container.

Left the elasticsearch `:9200/_cluster/health` checks on curl — those run in the unchanged `elastic`/`opensearch` containers, which still have curl.

docker/docker-compose-postgres-redis.yaml
```diff
@@ -0,0 +1,86 @@
+version: '2.3'
+
+# Hybrid stack: Postgres for metadata/execution/payload/indexing, Redis (standalone)
+# for the workflow queue. Modeled on docker-compose-postgres.yaml; adds a redis
+# service borrowed from docker-compose.yaml.
+
+services:
+  conductor-server:
+    environment:
+      - CONFIG_PROP=config-postgres-redis.properties
```
ambiorix2099: Might be missing context here...

Nothing references this file yet, and `config-postgres-redis.properties` isn't in `docker/server/config/`, where every other compose file's `CONFIG_PROP` resolves.

As it stands, the server logs `Ignoring ... since it does not exist` and comes up on healthy SQLite, instead of postgres and redis.

Can we pull this whole YAML file out of this PR and land it separately along with the properties file, or do you need it here for something?

nthmost-orkes: oh - yeah this was an accidental commit with git add -A

I will disappear this now.  thanks!

nthmost-orkes: Done — removed in 60b45a16b (pushed). The file's now gone from the PR diff entirely. It'll come back in its own PR alongside the tracked `config-postgres-redis.properties`.

c4lm:
@nthmost-orkes if we use alpine please install gcompat and add LD_PRELOAD string, I have hit this more than once in our proprietary repos - google does not test GRPC stuff on alpine and generally does not recommend running it, so suddenly you can get your pod to crash with a segfault on startup when you have the right dependencies and arch, and when you are not expecting it because it's JVM
e.g. [most recent](https://github.com/orkes-io/orkes-assistant/pull/51)

nthmost-orkes:
Good catch, thank you — done in b1356360f.

Confirmed we're exposed: the server JAR bundles `conductor-grpc-server` (`server/build.gradle` → `implementation project(':conductor-grpc-server')`), so grpc-java + netty native are on the classpath regardless of whether the gRPC server is enabled.

Installed `gcompat` + `libunwind` in both server Dockerfiles (`Dockerfile` and `Dockerfile.next`), following your orkes-assistant#51.

One difference from that PR: orkes-assistant sets `LD_PRELOAD` in the `ENTRYPOINT` because it runs `java` directly. Our image runs `startup.sh`, which starts **nginx first, then java** — so a global `ENV LD_PRELOAD` would also preload the shims into nginx and every busybox call (including the `wget` healthcheck). Instead I `export LD_PRELOAD=/lib/libgcompat.so.0:/usr/lib/libunwind.so.8` inside `startup.sh` right before the `java` line (after nginx has started), so only the JVM gets it and nginx/busybox stay on native musl.

Smoke-tested on `eclipse-temurin:21-jre-alpine` (arm64) with the real server jar and `conductor.grpc-server.enabled=true`:
- shim paths resolve: `/lib/libgcompat.so.0`, `/usr/lib/libunwind.so.8`
- `grpc: Server started, listening on 8090` + `Started Conductor in 8.3s`, no SIGSEGV / `UnsatisfiedLinkError` — no regression from the preload

(Couldn't force the segfault to reproduce on arm64 here, which lines up with your "right arch/deps" caveat — this lands as the defensive insurance you flagged, plus the "DO NOT REMOVE WHILE WE ARE RUNNING ALPINE" comments so it survives future cleanup.)

c4lm:
yeah, unless you actually deploy, you won't know for sure if it is happening, but it's better to guard proactively (in case if any grpc dependency is bumped or if it is unpinned and changes when transitive in the future)
when we hit it in orkes-conductor the first time ([here](https://github.com/orkes-io/orkes-conductor/pull/2172#issuecomment-2471855320) it was still on alpine) we could deploy to arm but not to amd64 because of this

c4lm (approved):

nthmost-orkes:
> yeah, unless you actually deploy, you won't know for sure if it is happening, but it's better to guard proactively (in case if any grpc dependency is bumped or if it is unpinned and changes when transitive in the future) when we hit it in orkes-conductor the first time ([here](https://github.com/orkes-io/orkes-conductor/pull/2172#issuecomment-2471855320) it was still on alpine) we could deploy to arm but not to amd64 because of this

thanks, i'll make sure the new image is test deployed on different architectures!

PR #1622, opened 2026-09-11
Fix #1615: re-run start() for redelivered SCHEDULED SUB_WORKFLOW instead of forcing TIMED_OUT

### LINKED ISSUE #1615: DYNAMIC_FORK: one SUB_WORKFLOW branch never created, force-TIMED_OUT after 60 min (regression in 3.32.0 from #1369)
## Report (from a customer in the Conductor OSS community Slack)

> Hi guys, we have DYNAMIC_FORK Task, which usually runs 80 SUB_WORKFLOW Tasks, but today conductor server haven't created sub-workflow for one of them and changed task and worfklow status to Timeout after 60 mins. What has happend? Tell me if you need more information about this incedent. We use the last version 3.32.0 of conductor server and we have never faced with this in previous one 3.30.2

## Cause

Regression in 3.32.0 from #1369 ("reserve an async system task's queue message while it runs", fixing #1321). The new redelivery logic in `AsyncSystemTaskExecutor` assumes a blocking `start()` never leaves a task `SCHEDULED` (comment at `AsyncSystemTaskExecutor.java:159`), so on redelivery it force-times-out a still-`SCHEDULED` task instead of re-running it (line 164).

That assumption is false for `SUB_WORKFLOW`: `SubWorkflow.start()` deliberately leaves the task `SCHEDULED` on a transient error (`SubWorkflow.java:206-207`). Combined with the new 1-hour message reserve (default `responseTimeoutSeconds` = `TaskDef.ONE_HOUR` = 3600s), a transient hiccup during child creation now becomes an unrecoverable 60-minute timeout instead of a fast, self-healing retry.

Sequence for the one failed branch out of 80:
1. Task picked up; `startTime` persisted, message reserved for 1 hour.
2. Child creation hits a transient error (plausible when 80 children are minted at once) or the worker restarts mid-launch -- task stays `SCHEDULED`.
3. Message stays invisible for the full hour (on 3.30.2 it would redeliver in seconds and retry).
4. On redelivery, `hasExceededResponseTimeout` is true, so the task is set `TIMED_OUT` instead of re-running `start()` -- the child is never created and the JOIN never completes, so the workflow times out.

This matches every detail: 1 of 80, sub-workflow never created, `TIMED_OUT`, and exactly 60 minutes (`ONE_HOUR` doubles as both the reserve window and the timeout threshold). Not fixed in 3.32.1-3.32.3.

## Confirming from logs

Look for `Timing out SUB_WORKFLOW/... no response within responseTimeout` (emitted only by this path), a node restart/GC/load spike ~60 min before the timeout, and a ~1-hour gap between the task's `startTime` and `endTime`.

## Proposed direction (for discussion)

Keep the #1321 reserve, but on redelivery of a still-`SCHEDULED` `SUB_WORKFLOW` re-run `start()` (it's idempotent via the deterministic child id) rather than blanket-timing-out. Flagging for maintainer input first, since the reserve was added deliberately.

## Environment

Conductor server 3.32.0 (regression from 3.30.2, still present through 3.32.3); `DYNAMIC_FORK` fanning out to ~80 `SUB_WORKFLOW` tasks.

nthmost-orkes:
Fixes #1615.

The conservative fix: add `WorkflowSystemTask.isStartRetriable()` (default false, true for `SubWorkflow`). On redelivery of a still-SCHEDULED task past its responseTimeout, the executor only force-times-out tasks that are NOT start-retriable. Retriable ones re-run `start()` instead, which is safe because `SubWorkflow.start()` is idempotent (deterministic child id plus the child-id lock in `startWorkflowIdempotent`).

This keeps the #1321 reserve and overrun-timeout for genuinely blocking tasks, and just stops force-timing-out the one case the issue describes (a SUB_WORKFLOW left SCHEDULED by a transient error or a worker restart mid-launch).

Not addressed here: recovery still waits out the reserve window before retrying (up to responseTimeout), instead of the ~60s it took pre-3.30.2. 

Tests: added a spec that a redelivered SCHEDULED start-retriable task past responseTimeout is re-run (not TIMED_OUT).


core/src/main/java/com/netflix/conductor/core/execution/AsyncSystemTaskExecutor.java
```diff
@@ -155,12 +155,13 @@ public void execute(WorkflowSystemTask systemTask, String taskId) {
...
             if (scheduled || task.getStatus() == TaskModel.Status.IN_PROGRESS) {
-                if (scheduled && hasExceededResponseTimeout(task)) {
-                    // A blocking start() never leaves SCHEDULED, so a redelivered SCHEDULED task
-                    // past responseTimeout means its run overran: time it out, don't re-run it
-                    // (#1321). IN_PROGRESS response-timeouts are
-                    // DeciderService.isResponseTimedOut's
-                    // job (it budgets responseTimeout + callbackAfterSeconds).
+                if (scheduled
+                        && hasExceededResponseTimeout(task)
+                        && !systemTask.isStartRetriable()) {
```
manan164, line 160: Why do we need to change the WorkflowSystemTask? Cant we add a method here just to filter the subworkflow task and make the responseTimeout agressive?

nthmost-orkes: OK, so couple of immediate responses here -- and I realize this file is a core one, so i understand hesitation to change it.

First off, the change I'm proposing follows the same pattern of overridable boolean predicates — isAsync(), isAsyncComplete(), isTaskRetrievalRequired() -- and also follows @mp-orkes 's addition to fix a JOIN latency in 2025 ( bf90790c7 ).

The second is -- i'm not sure that just making the responseTimeout aggressive actually fixes the bug?

And finally, "just add a method" seems like it means more additive changes, as opposed to a change that follows an existing pattern.

If i've misinterpreted your questions, let me know, happy to keep talking.

nthmost-orkes:
Closing in favor of #1655

PR #1630, opened 2026-09-15
[CCOR-13375]fix:agent guide scroll stuck

najeebkp:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----
[CCOR-13375](https://orkes.atlassian.net/browse/CCOR-13375)Create Agent page: scroll stops working after changing Framework dropdown, until the page is clicked once

## Demo
Screen Recording 2026-09-15 at 6 17 15 PM


ui-next/src/pages/agent/CreateAgentGuide.tsx
```diff
@@ -185,6 +185,22 @@ export default function CreateAgentGuide() {
...
+                    slotProps: {
+                      root: {
+                        // Let wheel/trackpad events reach the page under the menu
+                        sx: { pointerEvents: "none" },
+                      },
+                      paper: {
+                        sx: { pointerEvents: "auto" },
+                      },
+                    },
+                  }}
```
ajlai-orkes, line 195: wasn't the issue that after selecting an item the scroll is still locked? i think it makes sense to keep scroll disabled if the select menu is open. can we debug what's causing the scroll lock behavior even after selecting a menu item instead of just disabling scroll lock completely?

najeebkp: Good call — debugged it, and you're right that disabling scroll lock was wrong.
The real defect is that MUI leaves the **Modal root in the DOM** after close. It spans the viewport, so with default pointer-events it swallows wheel events (a captured wheel's target was `MuiMenu-root`) and blocks clicks.
Narrowed the fix to making just that root transparent to pointer input, paper stays `auto`.

ajlai-orkes: but shouldn't there be proper cleanup of the modal root after close? what's causing it to be left behind?

najeebkp: okey, so the root cause is -   onChange navigates via setSearchParams, and that re-render interrupts the Menu's exit transition. onExited never fires, so MUI's exited stays false, and since Modal returns null only when !keepMounted && !open && exited, it keeps rendering the root across the viewport.

and the cleanup I did now is -  Navigating from onExited lets the close finish first, so the Modal root is removed. Verified, and asserted by a test.

ajlai-orkes (approved):

PR #1631, opened 2026-09-15
fix(ai): compare recorded tool-call and tool-result numbers by value

ling-senpeng13:
Playback matches the normalized request against the recording, and Jackson number nodes compare by type as well as value. Python serializes an integral float as `15000.0` (DoubleNode); Go and JavaScript send the same value as `15000` (IntNode). A recording made with one SDK therefore never matches a request from another once a tool result or tool-call argument holds such a number, and the run fails with "no recording".

This canonicalizes numbers in tool-call arguments, tool results and the rewritten tool-result summary to the node Jackson would parse for the value: int, long or big-integer for integral values, double otherwise. Numbers that already have that form are unchanged, so existing recordings keep matching.

Found while replaying the Python `09_human_in_the_loop` recording from this branch against the Go SDK port: `check_balance` returns `15000.00`, the recording holds `15000.0`, Go sends `15000`. With this change that replay passes.

Tests: `RecordedRequestNormalizerTest` gains `toolResultNumbersMatchByValueAcrossSdks` and `toolCallArgumentNumbersMatchByValue`; the existing recorder and mock tests still pass.


ling-senpeng13:
Superseded by #1633, which carries this commit together with the agent-tool envelope fix.

PR #1632, opened 2026-09-15
fix(ai): replay agent tool results across sub-workflow runs

ling-senpeng13:
An agent used as a tool runs as a sub-workflow. Its structured tool result holds only `result`, but the generated `[TOOL RESULTS]` text carries the sub-workflow envelope (`subWorkflowId`, `finishReason`, `context`, `rejectionReason`) and names the entry by the task reference (`<uuid>_0_0_`) rather than the tool. The sub-workflow id is new on every run, so the parent's next request never matches a recording and playback fails with "No recorded response matches the LLM request".

This reduces an envelope-shaped summary output to its `result` before matching it against structured history, so the summary is rewritten the same way as any other tool result, and accepts generated result names with more than one index (the fork appends one to the call's reference; an agent tool carries a second). Only an object whose other keys are all envelope keys is reduced; any other shape is left unchanged.

Found with the Go SDK integration test `TestSkillAsAgentTool` (a parent agent calling a skill agent as a tool). With this change it replays; the rest of the suite is unchanged.

Tests: `RecordedRequestNormalizerTest.agentToolSummaryPlaysBackAcrossSubWorkflowIds`; existing recorder and mock tests pass. Independent of #1631 (numeric equality); the two can be cherry-picked in either order.


ling-senpeng13:
Superseded by #1633, which carries this fix together with the numeric-equality fix.

PR #1636, opened 2026-09-16
Return SQL batch polls as soon as one task is available (#142)

### LINKED ISSUE #142: Latency tradeoff could be better when polling with a long timeout and count > 1 in MySQL and Postgres persistences
**Describe the bug**
If all workers poll for tasks with a count=10 for example, and a timeout=5s, very few things get polled during low activities.

This is not really a bug, but maybe a tradeoff to clarify / change.

**Details**
Conductor version: 3.19.0

**To Reproduce**
1. create a workflow with 1 single task
2. start a worker polling for that task with count=2 and timeout=5s
3. execute the workflow
4. check how long the task stays in the queue
5. Repeat every 5 seconds

What happens is that the DAO implementations try at least "count" messages from the queue while the timeout is not elapsed.

MySQL: https://github.com/conductor-oss/conductor/blob/55268f0633969379874cc425ef32c9048bbddbfa/mysql-persistence/src/main/java/com/netflix/conductor/mysql/dao/MySQLQueueDAO.java#L349

Postgres: https://github.com/conductor-oss/conductor/blob/55268f0633969379874cc425ef32c9048bbddbfa/postgres-persistence/src/main/java/com/netflix/conductor/postgres/dao/PostgresQueueDAO.java#L169

Dyno: https://github.com/conductor-oss/conductor/blob/main/redis-persistence/src/main/java/com/netflix/conductor/redis/dao/DynoQueueDAO.java#L96, which relies on https://github.com/Netflix/dyno-queues/blob/dev/dyno-queues-redis/src/main/java/com/netflix/dyno/queues/redis/RedisDynoQueue.java#L343

**Expected behavior**
In case of low activity, the pop method should wait less time, and return a task anyways instead of trying to maximise the number of tasks polled.

Why?

I think we have multiple cases, depending on the number of tasks to poll:

- high activity: very high number of tasks in the queue, so long polling is barely relevant, and it's easy to poll for `count` number of tasks. Count is important to achieve high throughput.
- very low activity: tasks arrive in the queue at a slow rate, and so long polling is relevant to improve latency, but waiting for `count` tasks to be there is waiting for nothing. Long polling is more relevant than count in this case, and a task should be return earlier.
- medium activity: tasks arrive in the queue so that maybe 50% of workers get their count filled, and 50% of them get less. In this case, we get some tasks being processed fast, while others wait much longer. With medium activity, a slower but snappier flow would be better, and thus count should matter a bit less than long polling.

As a consequence, I think count should not be a value that should be that hard on the polling, but its impact is more important when long polling makes less sense.

I guess there are multiple ways to fix it:

- pop until `count` is 1 or more OR until the time has elapsed
- pop until `count` is at least half of the requested `count` OR until the time has elapsed.
- maybe depending on the value of pop, change the strategy: if pop is like 1000, then keep the current behaviour, as the server always expects high load and throughput is preferred. If count is low (like 3), then does it as a best effort with returning up to 3 within the time period (just like 1).

What do you think?

nthmost-orkes:
Fixes #142.

## What was happening

`MySQLQueueDAO`, `PostgresQueueDAO`, and `SqliteQueueDAO` waited for the **full requested `count`** of tasks (or the full timeout) before returning a batch poll. So under low activity, a worker polling with `count > 1` and a long timeout would block for the entire timeout even when tasks were already sitting in the queue ready to go. Poll timeout is capped at 5s, so that was up to 5s of added latency per poll just to hand back one task.

The Redis queue (`QueueMonitor.pop`) already does the right thing: it drains whatever is available up to `count`, and only blocks when the queue is empty. The three SQL backends were the outliers, and they also contradicted the documented long-poll contract ("waits until timeout or at least 1 task is available").

## The change

All three DAOs now return as soon as **at least one** message is available (up to `count`), otherwise they wait out the timeout:

- **MySQL** `popMessages`: loop guard `messages.size() < count` -> `messages.isEmpty()`
- **Postgres** `pollMessages`: return guard `messages.size() >= count` -> `!messages.isEmpty()`
- **SQLite** `pollMessages`: return guard `messages.size() >= count` -> `!messages.isEmpty()`

This lines the SQL backends up with the Redis behavior and the documented contract.

## Tradeoff

Under bursty load, a batch worker that previously waited to collect a full batch may now get a smaller batch (whatever landed in the first peek), in exchange for much lower tail latency. Redis already accepts exactly this tradeoff, so this brings all backends into agreement rather than introducing a new one.

## Tests

Added a regression test to each DAO test suite (`pollMessagesReturnsPromptlyWhenFewerThanCountAvailable`): push one message, poll with `count=5` and a 10s timeout, assert it returns promptly with the single available message instead of blocking. MySQL and Postgres run against testcontainers; SQLite runs against its embedded file DB. All three pass, and the full `MySQLQueueDAOTest`, `PostgresQueueDAOTest`, and `SqliteQueueDAOTest` suites stay green.

**User impact:** A worker batch-polling a quiet queue used to stall for up to 5 seconds before picking up a task that was already waiting. Now it grabs the task right away.

## Orkes parity

No new classes or layout decisions -- this is a small behavior fix to existing OSS DAOs. The fix intentionally aligns the SQL backends to the existing Redis/Orkes-MQ `QueueMonitor.pop` semantics (drain up to count, block only when empty).


mysql-persistence/src/main/java/com/netflix/conductor/mysql/dao/MySQLQueueDAO.java
```diff
@@ -379,7 +379,10 @@ private List<Message> popMessages(
         long start = System.currentTimeMillis();
         List<Message> messages = peekMessages(connection, queueName, count);
 
-        while (messages.size() < count && ((System.currentTimeMillis() - start) < timeout)) {
+        // Long-poll semantics: return as soon as at least one message is available (up to
+        // count), rather than blocking for the full timeout waiting to fill the whole batch.
+        // This matches the Redis queue behavior and keeps tail latency low under low activity.
+        while (messages.isEmpty() && ((System.currentTimeMillis() - start) < timeout)) {
```
ambiorix2099: A poll can now come back empty within a few ms of starting, with its whole timeout unspent. The worker treats that as "nothing queued" and polls again immediately, so with N workers and few tasks you get N-1 of them hitting the DB at round-trip speed instead of long-polling.

It happens when another consumer claims the rows between this poll's peek and its pop: the loop has already exited on the peek, and the pop only keeps rows whose UPDATE matched on `popped = false`, so there's nothing left to return.

I verified by testing locally with the MySQL configs, two consumers against a single message, and count=5, timeout=2s: both came back in ~4ms, one of them empty. On `main` both waited the full 2s, so the empty result isn't new, but the immediate re-poll is.

**Recommendation**: Loop while poppedMessages is empty rather than while the peek is empty.

chrishagglund-ship-it (approved):
seems good overall to me. one thing that was identified: with count=0 the new guard no longer short-circuits, so the existing zero-poll assertions (PostgresQueueDAOTest 197 and 265, SqliteQueueDAOTest 186) now wait out the full 10s timeout instead of returning instantly — they still pass, just ~30s slower. i think that only really happens in testing, but may be worth short-circuiting a fast return when the requested count is zero.

nthmost-orkes:
> seems good overall to me. one thing that was identified: with count=0 the new guard no longer short-circuits, so the existing zero-poll assertions (PostgresQueueDAOTest 197 and 265, SqliteQueueDAOTest 186) now wait out the full 10s timeout instead of returning instantly — they still pass, just ~30s slower. i think that only really happens in testing, but may be worth short-circuiting a fast return when the requested count is zero.

Yeah noted, and that's a good idea - i'll see if that's viable w/o issue before i pull this in.  THanks!

(Edit:  yep, you called it @chrishagglund-ship-it -- updated accordingly.)

PR #1644, opened 2026-09-17
Add opt-in SSRF guard rails to HTTP system task

### LINKED ISSUE #1018: Security: SSRF protection needed in HTTP task
## Summary

The `HTTP_TASK` system task passes user-controlled URIs directly to `restTemplate.exchange()` with no restrictions on what hosts can be reached. This enables Server-Side Request Forgery (SSRF): anyone who can define or modify a workflow can instruct the Conductor server to make HTTP requests to internal infrastructure.

**CodeQL Alert #2 · `http-task/.../HttpTask.java:175` · severity: error**

## The Problem

```java
// HttpTask.java:174-178
ResponseEntity<String> responseEntity =
        restTemplate.exchange(
                input.getUri(),   // ← fully user-controlled, no validation
                HttpMethod.valueOf(input.getMethod()),
                request,
                String.class);
```

There is no validation of `input.getUri()` before the call. An attacker with access to workflow definition APIs can target:

- **Cloud metadata endpoints** — `http://169.254.169.254/latest/meta-data/` (AWS IMDSv1, GCP, Azure) to steal instance credentials
- **RFC 1918 internal services** — `http://10.x.x.x/`, `http://192.168.x.x/`, `http://172.16-31.x.x/`
- **Loopback** — `http://localhost:8080/api/admin/...` (Conductor's own admin API)
- **Internal DNS names** — `http://internal-db-host:5432/`

## Why This Is By-Design — and Still a Problem

The HTTP task is *intentionally* designed to call external endpoints — that's its entire purpose. The issue is the absence of any configurable guard rails. Most production deployments need to reach external services, but should not allow the Conductor server itself to become a proxy into its own internal network.

## Proposed Solution

Add a configurable `HttpTaskProperties` class (or extend existing config) with:

1. **Block-by-default list**: RFC 1918 ranges, loopback (`127.0.0.0/8`), link-local (`169.254.0.0/16`), and `::1`
2. **Allowlist** (`conductor.system-task.http.allowed-hosts`): explicit patterns that override the block list for legitimate internal calls
3. **Denylist** (`conductor.system-task.http.denied-hosts`): additional patterns to block beyond defaults
4. URI validation at the start of `httpCall()`, before `restTemplate.exchange()` is called

Example config:
```yaml
conductor:
  system-task:
    http:
      deny-internal-network: true   # default: true
      allowed-hosts:
        - "api.trusted-partner.com"
      denied-hosts:
        - "*.internal.corp"
```

## Threat Model Note

Risk is higher in multi-tenant or externally-accessible Conductor deployments where untrusted parties can submit workflow definitions. In fully internal, trusted-user-only deployments the practical risk is lower, but the fix is still correct defense-in-depth.

## References

- [OWASP SSRF Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
- CodeQL alert: https://github.com/conductor-oss/conductor/security/code-scanning/2
- Epic: #1010

nthmost-orkes:
## What

Adds configurable host/IP block lists to the `HTTP` system task so operators can prevent the Conductor server from being used as an SSRF proxy into internal infrastructure (cloud metadata endpoints, RFC 1918 ranges, loopback, internal DNS names).

Fixes conductor-oss/conductor#1018.

A `ClientHttpRequestInterceptor` (`RestTemplateInterceptor`) resolves the target host and returns `403 FORBIDDEN` — before the request leaves the server — when the host or its resolved IP matches a configured block rule. Resolution flows through `HostAndIPChecker`:

1. If the host matches a **host** block rule (and is not explicitly allowed) → blocked.
2. If the host is explicitly allowed → allowed.
3. Otherwise resolve to an IP and check the **IP** block/allow rules.

## Configuration (opt-in)

```yaml
conductor:
  worker:
    http:
      block:
        ips:                        # CIDR or literal IPs to block
          - "169.254.0.0/16"        # link-local / cloud metadata
          - "127.0.0.0/8"           # loopback
          - "10.0.0.0/8"            # RFC 1918
        hosts:                      # host / single-level *.wildcard to block
          - "*.internal.corp"
        allowed-ips:                # override the block lists
          - "10.1.1.1"
        allowed-hosts:
          - "api.internal.corp"
```

All four lists default to empty. **With no configuration nothing is blocked**, so existing deployments — including those that intentionally call internal hosts, which is a by-design use of the HTTP task — are unaffected on upgrade. Blocking is a deployment-level policy an operator turns on.

## Design decision: opt-in vs block-by-default

Issue #1018 proposed *block-by-default* (RFC 1918 / loopback / link-local blocked out of the box). This PR ships **opt-in** instead, matching enterprise Conductor's existing model. Block-by-default would break any deployment that legitimately uses the HTTP task to reach internal services on upgrade; the sample config above documents the recommended hardening ranges an operator can paste in to get the same protection deliberately.

## Orkes parity

Source-of-truth for layout, class granularity, config prefix, and field names (namespace translated `io.orkes.conductor.*` → `org.conductoross.conductor.*`):

- `workers/.../http/config/HttpWorkerBlockConfig.java` — config prefix `conductor.worker.http.block` and the `ips` / `hosts` / `allowedIps` / `allowedHosts` fields
- `workers/.../http/providers/interceptors/RestTemplateInterceptor.java` — the RestTemplate-based interceptor (Orkes' RestTemplate variant, which matches OSS's RestTemplate-based `HttpTask`)
- `.../interceptors/HostAndIPChecker.java`, `HostChecker(Impl)`, `IPChecker(Impl)`, `IPCheckResult`, `IpAddressMatcher` — same class granularity and matching logic

## OSS-side additions

- None functionally new. `IpAddressMatcher` is Spring Security's matcher (same as Orkes) vendored to avoid pulling in the full spring-security dependency for one class; `IPCheckResult` uses static factories instead of Lombok `@Builder` to match this module's plain-Java (no Lombok) convention. Names, fields, and behavior are unchanged.

## Testing

New unit tests (16, all passing), no Docker/Testcontainers required:

- `HostCheckerImplTest` — exact match, case-insensitivity, single-level wildcard scope, allow-overrides-block
- `IPCheckerImplTest` — CIDR ranges, loopback/RFC 1918, allow-overrides-block
- `HostAndIPCheckerTest` — precedence, literal-IP metadata/loopback blocking, empty-config passthrough
- `RestTemplateInterceptorTest` — 403 on blocked host (execution never runs), passthrough on allowed host

Existing `DefaultRestTemplateProviderTest` / `HttpTaskUnitTest` still pass; the provider keeps its original 2-arg constructor and gains an autowired constructor that attaches the interceptor when present.

**User impact:** Operators can now stop the HTTP task from reaching internal-only addresses (like `169.254.169.254` or loopback) by listing them under `conductor.worker.http.block.*`. Nothing changes for anyone who doesn't configure it.


c4lm (commented):
implementation seems to match that of orkes-conductor except it does not check redirects

nthmost-orkes:
@c4lm -- good catch... The `RestTemplateInterceptor` only fires on the initial request; the Apache client follows 3xx redirects transparently, so an allowed host could 302 to `169.254.169.254` (or any blocked target) and slip right past the block list.

Fixed in 7ac13f05e by mirroring Orkes' redirect handling: a new `BlockingRedirectStrategy` re-runs `HostAndIPChecker` on every redirect target and aborts the exchange when it's blocked. It's wired into `DefaultRestTemplateProvider` only when the interceptor is present, so nothing changes for deployments with no block rules configured.

One deliberate difference from the Orkes source: instead of implementing `RedirectStrategy` and matching only 301/302, this extends `DefaultRedirectStrategy` so all redirect codes it recognizes (301/302/303/307/308) are re-checked -- a 307 to an internal host would otherwise have been another way through. Added `BlockingRedirectStrategyTest` covering the blocked-redirect rejection, allowed-redirect passthrough, and all the redirect status codes.

Issue #1010, opened 2026-04-14
Epic: Critical Security Updates


nthmost-orkes:
## Overview

Tracking all active security vulnerabilities flagged by CodeQL static analysis and CVE databases. Each row is a linked sub-issue.

> **Status as of 2026-09-17:** Alert #5 (path injection in `MockExternalPayloadStorage`) is **fixed** — PR #723 merged to `main` on 2026-07-08 with a canonical-path check. Its CodeQL alert still displays "open" because the code-scanning workflow hasn't re-run over that path (last scan 2024-05-12); it will auto-close on the next scan. Everything else remains open, and two new GitHub Actions permission alerts (#31, #32) have appeared since the last update.

## Code vulnerabilities (CodeQL)

| Alert | Severity | File | Issue | Status |
|---|---|---|---|---|
| #2 SSRF | 🔴 critical | `http-task/.../HttpTask.java:175` | #1018 | ⏳ Open — `restTemplate.exchange(input.getUri(), …)` with no host restriction; needs allowlist/denylist |
| #6 ReDoS | 🔴 high | `core/.../ParametersUtils.java:56,258` | #1638 | ⏳ Open — backtracking `PATTERN` run on every task param; reachable from worker output |
| #4 Path injection | 🔴 high | `core/.../DummyPayloadStorage.java:95` | #1639 | ⏳ Open — `new File(payloadDir, path)` unvalidated; dev-mode only |
| #3 ReDoS | 🔴 high | `postgres-persistence/.../PostgresIndexQueryBuilder.java:69` | #1640 | ⏳ Open — O(n²) regex + `Pattern` compiled per-call; low real risk |
| #23 Identity replacement | 🟡 medium | `ui-next/.../errorInspector/state/helpers.ts:194` | #1642 | ⏳ Open — no-op `String.replace`; intended transform not applied |
| #5 Path injection | 🔴 high | `test-harness/.../MockExternalPayloadStorage.java:96` | PR #723 | ✅ **Fixed & merged** 2026-07-08 (CodeQL alert stale, will auto-close) |

## GitHub Actions — missing workflow permissions

Tracked in **#1641** · rule `actions/missing-workflow-permissions` · 🟡 medium

| Alert | File | Line |
|---|---|---|
| #14 | `.github/workflows/ci.yml` | 314 |
| #17 | `.github/workflows/ci.yml` | 39 |
| #29 | `.github/workflows/ci.yml` | 134 |
[...]

PR #1648, opened 2026-09-18
fix(security): hoist Pattern and prevent ReDoS in IndexQueryBuilder (Fixes #1640)

### LINKED ISSUE #1640: Security: ReDoS + per-call Pattern compile in PostgresIndexQueryBuilder (CodeQL alert #3)
**CodeQL alert #3** · rule `java/polynomial-redos` · severity **high** (low real risk; borderline)
\`postgres-persistence/src/main/java/com/netflix/conductor/postgres/util/PostgresIndexQueryBuilder.java\`

Sub-issue of #1010.

## Problem
The condition regex \`([a-zA-Z]+)\s?(=|>|<|IN)\s?(.*)\` can backtrack to O(n²) on long all-alpha inputs with no operator. Query strings are short in practice, so real risk is low.

Secondary correctness issue: \`Pattern.compile(CONDITION_REGEX)\` runs **inside the method** (line 71) on every call instead of being a \`static final\` field.

## Fix direction
Hoist the compiled \`Pattern\` to a \`private static final\` field, and tighten the regex so the leading token and trailing value can't backtrack against each other.

jhaabhijeet864:
### Summary
Fixes #1640 (Sub-issue of #1010).

Addresses CodeQL alert #3 (`java/polynomial-redos`):
1. Hoists `CONDITION_PATTERN` to a `private static final Pattern` to eliminate per-call regex compilation overhead.
2. Tightens the condition pattern with a possessive quantifier `^([a-zA-Z]++)\s*(=|>|<|IN)\s*(.*)$` to eliminate catastrophic backtracking on non-matching inputs.
3. Applied the same hardening to `SqliteIndexQueryBuilder`.
4. Added unit tests for ReDoS resilience and whitespace parsing.


nthmost-orkes (approved):
Approving. I reproduced the vulnerability before signing off rather than taking the CodeQL label at face value, and it holds up under measurement.

I timed the old vs. new pattern against the worst-case input (all letters, no operator — the shape that reaches the `Condition` constructor straight from a search query):

| input length | OLD regex | NEW regex |
|---|---|---|
| 2,000 | 78 ms | 0 ms |
| 4,000 | 149 ms | 0 ms |
| 8,000 | 650 ms | 0 ms |
| 16,000 | 2,466 ms | 0 ms |
| 32,000 | 9,893 ms | 0 ms |
| 64,000 | **39,839 ms** | 0 ms |

Each doubling of input roughly quadruples the old runtime (650 → 2466 → 9893 → 39839) — textbook O(n²), exactly matching the `java/polynomial-redos` classification. A single 64k-char query would peg a request thread for ~40s, and it keeps scaling. The possessive `[a-zA-Z]++` plus `^...$` anchoring collapses it to constant ~0 ms and fails fast with `IllegalArgumentException`.

Two notes for the record (neither blocking):
- This tightens input acceptance slightly beyond just performance: leading whitespace or an `IN` with no preceding space (`statusIN(...)`) now throws where the old backtracking regex would have parsed them. Real queries always come through spaced, so this is edge-case-only and arguably more correct.
- I pushed a follow-up commit adding the same two tests to `SqliteIndexQueryBuilderTest`, since the SQLite module got the identical source change but no coverage. Now at parity with the Postgres tests.

Nice, tightly-scoped security fix. Thanks @jhaabhijeet864.

Issue #1638, opened 2026-09-17
Security: ReDoS in ParametersUtils parameter substitution (CodeQL alert #6)


nthmost-orkes:
**CodeQL alert #6** · rule `java/polynomial-redos` · severity **high**
\`core/src/main/java/com/netflix/conductor/core/utils/ParametersUtils.java\`

Sub-issue of #1010.

## Problem
\`PATTERN\` (defined at `ParametersUtils.java:56`) is a complex regex with nested quantifiers that is applied to **every string-valued task parameter** during workflow execution (`PATTERN.matcher(paramString)` at line 258).

Task output from an external worker flows into the next task's input parameters, so a crafted string returned by a worker can drive the matcher into O(n²)-to-exponential backtracking. This is reachable in externally-facing deployments — the highest real-world risk item in the epic after the SSRF.

## Fix direction
Replace the backtracking-prone `PATTERN` with a linear/iterative parser (or a bounded, backtracking-free regex) for `${...}` expression extraction. Preserve existing substitution semantics; add unit tests with adversarial inputs that would currently blow up.

Naman-Gururani:
Hi, I'd like to work on this. Proposed approach: replace `PATTERN` with a single-pass scanner that tracks brace depth — an expression starts at a `${` not preceded by `$` and ends at its matching `}`, so nested expressions stay inside the outer range exactly as today. To make sure semantics are preserved I compared it against the current regex on every string up to length 9 over `$ { } a` (plus runs with newlines and longer brace-only strings, ~2.6M inputs): identical results, including the existing quirk that an unclosed `${` hides any expression after it — I'd keep that as is. For reference, the current pattern takes ~2.4 s on `${${…}}` nested 500 deep and ~37 s at depth 1000 (a ~3 KB string); the scanner handles both in under a millisecond. I have it working locally with unit tests (expected ranges, a differential test against the legacy pattern, and adversarial inputs with a timeout). Does this direction work for you before I open a PR?

nthmost-orkes:
> Hi, I'd like to work on this. Proposed approach...

Sure, feel free to submit a PR.

PR #1654, opened 2026-09-22
fix(security): replace backtracking regex in ParametersUtils with a linear scanner (Fixes #1638)

Naman-Gururani:
Pull Request type
----
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] WHOSUSING.md

Changes in this PR
----

`ParametersUtils.PATTERN` found `${...}` expressions with nested lookaheads and backreferences, and it is applied to every string task parameter, including output returned by workers. Its matching time grows polynomially with the nesting depth: on my machine `${${...}}` nested 500 deep takes ~2.4 s and nested 1000 deep (a ~3 KB string) ~37 s, per parameter.

This PR removes the pattern and finds the expressions with a single pass over the string (`ParametersUtils.findExpressions`): an expression starts at a `${` that is not preceded by `$` and ends at its matching `}`, tracked with a brace-depth counter. `replaceVariables` iterates over the resulting ranges instead of regex matches; the substitution logic itself is unchanged.

Semantics are preserved, including `$${` escaping, nested expressions being part of the outer range, and the existing behaviour that an unclosed `${` hides everything after it. To check this I ran the old pattern and the scanner over every string up to length 9 over the alphabet `$ { } a`, plus runs including newlines and brace-only strings up to length 12 (~2.6M inputs in total) with identical results.

Tests added to `ParametersUtilsTest`:
- `testFindExpressions`: expected ranges for representative inputs.
- `testFindExpressionsMatchesLegacyPattern`: differential test against the previous pattern (kept in the test only, as the reference for its semantics) on every string up to length 6 over `$ { } a \n`.
- `testReplaceIsNotVulnerableToReDoS`: adversarial inputs through `replace()` (depth-1000 nesting, 100k unclosed braces, 50k expressions, 100k closing braces) under a 10 s timeout. It does not complete with the previous pattern.

`./gradlew :conductor-core:test` passes locally (972 tests), `spotlessApply` applied.

Issue #1638

Alternatives considered
----

- A bounded, backtracking-free regex: `${...}` with nesting is not a regular language, so a regex cannot match balanced braces without lookaheads/backreferences of the kind that caused the problem; only a depth-limited approximation would be possible, which would change behaviour for deeper nesting.
- Rejecting or truncating suspicious inputs: would change behaviour for legitimate parameters and not address the root cause.


nthmost-orkes (approved):
This is a really nice fix. Killing the backtracking \`PATTERN\` outright and scanning for \`${...}\` in a single linear pass is exactly the right move, and you kept the substitution logic itself untouched so the blast radius stays small.

The differential test is what sells it -- keeping the old regex in the test as the reference and asserting byte-identical results across every string up to length 6 over \`$ { } a \n\` (plus the deeper local runs you mention) is far more convincing than hand-picked cases. The three semantics that matter -- \`$${\` escaping, nested expressions belonging to the outer range, and an unclosed \`${\` hiding everything after it -- all line up between the scanner and the old pattern. And \`testReplaceIsNotVulnerableToReDoS\` actually reproduces the original blowup, so we know it is guarding the real thing.

One note for context (not a change request): I checked the enterprise side, and Orkes uses a completely different, split-based substitution engine that never had this regex -- so there is no enterprise mirror to chase and nothing exposed there.

Approving. Thanks for the thorough writeup and the exhaustive testing.

Naman-Gururani:
Thanks for the review! Good call on the StackOverflowError: the nested-expression substitution recurses once per nesting level, so at those depths it now fails fast instead of hanging in the regex. Happy to add an explicit depth limit in a follow-up if you think it's worth it.

PR #1655, opened 2026-09-22
fix(core): retry a stranded SUB_WORKFLOW instead of timing it out after an hour (#1615)

### LINKED ISSUE #1615: DYNAMIC_FORK: one SUB_WORKFLOW branch never created, force-TIMED_OUT after 60 min (regression in 3.32.0 from #1369)

manan164:
Alternative to #1622. Fixes #1615.

A worker that dies inside `SubWorkflow.start()` leaves the task SCHEDULED with its message reserved for `responseTimeout` — `TaskDef.ONE_HOUR` by default. Repair skips it (since #1369 the message is present-but-invisible, so `containsMessage` is true), and on redelivery an hour later the executor force-sets `TIMED_OUT` instead of re-running `start()`. The child is never created and the workflow times out.

Two changes, both needed:

1. Don't force-timeout a redelivered SCHEDULED task when `start()` is idempotent — retry it.
2. Size its reserve from `systemTaskWorkerCallbackDuration` instead of `responseTimeout`.

(2) alone doesn't work: `hasExceededResponseTimeout` reads the same value as the reserve, so lowering `responseTimeout` makes the message redeliver sooner *and* the timeout fire sooner. (1) alone leaves the hour, which #1622 notes as unaddressed.

Recovery drops from ~60min to ~60s. Non-idempotent tasks keep the full reserve and the overrun timeout, so #1321 is unaffected.

**On not touching `WorkflowSystemTask`:** #1622 adds an overridable `isStartRetriable()`, which does follow the `isAsync()`/`isAsyncComplete()` pattern. Keying off the task type instead keeps this to one file and adds no API surface to a base class every system task extends; `WorkflowRepairService.java:156` already does the same for the same task type. Happy to switch if maintainers prefer the predicate — the two changes above are the substance.

**Not addressed:** a crashed *non*-idempotent task still waits out its reserve before being timed out. Telling "dead" from "slow" there needs lease renewal while `start()` is in flight — separate change.

Tests: a redelivered SCHEDULED SUB_WORKFLOW past `responseTimeout` retries rather than times out (#1622's scenario, driven by the real task type so there's no predicate to stub), and a SUB_WORKFLOW reserves for the short window. Existing #1321 guards unchanged. 116 tests across the executor/worker/sweeper/repair/sub-workflow suites, 0 failures.


nthmost-orkes (commented):
Suggestion:  can we give 2 * systemTaskCallbackTime a name? Reads as a magic number in the reserve path otherwise, and i keep seeing a need to be more explicit with intent in the code.

nthmost-orkes (approved):
Approving -- I ran a before/after repro on Java 21 and it holds up.

Reverted just the source fix (kept your new tests) against main: both new tests fail exactly as #1615 describes -- the redelivered SCHEDULED SUB_WORKFLOW gets force-set TIMED_OUT and removed from the queue, and the reserve is `setUnackTimeout(..., 3600000)` (the full hour). With the fix restored, all 24 pass. That `3600000` is the smoking gun -- ONE_HOUR doubling as both the invisibility window and the timeout threshold.

Checked the concurrency angle since the short reserve means a slow start() could redeliver mid-run: it's safe. Both workers compute the same deterministic child id and serialize on the execution lock in startWorkflowIdempotent (returns the existing child, no duplicate). And prod sizing is right -- systemTaskWorkerCallbackDuration defaults to 30s, so 2x = 60s (tests override to 1s).

No regressions: 132 tests across the executor/sub-workflow/repair/sweeper/system-task suites, 0 failures. #1321's guards stay put since isStartIdempotent gates on task type.

Nice work scoping this to one file and driving the tests off the real task type instead of stubbing a predicate.

